PR #16, opened 2018-03-23
Stability improvements, bonus section

Stadicus:
### 2018-03-23
* Raspberry Pi: Instead of disabling swap file, move it to the external hdd. If already removed, re-install swap utility with sudo apt-get install dphys-swapfile first.
* Bitcoin: adjusted systemd unit files bitcoind.service , getpublicip.service and lnd.service to account for a problem with binding of bitcoind to port 18333 (see discussion on bitcointalk.org). Thanks, @whywefightnet! 
* Bitcoin: added PDF version of "Mastering Bitcoin" and "Learning Bitcoin from the Command Line"
* Lightning: in "lnd.conf"
  * set "debuglevel" to info in "lnd.conf" to avoid huge log files
  * add LimitNOFILE=65536 to increase number of File descriptors
* Bonus: added new bonus section (not yet in site navigation)


leblitzdick:
Hallo,
i will simply thank you for your great work! My Raspberry Lightning node is up (leblitzdick [LND]).
TOP!!

PR #46, opened 2018-04-04
getpublicip to detect externalip has changed, and restart lnd

robclark56:
This update makes RaspiBolt compatible with dynamic public IP addresses.

To test (if you only have a static IP)

1) sudo vi /run/publicip 
(change the IP address by one digit)

2) sudo systemctl status lnd
until notice that lnd has restarted (can take up to 10 mins)

### DIFF
--- a/raspibolt/raspibolt_40_lnd.md
+++ b/raspibolt/raspibolt_40_lnd.md
@@ -30,8 +30,8 @@ while [ 0 ]; do
  fi
  sleep 600
 done;
-
 ```
+
 * make it executable  
   `$ sudo chmod +x /usr/local/bin/getpublicip.sh`
 
--- a/raspibolt/raspibolt_40_lnd.md
+++ b/raspibolt/raspibolt_40_lnd.md
@@ -14,7 +14,7 @@ To announce our public IP address to the Lightning network, we first need to get
 * As user "admin", create the following script, save and exit  
   `$ sudo nano /usr/local/bin/getpublicip.sh`
 
-```
+```bash
 #!/bin/bash
 # RaspiBolt LND Mainnet: script to get public ip address
 # /usr/local/bin/getpublicip.sh
--- a/raspibolt/raspibolt_40_lnd.md
+++ b/raspibolt/raspibolt_40_lnd.md
@@ -13,17 +13,24 @@ To announce our public IP address to the Lightning network, we first need to get
 
 * As user "admin", create the following script, save and exit  
   `$ sudo nano /usr/local/bin/getpublicip.sh`
-```bash
+
+```
 #!/bin/bash
 # RaspiBolt LND Mainnet: script to get public ip address
 # /usr/local/bin/getpublicip.sh
 
 echo 'getpublicip.sh started, writing public IP address every 10 minutes into /run/publicip'
-while [ 0 ];
-    do
-    printf "PUBLICIP=$(curl -vv ipinfo.io/ip 2> /run/publicip.log)\n" > /run/publicip;
-    sleep 600
+while [ 0 ]; do
+ source /run/publicip
+ CURRENTIP=$(curl ipinfo.io/ip 2> /run/publicip.log)
+ echo  PUBLICIP=$CURRENTIP > /run/publicip;
+ if [ "$CURRENTIP" != "$PUBLICIP" ];then
+  echo Restarting lnd.service New external IP = $CURRENTIP
+  sudo /bin/systemctl restart lnd.service
+ fi
+ sleep 600
 done;
+
 ```
 * make it executable  
   `$ sudo chmod +x /usr/local/bin/getpublicip.sh`


Stadicus:
This is nice to detect public ip changes. However, once lnd is restarted, the wallet will be locked and the node will not be functional.

* On mainnet, lnd should anways be run with a strong wallet password
* I do not know of a safe way to unlock the wallet automatically
* To my knowledge, the public ip is used to announce the node on the network, but not necessary for regular operations or opening channels

So, sacrificing securty (to run lnd without wallet password) or restart lnd into a non-operational state is not really helpful. 

Am I missing something?

robclark56:
I don't think this is correct: "To my knowledge, the public ip is used to announce the node on the network, but not necessary for regular operations or opening channels"

My assumption is this:
If there incoming channels open to your node, and the external IP address changes (but the new IP address is not rebroadcast), then your node has effectively gone offline as nobody can connect to it and peers will see your node as offline.

I will do this test to confirm
1) change my external IP address
2) NOT re-broadcast (i.e. not restart lnd with new --externalip)
3) Check  [here ](https://explorer.acinq.co/#/)to see if the world sees my new IP address, or not
(wait 24 hrs)

If:

- The world does learn my new IP address - then all my concerns are unfounded

- ... does not, then changing externalip address (and no restart of lnd) is indeed an issue.

Stadicus:
Just had several discussions in the `lnd` slack. The `externalip` is needed to advertise your node to the network, so that others can open channels with you and use your node. Of course it's not optimal if you announce your ip, and you just vanish for other peers explicitly looking for you.

If your ip changes, this won't affect normal operations, as you can still connect to and communicate with your peers. Also, created bot not yet paid invoces are not affected. Only if both nodes Alice & Bob both change their ip, then they can no longer communicate and the channel needs to be closed.

So IMHO it's not worth it to bring down an operational node to update the ip address, when the wallet is locked afterwards. The whole point of the RaspiNode is that it's running 24/7 without manual intervention.

You mentioned a script that unlocks the wallet automatically. Is this working for you? How is security guaranteed that someone with physical access to the node (without login, so not the running system) cannot gain access to your funds?

The best way IMO is to enable `lnd` to announce an ip change to the network without restart. This would solve all our issues and can be completely automated with your script. Vegard (@vegard in Slack) is looking at our feature request and might be able to implement it in `lnd` "over the weekend"... :-)

robclark56:
Really appreciate the feedback here.

1) I can not see how peers can still communicate with my node if the external IP changes, but not announced.

12hs ago, I changed my external IP, but did not restart lnd. If I query my own node, it returns the wrong (old) IP address

admin ~  ฿  lncli getnodeinfo <mypublickey>
{    "node": {
        "addresses": [
            {            "network": "tcp",
                              "addr": "106.68.112.250:9735"               <----  Old IP
            }
}

But 
admin ~  ฿  curl ipinfo.io/ip
58.6.87.18 <---- Current IP

Also https://explorer.acinq.co still shows my Old IP address after 12+ hrs

2) All this is irrelevant if Vegard can upgrade lnd to manage this issue without a restart. This is a much better solution.

:-) 

BTW ... if I was designing the logic it would be this
 - Today, --externalip does 2 things: (1) advise the IP, (2) announce that this node is accepting connections
- Instead, I would (1) Announce "I want to accept connections", but never announce my IP. My regular broadcasts/messages from my node have my external IP address in them  (TCP Source IP). So the LN network can capture my External IP from these messages. More than that .... it can notice if my External IP has changed (by comparing source IP in each message to what it thinks my source IP should be) and update as needed.
- Alternatively  ... add new parameter called --externalfqdn,  
(e.g. --externalfqdn=raspibolt.my.domain.com) and as that never changes (so long as you uses a dynamic DNS service, or have a static IP). Then the LND network simply remembers the fqdn instead of the ip, and peers connect using the fqdn. This follows the philosophy of the Internet/DNS much better than sharing IP addresses.

Fingers crossed that lnd is updated to handle dynamic IPs as it seems this was never considered in the original design.

3) "You mentioned a script that unlocks the wallet automatically."
Q: Is this working for you? 
A: Yes - perfectly. I have it documented in a commit, but not made the pull request. So you want me to make a pull request  so you can see it (or can you see commits that have not had a pull request.?)

Q: How is security guaranteed that someone with physical access to the node (without login, so not the running system) cannot gain access to your funds?
A: If a person has physical access to my system but can't login ... then I can't see that there is any difference between a  manually or automatically unlocked physical system. If they stole my raspibolt and booted it up - the wallet would be unlocked - Yes. Whereas if yours was stolen and powered up, it would have a locked wallet. 

But if they can't login, then I can't see how they can spend my funds???

BTW ... if they can't login, they can't read my wallet PW
If they CAN login, they can see my password .. but they also have my macaroons which is just as bad (??)

Stadicus:
I think this conversation needs to be on the LND slack, as my knowledge is kinda limited... :-)

A few remarks:

* The public ip will stay "incorrect" unless you restart `lnd`, so other nodes and also Lightning explorers (also just a node) will not update their information and cannot connect to you. Open channels and invoices continue to function, however.

* Support for FQDN would be great, I agree. 

* The security I was referring to is physical access. It's trivially easy to just take the SD card or external hard disk and mount it on a different machine. If both wallet password and wallet are accessible, it's game over. No need to login or know your user password if the file system is not completely encrypted.

* So, the external unlocking of the wallet with the readonly.macaroon might be worth implementing on a different machine. You already started describing it on Windows, I'll probably do the same on Linux.

Running of the a long weekend in Venice, back next week. 👍

Stadicus:
See comments in https://github.com/Stadicus/guides/pull/59

PR #50, opened 2018-04-05
Add invoice macaroon after lnd upgraded to 0.4.1-

robclark56:
Before 0.4.1
macaroons generated = readonly and admin

After 0.4.1
macaroons generated = readonly and admin and invoice


Stadicus:
See https://github.com/Stadicus/guides/pull/51

PR #52, opened 2018-04-05
Possible Inclusion in the Bonus Section

robclark56:
Feel free to reject this.

I decided to learn how to run lncli on my Windows PC - and manage the local RaspiBolt.

So - I documented it.

As a side benefit: maybe a script on the remote computer could monitor & unlock the wallet as needed. This way - the wallet password is not stored anywhere on the RaspiBolt.  (but it would be stored on the remote PC).

If the remote computer was on the WAN and only had readonly.macaroon or invoice.macaroon, then the funds in the wallet would be safe - and yet the wallet can still be unlocked. If a hacker read the Password - they would still need either physical access to the RaspiBolt, or the admin.macaroon to do any harm.

Rob


Stadicus:
That's really cool! Unlocking the wallet from a remote, secure machine is actually the solution everyone is pointing to. It's just not easy to set up without having a second machine running 24/7, but this could be handled e.g. by a Synology NAS or a second Pi, as I also run a dedicated Pi-Hole. And a RetroPie. And a Kodi Streaming client. :-) Will need to check this out.

robclark56:
1) Yes .. will remove section on access from outside LAN. Power users can work that out easily themselves

2) I am also working on using a cron job on a 24/7 server I have in USA (Hostgator) that is quite secure - that is what I will use for my unlock_wallet trick.  But .... for the purposes of documentation that everyone can try - was thinking of setting up also on https://cloud.google.com/free. I think that if only using the readonly macaroon ... then it is pretty safe. Even if the user's Google Cloud acct was hacked ... the hacker would still need physical access to the RaspiBolt. Of course - I could also add "This was just a testnet demonstration. For mainnet, you should only do this on a server you control"

What do you think?

(Enjoy Venice!)

robclark56:
Removed the part on how to use outside LAN.

:-)

PR #54, opened 2018-04-08
External Wallet Unlock

robclark56:
Controversial - as this technique requires the wallet password in clear text in an off-site script.

It is the best I can come up with.

Good news - it that it works, and seems a safe option,

Please have a look.

I wanted to document this for my own purposes .... so no offence taken if you decide not to use this.

Rob


Stadicus:
Hej Rob, that's amazing, I didn't know one could do that. :-)

I think it's a bit outside of the scope of the Beginner's Guide, though. I have concerns regarding complexity, our own knowlege regarding "public RPC best practices" and the need for user support. 

Do you want to publish it on your own somewhere? I'd be happy so link to it.

robclark56:
As you recommended, I have published the guide myself here

https://github.com/robclark56/RaspiBolt-Extras/blob/master/RB_extra_01.md


I am working on a number of 'extras'. 

1. Remote Unlock (link above)
1. RaspiBoltDuo: Having mainnet and testnet running simultaneously. (ie 2 x bitcoind & 2 x lnd) on the same RasPi  (Status: 2 x Bitcoind = Working. Next to get 2 x lnd)
1. Working with dynamic IPs (looks like it will be in the next release: https://github.com/lightningnetwork/lnd/compare/master...vegardengen:change-externalip :-) Thanks for your help getting that to happen)
1. External Webpage to display a Payment Receipt (status: not started)

Stadicus:
Cool, will include the link in the bonus section. :-)

Regarding your point 3: I'm not yet sure this will be merged Roasbeef is not too happy with this change, but it could also be achieved using an external script, according to him: issue https://github.com/lightningnetwork/lnd/issues/1057

PR #59, opened 2018-04-15
Restarts lnd if externalip changes

robclark56:
As it is not clear if/when we will get a fix to lncli that allows updating externalip on the fly, at this point in time .. this is the next best.

If externalip changes and not restarted lnd becomes pretty useless as this node becomes invisible on the LN. 

Needless to say ... this locks the wallet .... and a possible solution is here:
https://github.com/robclark56/RaspiBolt-Extras/blob/master/RB_extra_01.md


Stadicus:
Not suitable for this beginner's guide, as the external announcement of the ip address is not relevant for the correct operations of the node. Connected channels are still working, a more basic solution is worked on directly on LND. 

I linked to your guide in the "Bonus" section.

PR #66, opened 2018-05-01
added flags to systemd bitcoind.service file

LilChange:
i needed to add the `-daemon` flag to make the systemd job not hang and crash. Also added the `-conf` flag for more precise. Was there a reason you omitted these? If no, here is a pull-req.

might it also be good idea to add the hardening configs as suggested in official [bitcoin systemd file](https://github.com/bitcoin/bitcoin/blob/master/contrib/init/bitcoind.service#L23-L41) ?


Stadicus:
Thanks for the PR, strictly speaking these options should not be necessary. But following the official template and your experience that it helps to prevent issues makes it worthwile in any case.

PR #100, opened 2018-06-29
adding allow rule to connect with ZAP wallet

elkimek:
See more at https://github.com/LN-Zap/zap-desktop


Stadicus:
I don't think the Zap wallet is ready yet for a beginner's guide. Will be interesting to reinvestigate once connecting to your own LND node is a standard functionality.

PR #121, opened 2018-07-12
Check public availability of Bitcoin Core

Stadicus:
New method to check public availability of Bitcoin Core, using netcat


lavenetz:
Hi @Stadicus et al.
Also have "Public  Not reachable" since a few days! Any tipp?

PR #131, opened 2018-07-15
Add check to see if IP:PORT is open for inactive channels

robclark56:
Enhancement to lnchannels to show if a node is not responding if a channel is inactive

In this example:

*  Channels to KRYPTO.KOELN are inactive, but the node seems active
*  Channels to _poole_party_ and _Stadicus RaspiBolt_ are inactive, and the nodes seem down (?)

image


Stadicus:
Yes, my node can be down from time to time. So many ideas, so little time... :-)

PR #160, opened 2018-07-30
Add information on 32 bit Bitcoin version download

fjahr:
The latest RPis are running a 32 bit operating system even though they have 64 bit processors. This tripped me up for a moment when downloading the latest version of Bitcoin and may be valuable information for anyone who is not experienced with Raspberry Pi.


Stadicus:
Thanks, great addition!

PR #168, opened 2018-08-11
restart typo removed

notbrain:
`shutdown -r now` not `restart shutdown -r now` after setting new swap file.


Stadicus:
Thanks for this nice catch! :-)

PR #205, opened 2018-09-15
Distinct ports for zmqpubrawblock, zmqpubrawtx

rodasmith:
`lnd` now needs distinct ports for `zmqpubrawblock` and `zmqpubrawtx` so just use the defaults


Stadicus:
Thanks for this PR, much appreciated! 👍

PR #209, opened 2018-09-17
Update README.md

rek79:
Updated bitcoin.conf file to have two distinct zmq sockets.  LND v0.5 requires this otherwise it will not start.  Download links for latest version of Bitcoin Core and LND updated.


Stadicus:
Thanks for the PR, but there are more changes to be considered when upgrading to LND v0.5, this would break stuff. The RaspiBolt guide is already updated, maybe I'll start the next Odroid Magazine article with the upgrade instructions, let's see.

For the moment being, I'll keep the guide on the version 0.4.

PR #214, opened 2018-09-17
Update macaroon path

frennkie:
for LND v0.5-beta


Stadicus:
Thanks, the huge has been updated with additional comments.

PR #229, opened 2018-09-21
update lncli-web macaroon folders

CodingMuziekwijk:
Added a few lines to choose the right folder to copy the admin.macaroon from. The location where admin.macaroon is created changed since LND-0.5. If you run this tutorial for the first time, it will now be compatible with LND-0.5.


Stadicus:
Thanks!

PR #231, opened 2018-09-22
Add support for Litecoin and other coins.

coblee:


Stadicus:
Thanks for that addition! :-)

PR #232, opened 2018-09-22
More examples of lightning mainnet applications

hakonamatata:
Hi

Added another example of lightning applications where one can pay with lightning and also instantly cash out using lightning.


Stadicus:
Hej, thanks for the link. This is not meant to be a directory of individual apps, but I'll put a link to https://lnapps.info/ there instead.

PR #234, opened 2018-09-23
Update to fix minor directory/screenshot inconsistency

Sowiedu:


Sowiedu:
I think I will create a pull req from my fork since I changed a screenshot file. I don't really know how git hub works yet :-)

PR #246, opened 2018-10-02
Fix for the LND 0.5 break

martinatime:
With LND 0.5 the macaroon location was moved. This commit fixes the location and is variable based on whether testnet is being used.


martinatime:
> unfortunately still not working for me

@q0de any change now or is it still not working?

Stadicus:
Nice, thanks for the fixes. It works for me on startup and if I run it with `sudo` as "admin".

It's very slow, though, taking almost 30 seconds to log in... Not sure which part causes this, but it might be helpful to use timeouts for some slower parts. (eg. https://stackoverflow.com/a/10225072)

PR #256, opened 2018-10-17
LND autostart (fix?)

CodingMuziekwijk:
Changed: 
$ sudo systemctl enable lnd
To:
$ sudo systemctl enable lnd.service

Not sure if this is a good commit, cause i'm not 100% sure this actually was my problem so please give some feedback on this one. On my raspibolt the LND client didnt startup at boot without the '.service' behind the line.


Stadicus:
Both commands work, that should not make a difference

PR #265, opened 2018-10-27
Update lnchannels

Janaka-Steph:
For testnet, calling `lncli` with --network=testnet is required as of v0.5. 
--rpcserver=localhost:11009 is not correct for me, works fine without anything.  
PID should be lnd instead of lnd_testnet


Stadicus:
Thanks for your contribution! 👍

PR #277, opened 2018-11-16
Edited raspibolt script to use variable for HDD location

oitdmser:
changed the external HDD section to use a variable that the user can specify at the beginning of the script to make it easier to adapt to different usage scenarios


Stadicus:
Changed the values for lnd & bitcoind directories back to valid defaults

PR #311, opened 2018-12-25
Update to Bitcoin 0.17.1

dennisreimann:
These are the necesarry changes for the latest [Bitcoin v0.17.1](https://bitcoincore.org/en/2018/12/25/release-0.17.1/), which has been released today.


Stadicus:
Explicit fingerprint check does not work with PGP anymore. As it's redundant, I removed it.

PR #317, opened 2018-12-28
Include -rf when clearing admin download dir

cilphex:
One of the items to be removed when running this command is the bitcoin download directory. Since it's a directory, we should specify `-rf` when running `rm`.


Stadicus:
Good point. I added full path to quite dangerous `rm -rf` command, just to be sure

PR #325, opened 2018-12-31
Remove txindex

Stadicus:


Fillippone:
Canu out comment on why you removed this options? Txindex is needed to run a full node isn’t it?

Fillippone:
Got it. Bitcoin Core v17.0 allowed that. Commented more on a issue page.

PR #333, opened 2019-01-06
Add aliases section in Additional Scripts, fixed link to yalls.org to…

Septem151:
… be testnet version in LND section, added (optional) alias header in LND and Mainnet sections.


Septem151:
I wanted to share some bits of info for users not familiar with linux and create a section of the Additional scripts guide that includes setting up short little aliases for common commands in lncli. I personally use these on my own node and it's helped me a ton (especially on testnet since I don't have to type --network=testnet every time). I know this isn't a necessary addition, but it is a very helpful tool to have. This is why I've decided to include the scripts in the "Bonus" section.

Stadicus:
I agree that these alias' can be useful. Not sure if the additional shell scripts are worth the effort, though. Isn't that also achievable by just creating an alias?

Septem151:
I wasn’t aware aliases could accept parameters, so you are right. I’m new to github pull requests so let me figure out how to edit this :)

PR #339, opened 2019-01-10
Consistent header formatting

badokun:
It's the little things ;-)

I noticed not all the headers appeared the same, lacking in bold styling on some


badokun:
Will create the pull request after I add the metrics monitoring section

PR #358, opened 2019-01-23
add  add the Bonus guide: Zap Desktop Lightning Wallet 

openoms:
This is a Bonus guide to have the Zap Wallet connected on LAN. It requires configuring the Pi and exporting the admin.macaroon and tls.cert, described in the guide. Please tell me to change anything to fit in with the standard followed so far. There are no more two digit bonus guide places, I was thinking to go with raspibolt_71_zap.md vs. raspibolt_610_zap.md or just make a pointer to a separate repository of mine like @robclark56 did. 
Please keep in mind this is my first pull request also. I love what you do here!


openoms:
ssh and scp might even work on windows since openssh is built in to PowerShell since Win 10 ver 1709. Will test the commands later today.

openoms:
Hi, thanks for looking at it. I am so happy to be able to contribute a little to this project, since I am using it daily. Yes, I agree the Windows bits and screenshot would be more consistent with the rest of the guide and probably helpful for more users. I am using Ubunu as my main OS, so did not get around to do the Windows part. ( I also have a page for Joule, but it probably can be simplified a bit as it is now possible to pair Joule with hexstrings)

PR #403, opened 2019-03-06
Add bitcoind channels, lnd pending/total sats

robclark56:
newwelcome

CHANGE 1
Cleaned up the repeated instances of "/usr/local/bin/lncli --macaroonpath=${macaroon_path} --tlscertpath=${lnd_dir}/tls.cert"

CHANGE 2
Added in/out channels on bitcoind (As requested here: Stadicus#370)

CHANGE 3
Added balance locked in Pending Channels.
The importance of this is recognised by anyone that has had to do a lncli closechannel --force. The channel's csv_delay value may cause the pending balance to be locked up for a period of 1 week (or more). Seeing the funds are 'pending' is a bit less likely to cause panic.

CHANGE 4
Added a grand total of sats in 
* wallet
* channels
* pending on-chain transactions

### DIFF
--- a/raspibolt/resources/20-raspibolt-welcome
+++ b/raspibolt/resources/20-raspibolt-welcome
@@ -136,7 +136,7 @@ else
   macaroon_path="${lnd_dir}/data/chain/bitcoin/mainnet/readonly.macaroon"
 fi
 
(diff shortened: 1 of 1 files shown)


giulnz:
Hi i tried it and it works. thanks

Fillippone:
It is ok! Great improvement

janglorumo:
Worked great! Thank you very much :+1:

PR #413, opened 2019-03-21
Wrong binary checksum

giulnz:
the binary to check is arm , not windows


Fillippone:
We are talking about the auxiliary PC, aren’t we? That’s a win machine.

giulnz:
Ah i see ... raspibolt_10_preparations.md -> "Usually, you would also need to check the signature of this file, but it's a pain on Windows, so we will do it on the Pi later on."

PR #414, opened 2019-03-21
Add zap ios guide

giulnz:
I found this pull request Stadicus#408 from AmazingDroopy and first i wanted just format better his guide but finally i re-make it. Sorry AmazingDroopy :(

I just tryed it last week , but for me using putty the qr was too big and can't resize.
So i export as a url (-j option) and copy past with telegram (i know is so bad but i was in testnet)
I also find out a better way with (-o option) Output QRCode to file, and after transfert with WinSCP to my windows and scan from there.

Hope this guide can be usuful Don't know if was better to fuse zap desktop and ios toghether but actually i made this.

NOTE : i don't know if images links are correct , i can't see in preview.


giulnz:
@walterkok, you right. Thanks for the check !

walterkok:
Got the. Zap IOS app to work and at the same time RTL on my local network at home, with the following setting in lnd.conf (solution from saubyk)

restlisten=0.0.0.0:8080 # for RTL
rpclisten=0.0.0.0:10009
tlsextraip=0.0.0.0

I ised to be able to access with Zap IOS my pi remotely ( with tlsextraip=publicipaddress ) and portforward of 10009 to my Pi, but than RTL does not work. 

Any ideas on a solution? Thanks Walter

giulnz:
Don't know sorry.

Stadicus:
Thanks for this great bonus guide. I did some wording changes and streamlined the Pi part. As I don't use Apple products, could any of you check that the iPhone part works reliably?

walterkok:
Hi, I am using an Iphone and the above works well indeed (with the two changes I suggested). Both on the internal wifi network as well as when you approach it from public mobile infra (with external router ip and portforward of 10009)

In the combination with RTL, you need to add in the comments as per above
restlisten=0.0.0.0:8080 # for RTL
rpclisten=0.0.0.0:10009
tlsextraip=0.0.0.0
It than also works fine as long as you are on the same wifi network. I still do not have both working at the same time from public mobile network.
Regards W

giulnz:
@walterkok as there is no RTL guide in RaspiBolt, i think we can merge Zap iOS guide how it is now. 
If we find out how to make work both together we will update later. 

@Stadicus with your change now is better. I was looking for make lndconnect work without set go and now i understand. I just re-test and all works great. I don't really like how the 2 images example are visualized in the botton of the page but i think it's readeable.

Stadicus:
Thanks for your work and feedback. I finally surrender to using `0.0.0.0` to listen from all ip addresses, I guess UFW is doing it's job to block malicious stuff from the outside. :-)

walterkok:
hi Stadicus, i just looked at your guide, and what seems to be missing is
export GOROOT=/usr/local/go  
export GOPATH=$HOME/gocode 
export PATH=$GOPATH/bin:$GOROOT/bin:$PATH 

If I do not execute these commands on my PI first lndconnect will not execute, could be a setting on my PI but just to be sure it has not been an oversight

giulnz:
@walterkok you probably follow previous version, if you check the commit https://github.com/Stadicus/guides/pull/414/commits/bbcb674ef09a1810581d4ceb45a5291e1b1b0e6d you will see that we before compile lnd connect with go , with stadicus method you just use the birary

giulnz:
you just need to https://github.com/Stadicus/guides/pull/414/commits/bbcb674ef09a1810581d4ceb45a5291e1b1b0e6d Install LND Connect

walterkok:
Ah many thanks will do

PR #426, opened 2019-04-12
Added LN Terminal Dashboard

cidadededeus36:
View lightning network channels in an interactive text-mode viewer for Unix systems.


giulnz:
Great tools, i like to compile myself, but seems more standard as other extra raspibolt guides to use binary. Also we can wait a little to integreate it (i think a lot of noobs follow raspibolt guide) so it will more tested. I see Edouard Paris just release firs 0.0.1 version but we need linux-armv7. So we can ask for it and change your guide with binary.

Stadicus:
Thanks @cidadededeus36! I agree that using the binary directly is preferable, but as it's in the Bonus section, I think we could still add it with a warning that it's a reeeeally alpha release. Need to try it out first, probably within the next two weeks.

Stadicus:
I asked Edouard to provide binary images for ARM architectures and he promptly followed up:
https://github.com/edouardparis/lntop/releases

Maybe you can test them and update the guide to download it directly?

PR #428, opened 2019-04-14
Add a static IP for the Pi with the DCHP config file

daniel-vera-g:
I am using a router, which does not have the possibility to set specific IP addresses for specific devices. Therefore, I changed the DHCP-Client config file on the Pi to always advertise the same IP to the DHCP-Server. Could be helpful to peers which do not have the possibility to set specific IP's on their routers.

### DIFF
--- a/raspibolt/raspibolt_20_pi.md
+++ b/raspibolt/raspibolt_20_pi.md
@@ -77,11 +77,43 @@ We now need to set the fixed (static) IP address for the Pi. Normally, you can f
 
 :point_right: need additional information? Google “[your router brand] configure static dhcp ip address”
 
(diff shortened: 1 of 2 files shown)


Stadicus:
I'd rather have that under the FAQ section as this is not very common. Would you mind putting it in there?

daniel-vera-g:
> I'd rather have that under the FAQ section and link to it from here, as this is not very common. Would you mind putting it in there?

Done. Would link it, when it's merged ;)

Stadicus:
I merged you addition to the FAQ manually, as I've just remodelled the whole repo structure.
https://github.com/Stadicus/RaspiBolt/commit/1dc1390cd48d09e6b10e31f7e281d0925e3ce351

PR #446, opened 2019-05-10
keep lnd and bitcoin last version link in the main section

giulnz:
i think it's more easy to keep lnd and bitcoin last version link in the main section so i try to modify the FAQ upgrading LDN section and preserve the "If upgrading from a version lower than v0.5"
Don't know if it is possible to immidiatly upgrade to last or need to do in 2 step ( first to 0.5 and after to last)


giulnz:
Need new structure for GitHub Pages

giulnz:
is it the right way to use branch ? don't know why there are this 3 voice : 
- png update with UPnP discovery 979324d
- Merge pull request #3 from Stadicus/master e1e00d6
- Merge pull request #4 from Stadicus/master 57d66dc
but the only file it change with the pull request is FAQ ... 
what do you suggest for future ? which is the right way ? thanks

Stadicus:
Thanks for updating the FAQ and don't worry about your commit history. I'll "squash" it into a single commit anyway when merging, so this is exactly how contributions should work. :-)

PR #447, opened 2019-05-10
Dropbox static channels backup guide

giulnz:
I think download a script and use dropbox is an easy and safe way.
The only problem i see is the script download from an external source , that's also happen in other RaspiBolt bonus guide and so i decided to force user to manually check & modify the script (you need to insert the dropbox token)

I check and tested since 3 week this script. For sure it will be better to have a general one that give you the choise between dropbox, own server, aws, or other. But as the channels.backup file is encrypted i think Newbie user can follow this and advance user have no problem to make their own solution


giulnz:
Need new structure for GitHub Pages

Stadicus:
Thanks, will try this out in the next few days.

PR #462, opened 2019-07-01
Create SECURITY.md

GhostsOfHiroshima:


Stadicus:
I like the idea, but as I'm not running a bug bounty program and no custom software is contained in this repository, I don't think that's necessary.

PR #471, opened 2019-07-19
Update Raspbian from Stretch to Buster

kristapsk:
There are no Stretch in given Raspbian download link anymore.


Stadicus:
Thanks!

PR #473, opened 2019-07-28
Added missing mkdir /mnt/hdd/bitcoin command

findiggity:
Symbolic link will be red as the /mnt/hdd/bitcoin directory doesn't exist.  Added required mkdir command.


Stadicus:
Thanks. This is not necessary, as the directory `/mnt/hdd/bitcoin` is created when mounting the hard disk:  
https://github.com/Stadicus/RaspiBolt/blob/master/raspibolt_20_pi.md#mounting-external-hard-disk

PR #492, opened 2019-10-09
Update motd and update to LND 0.8.0-beta

53645714n:
I updated the motd to display the version of bitcoind and lightning and to check for updates and show when an update is available. It does this by checking the -version to the releases/latest on github.
Screenshot from 2019-10-10 14-59-00

I also re-added support for earn.com since my router does not support NAT loopback as do so many others and wrote instructions on how to switch in the manual.

### DIFF
--- a/raspibolt_40_lnd.md
+++ b/raspibolt_40_lnd.md
@@ -33,8 +33,8 @@ $ gpg ./pgp_keys.asc
 $ gpg --import ./pgp_keys.asc
 $ gpg --verify manifest-v0.8.0-beta.txt.sig
 > gpg: Good signature from "Olaoluwa Osuntokun <laolu32@gmail.com>" [unknown]
-> Primary key fingerprint: BD59 9672 C804 AF27 7086  9A04 8B80 CD2B B8BD 8132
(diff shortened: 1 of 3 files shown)


Stadicus:
Thanks for your contribution! Will test first, probably needs to wait until after Lightning Conference in Berlin.

53645714n:
> Thanks for your contribution! Will test first, probably needs to wait until after Lightning Conference in Berlin.

No problem, have a greate conference and if you have any questions, just ask.

53645714n:
I will reopen when I fixed the issue. I need to do a better version check, since the new lnd version had more characters then the old one, it displays a false update flag now.

53645714n:
Fixed the issue with the version number and added update to LND v0.8.0-beta.

53645714n:
@Stadicus , I see the version update has been taken care of in another PR, should I close this PR and do another for only the MOTD? (Which might have been better in the first place..)

Stadicus:
Opening an update just for the MOTD would be great, thanks!

53645714n:
New pPR #505 created

PR #498, opened 2019-11-22
Update Zap iOS instruction for Tor

dennisreimann:
To make it work I also had to add the `tlsextradomain` to the lnd.conf to make it work.


dennisreimann:
Let me know in case this makes sense. I can also provide an update to this PR that does not affect so many lines (mostly trimmed whitespace). It went this way because initially I changed the doc to make use of the gRPC API, but then found out it works with the REST one.

PR #504, opened 2019-12-06
Improve wording in UFW setup

dnlggr:
Changes [redundant](https://en.wikipedia.org/wiki/RAS_syndrome) usage of _local **L**ocal **A**rea **N**etwork_ to just _**L**ocal **A**rea **N**etwork_.


Stadicus:
Good point 😂

PR #505, opened 2019-12-08
Version info and earn.com support

53645714n:
The motd script now shows version info and had re-added support for earn.com since my router did not support nat loopback as used in the previous version of the script.


Stadicus:
Thanks, will test

Stadicus:
Tested, and works with some minor permission fixes in the updated guide.

I'll overhaul it a bit to also show the Tor .onion and no longer showing the IP address, as the updated v2 guide is Tor-only now.

Stadicus:
BTW: I love the "update checker" feature! :+1:

PR #514, opened 2019-12-18
Correct `lnd --version` example output for 0.8.2

kristapsk:
Guess this was just missed in #507. Noticed when upgrading LND on my RaspiBolt.


Stadicus:
Thanks!

PR #545, opened 2020-01-23
Update raspibolt_10_preparations.md

GlenCooper:
Suggest KeepassXC in favor of Keepass.  KeepassXC (https://keepassxc.org) updates are frequent on their github (https://github.com/keepassxreboot/keepassxc).


Stadicus:
Good point, thanks!

PR #547, opened 2020-01-26
Add installation of BTC RPC Explorer

diegogurpegui:
Add a bonus guide for the installation of BTC RPC Explorer.

I reorganized the Bonus menu entries so this new one is just before "System recovery".

The BTC RPC Explorer is installed locally on the SD Card (home dir), since it doesn't take much space and doesn't have a database. However an improvement of this guide would be to install it in the HD along with the rest of the software. `git clone` has issues when using external drives so the guide could be update to deal with that with proper instructions.


Stadicus:
Awesome bonus guide, thanks! 

The one issue I see at the moment is that the RaspiBolt guide standard setup does not have `txindex` enabled. This means that a full blockchain rescan over multiple days is necessary.

I will play around with it a bit, also in combination with the Electrs server.

diegogurpegui:
That's a very good point, the `txindex` issue. I always have it on, that's why I didn't noticed that pre-requisite.
Perhaps just mentioning it in the **Preparations** section as a requirement is enough? But of course it wouldn't be very helpful for the user to find that out at this point of the configuration. Perhaps, instead, when setting up the Bitcoin node it's worth mentioning this parameter and what it can be helpful for. And even link other sections of the guide where it's going to be required so the user can plan ahead.

Regarding Electrs, as far as I understand it should work the same as the other Electrum servers since the API spec is the same. But I didn't try it so I'm not sure.

csh2000:
Hi, would be nice to include also a "Tor hidden service" for nodes running with Tor (Raspibolt II) in order to be able to accesss remotely (similar to Remote access in Electrum).

Not sure if it is neccesary to add something in "nginx.conf".

Thanks!

diegogurpegui:
Good call @csh2000. The guide is meant for accessing it from the local network, but we could add some kind of "annex" at the end in case the user wants to configure Tor for using the explorer from outside the local network.

The explorer does not use nginx, it just runs withe the "node" binary.

csh2000:
Thanks @diegogurpegui 

Running BTC RPC Explorer on a Raspibolt I (clearnet) following your guide, and everything seems to run fine in the Pi (image attached).

But when I try to connect from W10 explorer I cannot connect: "192.168.1.33 refused to connect"....checked everything several times and open port 3002 in router, but I get always this message. 

I guess it is an issue with W10/explorer, any advice?

Thanks!

BTC RPC Explorer

csh2000:
Hello, @diegogurpegui , solved the issue.....I forgot to comment BTCEXP_HOST & BTCEXP_PORT (image attached).

I connect from another computer in my network and now works fine.

I think it would be fine to include this in the guide...it is obvious, but saves time if it is indicated.

Thanks again!

BTC RPC Explorer

diegogurpegui:
Good catch @csh2000, thanks!
I'll update the guide to clarify on that configuration field in case someone wants to use it outside localhost.

BTW, I'm glad to made it work!

diegogurpegui:
I already added the comment about the bind parameters in the BTC RPC Explorer config file, as suggested by @csh2000.

Also, @Stadicus, I took the liberty of adding a small paragraph in the Bitcoin section to warn the user to use txindex configuration in case he will need it in the future. Take a look to see if it's ok for you.
I also added a note to that same txindex configuration in the "Preparations".

Stadicus:
Hej Diego, I went over your great guide and made some changes in https://github.com/Stadicus/RaspiBolt/pull/547/commits/447e044ba957428fffa3de6de047daaba3ca8567, mainly for consistency with the rest of the RaspiBolt guide:

* use UFW subnetmask `/16` which covers all IP addresses `192.168.*.*`, this should make adjusting the rule unnecessary
* clone specific release of Git repo (`--branch`) instead of master and don't clone the whole history (`--depth=1`)
* Use Electrs as Electrum server by default

Please have a look at it and provide some feedback if the changes are ok with you. I think, even moving this guide into the main part of the guide as its own section after Electrs would be a good next step.

diegogurpegui:
I just reviewed the changes @Stadicus, they look great!

I can get behind the idea of moving it to the main section :).
On one hand, it's not a critical part of the environment for a user and we probably don't want to make the main guide too heave on users. But on the other hand, having a controlled block explorer do provide a much better degree on privacy and control. Plus, it's a user-friendly way of checking the status of the node (% validated, chain tip, peers connected, mempool, fees, etc).

Stadicus:
Created a bit of a mess when I rebased your commits on top of the latest version of the guide. Had to merge your branch manually (see https://github.com/Stadicus/RaspiBolt/commits/master), but your guide is live!

Will close this pull request, as it already has been merged manually.

PR #552, opened 2020-02-10
Replace usermod with adduser

interrupt00:
Hi,

I suggest to replace `usermod` with `adduser` to add users to groups, since `adduser` is more user friendly and it is already used in other places (e.g. `raspibolt_20_pi.md`).

### DIFF
--- a/raspibolt_30_bitcoin.md
+++ b/raspibolt_30_bitcoin.md
@@ -278,7 +278,7 @@ After rebooting, the bitcoind should start and begin to sync and validate the Bi
 * See bitcoind in action by monitoring its log file (exit with `Ctrl-C`)
 
   ```sh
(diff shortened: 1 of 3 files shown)


raspibolt_30_bitcoin.md
```diff
@@ -278,7 +278,7 @@ After rebooting, the bitcoind should start and begin to sync and validate the Bi
 * See bitcoind in action by monitoring its log file (exit with `Ctrl-C`)
 
   ```sh
-  $ sudo tail -f /mnt/ext/bitcoin/debug.log
+  $ tail -f /mnt/ext/bitcoin/debug.log
```
interrupt00: `admin` belongs to `bitcoin` group already.

PR #553, opened 2020-02-10
Replace usermod with adduser

interrupt00:
Hi,

I suggest to replace `usermod` with `adduser` to add users to groups, since `adduser` is more user friendly and it is already used in other places (e.g. `raspibolt_20_pi.md`).


Stadicus:
Good suggestion. I was following the official Tor guide, but `adduser` makes more sense in that regard.

PR #554, opened 2020-02-13
Use conf syntax highlighting

interrupt00:
Hi,

Rouge library (a library used by Jekyll for syntax highlighting) has options to highlight config snippets.
I added `conf`, `ini` language specifications for snippets with configurations. Also I added `sh` language specification where it was missed.


Stadicus:
Nice, didn't know about the `ini` and `conf` options.

Stadicus:
Nice, didn't know about the `ini` and `conf` options.

PR #555, opened 2020-02-13
Set blocksonly=1 to speed up IBD

interrupt00:
Hi,

Bitcoin Core has an option to disable the node to listen to and relay transaction.
I set it as an optimization for initial block download.


Stadicus:
Nice. What's your estimate about the performance gain?

interrupt00:
I haven't measured and it is difficult to estimate. We definitely lack IBD analysis.

PR #558, opened 2020-02-21
SSH to your raspibolt using Trezor hardware wallet

danyR85:
I rewrote the HW SSH guide using trezor-agent as a bonus guide.

I wrote the guide for the Trezor because I was able to successfully test on it but Ledger and Keepkey should work as well


Stadicus:
Great addition, thanks! Might change this guide later to use the new OpenSSH U2F funtionality (see https://blog.trezor.io/openssh-with-fido2-and-trezor-e565c2277)

PR #560, opened 2020-02-29
Update 20-raspibolt-welcome

dalamanster:
URL update
(getting reply "301 Moved Permanently" with old one)


Stadicus:
Thanks. The URL `bitnodes.earn.com` still works for me, but your proposed URL is the correct one mentioned in the API docs at https://bitnodes.io/api/#node-status.

PR #564, opened 2020-03-04
Setup SSH keys before fail2ban

interrupt00:
Setup SSH keys access before enabling `fail2ban` to avoid locking yourself. (That is what I did :).

If it is merged I would like to propose to copy keys with `ssh-copy-id` command which deals with folder creation and keys transmission.


Stadicus:
Addressed a minor nit, will merge now. Thanks a lot for this pull request. 

Replacing the SSH key copy with a more standard method would be great.

PR #568, opened 2020-03-06
Clean BTC RPC Explorer instructions

interrupt00:
Minor fixes of  BTC RPC Explorer instructions.


Stadicus:
Thanks for the cleanup. I will remove the `--depth=1` later, as it makes updating more complicated.

PR #585, opened 2020-03-24
Copy keys with ssh-copy-id

interrupt00:


Stadicus:
I know this command from Linux, but cannot verify if it works on MacOS as well. Do you know?

interrupt00:
It should work since `ssh-copy-id` is a part of `OpenSSH`, wich is most likely to be used by MacOS users.
An excerpt of the man page:
```man
SSH-COPY-ID(1)            BSD General Commands Manual           SSH-COPY-ID(1)

NAME
     ssh-copy-id -- use locally available keys to authorise logins on a remote
     machine

SYNOPSIS
     ssh-copy-id [-f] [-n] [-i [identity_file]] [-p port] [-o ssh_option]
                 [user@]hostname
     ssh-copy-id -h | -?

DESCRIPTION
     ssh-copy-id is a script that uses ssh(1) to log into a remote machine
     (presumably using a login password, so password authentication should be
     enabled, unless you've done some clever use of multiple identities).  It
     assembles a list of one or more fingerprints (as described below) and
     tries to log in with each key, to see if any of them are already
     installed (of course, if you are not using ssh-agent(1) this may result
     in you being repeatedly prompted for pass-phrases).  It then assembles a
     list of those that failed to log in, and using ssh, enables logins with
     those keys on the remote server.  By default it adds the keys by append-
     ing them to the remote user's ~/.ssh/authorized_keys (creating the file,
     and directory, if necessary).  It is also capable of detecting if the
     remote system is a NetScreen, and using its `set ssh pka-dsa key ...'
     command instead.

     The options are as follows:

     -i identity_file
...
```

Stadicus:
The command `ssh-copy-id` should be installed on macOS since about 2017. Before that, it needs to be enabled manually first: https://stackoverflow.com/a/43907057/11424365

I'll add a link to this article (https://devtip.io/ssh-keys-remote-server) if users encounter an error.

Stadicus:
Added a commit to your branch, please check. Will merge if you're ok with it.

interrupt00:
Looks good to me. Thanks.

PR #587, opened 2020-03-25
Link signature verification guide

interrupt00:
There is a very good guide from Luke-Jr about GPG and signature verification: https://medium.com/@lukedashjr/how-to-securely-install-bitcoin-9bfeca7d3b2a
I believe it is worth mentioning in the doc.


Stadicus:
Nice link, thanks. When I started this guide there was literally nothing out there, good to see progress.

PR #614, opened 2020-06-07
Replacing apt upgrade with apt full-upgrade, fixing locale error and notice Bitnode not connectable

2pac1:
sudo apt full-upgrade should be better due to the fact that it should install new packages if necessary to satisfy dependencies of upgrading already installed packages.

### DIFF
--- a/raspibolt_20_pi.md
+++ b/raspibolt_20_pi.md
@@ -220,6 +220,16 @@ The following two potential error messages are expected:
   ```sh
   $ sudo dpkg-reconfigure locales
   ```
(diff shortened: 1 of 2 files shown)


2pac1:
Added the fix for locale error that was not going way with the instructions in the guide.
Fix is to stop Pi accepting locale sent via SSH.

Regarding the Update, per official PI's site:
https://www.raspberrypi.org/documentation/raspbian/updating.md

"Note that full-upgrade is used in preference to a simple upgrade, as it also picks up any dependency changes that may have been made."

Stadicus:
Thanks for your contribution, great stuff! :-)

2pac1:
You are more then welcome!

PR #621, opened 2020-06-19
Password protection for the block explorer + electrum pinned server mac app + proxy tab + electrs upgrade

2pac1:
Hey, 

 Apparently you can set a password for the explorer.
On the side of caution, maybe it is better not to share the link publicly.


2pac1:
Hi,

-added info on how to run Electrum on Mac in the form of an .app
-added info on how to make sure you get direct connection to your node in case you have been previously using TOR under Proxy tab in Electrum
-added section on how to proceed with updating Electrs if you have files that are changed locally

raspibolt_50_electrs.md
```diff
@@ -426,6 +427,17 @@ You can force Electrum to only use your own server on startup with the following
...
 ```
+BONUS INFO:
+On Mac, you can use Script Editor to automate this and create a new App with the following content:
+```sh
+tell application "Terminal"
+	activate
+	do script "/Applications/Electrum.app/Contents/MacOS/run_electrum --oneserver --server Raspibolt-IP-Goes-Here:50002:s"
+end tell
+```
+This "app" will basically do same thing with Electrum 4.0.2 as if you took the command line from above that is specific to Electrum's location on Mac and just ran it in your Terminal.
```
lucasmoten, line 438: Regarding `This "app" ...` I'm not familiar with Macs, but can you verify whether this is the correct terminology?  Is this really an "app", or is it an "Apple shortcut" or something else? I think it would be best if we use whatever the standard nomenclature, and then drop the quotes.

raspibolt_50_electrs.md
```diff
@@ -508,6 +520,10 @@ Updating a [new release](https://github.com/romanz/electrs/releases){:target="_b
   $ sudo cp ./target/release/electrs /usr/local/bin/
   $ sudo systemctl start electrs
   ```
+BONUS INFO:
+When installing Electrs, you might notice alert in terminal warning you about insecurities in some of the files that are being downloaded.
```
lucasmoten, line 524: This whole bonus section still worries me.  We don't want to end up training people to ignore any security concerns.  Does this actually happen on version 4.0.2?

raspibolt_50_electrs.md
2pac1: It has an .app extension.
The same way any other program has .app extension.
It is done in script editor where when you select export > application > you get a file.app
It might be misleading in terms of "We coded something to be an app with few lines" but again, if it has .app I would call it that especially if Script Editor where it is done calls it Application.

raspibolt_50_electrs.md
2pac1: Will report back soon. 
Haven't split all the pulls and now all is bundled into one. Grrr

Kixunil: What insecurities? Installing `electrs` does not report any insecurities. I guess you saw something else and got confused?

Stadicus:
Thanks for your contribution. This pull request was resolved with the update to RaspiBolt version 3.

PR #629, opened 2020-07-11
Update disk size recommendations

lucasmoten:
Reference https://github.com/Stadicus/RaspiBolt/issues/599 regarding 500GB not being large enough for full blockchain and electrs database before compaction.


Stadicus:
Thanks!

Issue #638, opened 2020-07-25
WireGuard VPN setup for RaspiBolt


Pantamis:
I installed WireGuard VPN on my RaspiBolt to access it from outside without openning more than one port of my LAN. I think it is better for security and makes easier to run many more services on the Pi (no more Tor hidden services or port forwarding) and it also gives users a good security to use their node or surf on the web from public networks.

WireGuard is easy to install from unstable debian repo and the hardest part is writing the right AllowedIP in configuration file. It also have a very nice and simple app on smartphone to setup the vpn and enjoy support on all computer OS, even windows.

There are trade-off using a DIY VPN against Tor, particulary anonymity against LAN security and bandwidth. Moreover it is still experimental (but soon in linux kernel) so maybe not safe enough to include it in RaspiBolt guide. They are also many others guides to setup WireGuard on a Pi on the web. But it could be nice to have the setup in your guides at least when it will be in linux kernel.

I am curious to hear what you think about it.

Pantamis:
I submitted a PR #644 with commands and explanation, it should be fairly complete.
I would be happy to have your opinion on it !

PR #644, opened 2020-08-25
WireGuard VPN setup for RaspiBolt

Pantamis:
Proposition from issue #638

I added information about VPN in the guide for people to understand the use-case and trade-off of a VPN. I detailed how configuration files work which makes the guide fairly long, but I expect that just copy-pasting command lines should work (except private-public keys).

I didn't understand how table of content format works but I tried to make a one consistent with what I see in other files.

Maybe you will find this not relevent here, I would be happy to find another place to share it.

Thanks for your suggestion !

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -9,3 +9,5 @@ aux_links:
 
 # enable theme for local usage
 #theme: "just-the-docs"
+
+theme: jekyll-theme-minimal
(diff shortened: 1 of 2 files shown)


Pantamis:
Just squashed the commits for easier review

asvdf:
Been using Wireguard for the last few years on my usual home server, I do own several raspi's so that makes it easier to have them share the load, and yet I would be partial to install wireguard on another device without loading the raspi node with things that are not strictly LN related.
In any case, a wireguard installation will work the same way no matter where you do install it, so if your node is the only home server you own, it's still a useful no-strings-attached improvement to have it run Wireguard as well, as opposed to accessing it via Tor.

PR #649, opened 2020-10-10
Run The Numbers

lucasmoten:
A supplemental/bonus script and service to periodically Run The Numbers.

Based on https://github.com/BashCo/RunTheNumbers, but tailored to the Raspibolt environment.

Run The Numbers is part of a synchronized decentralized audit of the Bitcoin supply.  A target block height is agreed upon whereby many participants will simultaneously run a similar script on their node.  This version is setup to automatically run as a service, and by default, run the numbers every 5000 blocks, or about once every 34 to 35 days based on a target of 10 minute block times.  Results are saved in the /home/bitcoin/.runthenumbers folder. Instructions for reviewing the results are also outlined within.

---

Known Issues: Documentation -- Modifications will probably be needed within the Just the Docs configuration in the markdown.

### DIFF
--- a/raspibolt_78_runthenumbers.md
+++ b/raspibolt_78_runthenumbers.md
@@ -1,0 +1,148 @@
+---
+layout: default
+title: Run the Numbers
+nav_order: 78
+---
+<!-- markdownlint-disable MD014 MD022 MD025 MD033 MD040 -->
+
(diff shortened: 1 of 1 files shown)


lucasmoten:
Going to retract this for now and turn it into a more fully functional service

PR #584, opened 2020-03-24
Update raspibolt_72_zeus-over-tor.md

csh2000:
Hello, connection with Tor v3 is much easier, as you do not need Client Cookies for Orbot.

Followed Openoms guide and adapted it to Raspibolt II, https://github.com/openoms/bitcoin-tutorials/blob/master/Zap_to_RaspiBlitz_through_Tor.md

Thanks!

PR #650, opened 2020-10-11
Update raspibolt_72_zeus-over-tor.md

lucasmoten:
Use newer version of Go.
- Go 1.11 has reached end of life. Updating guidance to use Go 1.15.2

Don't remove all .gz files in the current folder, only those that we downloaded

Use newer version of lndconnect
- v0.2.0 was released November 2019, and includes changes since initial v0.1.0 release in January


lucasmoten:
Closing. I missed #584 which is a better approach and has most of the updates referenced.

PR #654, opened 2020-10-20
There is a new primary release signing key for LND v0.11.1

kristapsk:
roasbeef's key now is secondary, see https://github.com/lightningnetwork/lnd/releases/tag/v0.11.1-beta

Verify both signatures.


lucasmoten (approved):
Checks out.  Also appreciate the retrieval of the pgp_keys saving as different filename to reduce ambiguity.

Stadicus:
Great, thanks a lot!

PR #666, opened 2020-11-29
Fix installed electrs version request in raspibolt welcome script

bmml1:
Fix https://github.com/Stadicus/RaspiBolt/issues/663#issue-751544237

### DIFF
--- a/resources/20-raspibolt-welcome
+++ b/resources/20-raspibolt-welcome
@@ -329,7 +329,7 @@ if [ -z "${electrs_running##*inactive*}" ]; then
   electrsversion_color="${color_red}"
 else
   electrs_running="up"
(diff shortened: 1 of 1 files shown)


lucasmoten:
I'm not sure what this is intended to do. As this is sending parameters in the request I get the same result whether it includes `"params": [ "raspibolt", "1.4" ],` or not with 0.8.5.

I'm in the process of upgrading to 0.8.6 with documented steps now and will check again and report back once that is complete.

lucasmoten (approved):
Code Review ACK
Confirmed that this works with both Electrs 0.8.5 and the recently released 0.8.6

Stadicus:
Thanks for the fix @bmml1 & the review @lucasmoten, much appreciated!

PR #668, opened 2020-12-02
Electrs update to newest version

lucasmoten:
Primary install now dynamically determines latest release version from repository tags
FAQ now includes a section on how to upgrade Electrs. Note that each version backup is about 350-400 MB.

### DIFF
--- a/raspibolt_50_electrs.md
+++ b/raspibolt_50_electrs.md
@@ -91,11 +91,12 @@ The whole process takes about 30 minutes.
   ```sh
   # download
   $ cd /home/admin/rust
-  $ git clone --branch v0.8.5 https://github.com/romanz/electrs.git
+  $ electrsgit=$(curl -s https://api.github.com/repos/romanz/electrs/tags | jq -r '.[0].name')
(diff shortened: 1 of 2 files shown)


2pac1:
This is good idea regarding getting the latest version, without the need for guide to be updated right away when the new version is released.
Or the need for a user to check manually to see if the latest version available is the one mentioned in the guide.

Stadicus:
I usually try to avoid too complex logic that is not obvious to command line newcomers. But I see how this improves the overall installation flow, so let's give this a try :)

Stadicus:
Tested, works like a charm. I'll also update to the latest Rust version in a separate commit. Thanks!

2pac1:
We could change text "specific release" to: always latest release fetched
with this more complex command

Stadicus:
Thanks for the suggestion, changed phrasing to "the latest release" in https://github.com/Stadicus/RaspiBolt/pull/673/commits/4a59a8e541f4cb34c67f2cb4289385cd54271455

2pac1:
Msg that "You are in 'detached HEAD' state" is expected should be put in there, when using the new style of getting the latest version.

PR #690, opened 2021-01-17
Update raspibolt_50_electrs.md

lucasmoten:
Uses same guidance from [FAQ section](raspibolt_faq.md#how-to-upgrade-electrs) to check current vs available version of Electrs and not hard code version in documentation


Stadicus:
Thanks!

PR #661, opened 2020-11-24
fix LND gpg keys download

Stadicus:
Commit 8e3810afb6c31c1f27dcbb0c69983d6b9101f10a changed how the PGP keys of LND maintainers are downloaded. The commands are not correct and use a lowercase `-o` instead of param `-O`.

This commit

* changes `wget -o` to `wget -O` to get the plaintext PGP keys
* fixes #660

PR #693, opened 2021-01-20
Update raspibolt_40_lnd.md

umuttuncer:
fix for wget download options. (using -o not working needs to be upper case -O)
fix for gpg typo (pgp to gpg)


lucasmoten:
#661 accomplishes the same fix.

PR #703, opened 2021-02-28
Moving from rpcuser/rpcpassword to cookie_file authentification/rpcauth

luneth:
Using current configuration recommendation when starting bitcoind the following message appear:
`Config options rpcuser and rpcpassword will soon be deprecated. Locally-run instances may remove rpcuser to use cookie-based auth, or may be replaced with rpcauth. Please see share/rpcuser for rpcauth auth generation`

The following modification allows us to use rpcauth using rpcuser.py script available at [bitcoin/bitcoin](https://github.com/bitcoin/bitcoin/blob/master/share/rpcauth/rpcauth.py)
Once rpcauth is set up and the rpcpassword line replaced by rpcauth bitcoind creates a .cookie cookie authentication file at startup. (Actually when no rpcpassword is set in bitcoin.conf bitcoind creates a .cookie file when the service starts et deletes it at exit)
We use this method of authentication instead of the deprecated rpcuser/rpcpassword one to authenticate Electrs to our node.

Finally, as btcrpcexplorer does not have access permission to the cookie authentication file we use rpcuser/rpc password but this time using the password created by rpcuser.py script as recommended.

More info [here](https://bitcoin.org/en/release/v0.12.0#rpc-random-cookie-rpc-authentication) and [here](https://bitcoin.stackexchange.com/questions/46782/rpc-cookie-authentication).

EDIT:
The .cookie file is created with umask 0022 setting its permission to 600 -rw------. Making it unreadable to admin user. We could add In the service unit the ```UMask=0027``` value and ```sysperms=1``` in bitcoin.conf so the default system permission is 640. 

EDIT2:
We could give access permissions to the .cookie file for btcrpcexprorer user adding it the the bitcoin group. Thoughs?

### DIFF
--- a/raspibolt_30_bitcoin.md
+++ b/raspibolt_30_bitcoin.md
@@ -106,8 +106,19 @@ Instead of creating a real directory, we create a link that points to a director
 
 ### Configuration
 
+First we neet to generate a password for rpcpassword:
+
+```wget https://raw.githubusercontent.com/bitcoin/bitcoin/master/share/rpcauth/rpcauth.py
+   $ python ./rpcuser.py raspibolt
+     String to be appended to bitcoin.conf:
+     rpcauth=raspibolt:[hashed password]
+     Your password:
+     [rpcpassword]
+
+Keep note of the rpcpassword, needed for block explorer
+
 Now, the configuration file for bitcoind needs to be created.
-Still as user "bitcoin", open it with Nano and paste the configuration below. Save and exit.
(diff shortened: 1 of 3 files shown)


Stadicus:
Thanks @luneth, and sorry for not merging this earlier. By moving this guide to https://github.com/raspibolt I hope I can get rid of myself as a bottleneck. 

I overhauled the RPCAUTH part myself in PR https://github.com/raspibolt/raspibolt/pull/778 as part of my general guide overhaul. I opted to keep the [ Password B ], as I think it's easier overall. I will close this PR, but would like to thank you nonetheless for the work you put in. It's much appreciated.

PR #711, opened 2021-03-25
Recommend creating password file BEFORE writing

Kixunil:
This changes the recommendation to write the password and `chmod` the file afterwards to creating an empty file with correct permissions first and writing the password into it afterwards. This avoids theoretical attack of a malicious app seeing the password between writing it and protecting the file.

Note that this whole exercise, especially setting the owner as root, may be pointless because any process that can access TLS private key or run under the same user as lnd can steal password/sats no matter what kind of unlocking mechanism is used. Protecting lnd in such scenarios is physically impossible. And if you assume these scenarios will not happen, then it makes sense to assume `pwd` file will not be accessed by other user than the one running `lnd`, so in conclusion `pwd` file **may** be owned/accessible by the user running `lnd` without any loss of security. Running the unlocker script under same user may be preferred as it doesn't endanger the rest of the system.

I will suggest to reword the introduction in the followup PR. I felt it's better separating them.

### DIFF
--- a/raspibolt_6A_auto-unlock.md
+++ b/raspibolt_6A_auto-unlock.md
@@ -18,8 +18,10 @@ This is why a script that automatically unlocks the wallet is  helpful. The pass
 `$ systemd --version`
 
 
-* As user "admin", create a new directory and save your LND wallet password [C] into a text file
-  `$ sudo mkdir /etc/lnd`
+* As user "admin", create a new directory, prepare it and save your LND wallet password [C] into a text file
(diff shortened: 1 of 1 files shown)


Stadicus:
Great, thanks! IIUC, LND will add a native auto-unlock feature in the upcoming v0.13, so that this section could be retired at some point in the future.
https://github.com/lightningnetwork/lnd/releases/tag/v0.13.0-beta.rc5

Kixunil:
Yeah, I was the one who pushed for that change in lnd. :) Let's wait for the release with those changes then.

PR #717, opened 2021-04-12
Add Specter Desktop steps

vindard:
## Description

This PR adds a new guide under the **Bonus** section with the steps for setting up Specter Desktop on the Raspibolt and then accessing it via browser from the local LAN.

I wasn't sure about how best to work this into the Jekyll site, so I did some assumed configs to the top of the new md file.

_More info about Specter Desktop here: [Specter github repo](https://github.com/cryptoadvance/specter-desktop)_

image


lucasmoten (approved):
Reviewed. Untested ACK

Stadicus:
Thanks, great addition. Will give it a spin to test as soon as I can!

VajraOfIndra:
Regarding adding `disablewallet=0` to the bitcoin.conf file: afaik a wallet has to be created and loaded to use JoinMarket on the Raspibolt which is in the Bonus section of the guide. It might mean the 2 programs cannot be installed on the Raspibolt at the same time, unless there is a way to avoid this apparent conflict? If not, a warning message might be needed in both guides.

PR #718, opened 2021-04-14
Enable Wumbo channels and Anchor channels for LND

lucasmoten:
Enable Anchor Commitments and Wumbo Channels

See https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf for all lnd.conf options

; If set, lnd will use anchor channels by default if the remote channel party
; supports them. Note that lnd will require 1 UTXO to be reserved for this
; channel type if it is enabled.
; protocol.anchors=true

; If set, then lnd will create and accept requests for channels larger than 0.16
; BTC
; protocol.wumbo-channels=true


Stadicus:
Thanks!

PR #727, opened 2021-05-06
Small fixes

TKone7:
Parameter `-o` (lowercase) redirects the log output into a file, while `-O` (uppercase) stores the document to the file (what is wanted here).


Stadicus:
Good catch, thanks!

Issue #737, opened 2021-06-23
Missing macaroons when upgrading Lightning Terminal (to v0.5.0)


VajraOfIndra:
I had Lightning Terminal v0.4.1 installed on the Raspibolt (following this guide: [https://github.com/Pantamis/RaspiBolt/blob/LiT-install/raspibolt_40_lnd.md](https://github.com/Pantamis/RaspiBolt/blob/LiT-install/raspibolt_40_lnd.md)).

I upgraded to the latest version (v0.5.0) following the guideline above, which refers to the LND upgrade guide: [https://stadicus.github.io/RaspiBolt/raspibolt_faq.html#how-to-upgrade-lnd](https://stadicus.github.io/RaspiBolt/raspibolt_faq.html#how-to-upgrade-lnd). The guide contains this section:

> Remove old stuff `sudo rm /home/bitcoin/.lnd/data/chain/bitcoin/mainnet/*.macaroon`, then download, verify and install the latest LND binaries as described in the Lightning section of this guide.

After restarting litd, litd fails to start.

The journal log (sudo journalctl -f -u litd) gives:

>     Jun 23 15:11:39 raspibolt systemd[1]: Started LND Lightning Network Daemon.> 
>     Jun 23 15:11:40 raspibolt litd[10527]: Attempting automatic RPC configuration to bitcoind> 
>     Jun 23 15:11:40 raspibolt litd[10527]: Automatically obtained bitcoind's RPC credentials> 
>     Jun 23 15:11:40 raspibolt litd[10527]: 2021-06-23 15:11:40.184 [INF] LTND: Version: 0.13.0-beta commit=lightning-terminal-v0.5.0-alpha, bu ild=production, logging=default, debuglevel=info> 
>     Jun 23 15:11:40 raspibolt litd[10527]: 2021-06-23 15:11:40.185 [INF] LTND: Active chain: Bitcoin (network=mainnet)> 
>     Jun 23 15:11:40 raspibolt litd[10527]: 2021-06-23 15:11:40.188 [INF] RPCS: RPC server listening on 0.0.0.0:10009> 
>     Jun 23 15:11:40 raspibolt litd[10527]: 2021-06-23 15:11:40.188 [INF] LITD: Dialing lnd gRPC server at 127.0.0.1:10009> 
>     Jun 23 15:11:40 raspibolt litd[10527]: 2021-06-23 15:11:40.192 [INF] RPCS: gRPC proxy started at 0.0.0.0:8080> 
[...]

PR #738, opened 2021-06-23
Add installation of LiT in Lightning section

Pantamis:
Add a section to install LiT instead at the end of LND installation guide

Much easier to install than having to install loop and pool in standalone so it should be straightforward to review !

(please give some love to my other PR #644 too, I really have a great experience with WireGuard VPN on my RaspiBolt instead of Tor)

### DIFF
--- a/raspibolt_40_lnd.md
+++ b/raspibolt_40_lnd.md
@@ -512,4 +512,229 @@ If you want to upgrade to a new release of LND in the future, check out the FAQ 
 
 ---
 
+## Optional: Using LiT (Lightning Terminal) instead of LND
+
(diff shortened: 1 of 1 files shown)


VajraOfIndra:
tACK
A warning in the upgrade section that the LND macaroons should not be deleted (as suggested by the LND upgrade section of the Raspibolt guide, to which this LiT upgrade guide refers to) might be necessary (see Issue #737).

Pantamis:
Thanks for testing @VajraOfIndra :)

I add a warning in bold about not deleting the LND macaroons and also the needed symbolic link to the directories of other packaged software in LiT so that `admin` user can use them in CLI too !

Maybe worth to add the alias in `bitcoin` user's `.bashrc` too tho

I can squash the commits before merge if needed, just ask.

Stadicus:
utACK, thanks a lot!

PR #751, opened 2021-08-09
Update raspibolt_73_static_backup_dropox.md

VajraOfIndra:
1. Corrected typo channels.backup to channel.backup
2. The Dropbox app creation webpage has changed since the guide has created. I updated the section (both instructions and pictures) to match the new layout and options of the webpage.
    Note 1: The pictures are hosted on https://postimages.org/... maybe this is not optimal and a better hosting location and URL could be used?
    Note 2: Dropbox mentions that the no-expiraton token is less secure than the short-lived token and will be deprecated in the future. Is it possible to use the short-lived token for our use case?
3. In the 'Preparation on the Raspibolt, there is no /home/admin/download/ folder so the first command gives an error message. Updated the command to use the /tmp/ folder as is used for the 'System overview' script.
4. Added a note about placing the API token string within  the double quotes to avoid confusion about using them or not.
5. Added a 'sudo' for the journal monitoring command

### DIFF
--- a/raspibolt_73_static_backup_dropox.md
+++ b/raspibolt_73_static_backup_dropox.md
@@ -13,45 +13,33 @@ This guide explains one way to automatically upload the channels.backup file on 
 The following scripts were created by [Vindard](https://github.com/vindard)
 
 #### *Risk : Minimal* 
-The channels.backup file is encrypted so that it is safe to transmit over the Internet and to store on (e.g.) a cloud server.
+The channel.backup file is encrypted so that it is safe to transmit over the Internet and to store on (e.g.) a cloud server.
 
 #### *Requirements: lnd version higher than 0.6*
-
-
 
 ### Setup Dropbox API Key
     
 In your web browser, do the following:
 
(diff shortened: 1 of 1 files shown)


VajraOfIndra:
Regarding Note 2 above, Dropbox announced that as of September 30th, the Dropbox OAuth flow will no longer return new long-lived access tokens. From their newsletter:
> Action required: tokens and permissions
> As a reminder, on September 30th, 2021, the Dropbox OAuth flow will no longer return new long-lived access tokens. It will instead return short-lived access tokens, and optionally return refresh tokens. Existing tokens are not impacted.
> Apps that require background access will need to update their code to use refresh tokens, which is made easier by using our updated SDKs. All other apps should ensure that users are directed to re-authorize upon token expiration.
> The Dropbox API now supports more granular permission scopes. While you review your application’s OAuth flow, be sure to check for scopes that may not be required.

Can we use the short-access token rigth now or does the script running on the Raspibolt needs to be updated first?

PR #754, opened 2021-08-18
Bonus installation guide: Balance of Satoshis

VajraOfIndra:
- An optional guide to install Balance of Satoshis (bos).
- This guide is a modified version of Openoms manual installation guide for the Raspiblitz (see acknowledgments).
- Uncertain about the section regarding Node JS upgrade, is there a better/safer way to do it maybe?
- The nav-order (at the top) has not been updated (130), I'm not sure how this works and what should be the value for this new guide.


Stadicus:
Thanks for the bonus guide. You stop & disable the BTC RPC Explorer. What's the exact reason? Also, it should be enabled & started again after installation

VajraOfIndra:
> Thanks for the bonus guide. You stop & disable the BTC RPC Explorer. What's the exact reason? Also, it should be enabled & started again after installation

I don't know how updating NodeJS would affect, or not, the explorer, so I assumed that it would be safer to stop the explorer... but maybe it's not required at all? In that case, I could just remove the 2 lines that stop and disable the btcrpcexplorer service.

VajraOfIndra:
@Stadicus: See comment above and latest commits. I decided to remove the commands that stop and disable the BTC RPC Explorer as it is probably unnecessary. I checked the NodeJS update script of the Raspiblitz and they do not seem to stop it either.

acburkhardt:
would it be worth adding a disclaimer that the node's IP is exposed when communicating with the bot? and/or including the option to set `AllowOutboundLocalhost 1` in torsocks.conf and then `torify bos telegram` according to Openoms' suggestion [here](https://github.com/alexbosworth/balanceofsatoshis/issues/54#issuecomment-678791066)?

is there a benefit to keeping the bot alive with tmux rather than a system service?

VajraOfIndra:
> is there a benefit to keeping the bot alive with tmux rather than a system service?

Probably not, I just used tmux because I was not sure what was the best way to do it. You and a Redditor both suggested to use a systemd service as it would allow to autmatically restart the command when rebooting so seems superior to tmux for sure. I used your linked guide as inspiration and adapted it to the Raspibolt style/format. Let me know what you think of the new section.

> would it be worth adding a disclaimer that the node's IP is exposed when communicating with the bot? and/or including the option to set `AllowOutboundLocalhost 1` in torsocks.conf and then `torify bos telegram` according to Openoms' suggestion [here](https://github.com/alexbosworth/balanceofsatoshis/issues/54#issuecomment-678791066)?

Adding a disclaimer is a great idea. Just for me to understand, who is getting the node IP address when using the TG bot, is it the Telegram server, i.e. Telegram? or could someone else get the IP address as well? Maybe we could add a disclaimer at the start of the TG bot guide such as: 

> Privacy warning: Using the Telegram bot will leak your node IP address to Telegram.

Regarding using `torify`, did you try this yourself? does it work well? Is there any downsides to using this and is there any risk associated with setting `AllowOutboundLocalhost` to 1? (I don't fully understand what it does so I don't grasp if there is any associated risks).
Should `torify` or `torsocks` be used? 
Finally, if we want to use this, how the service file should be modified to torify the bos command..? (I suppose it's a modification of the `ExecStart=` line but I don't know how to modify it)
I don't know enough about torify to write this optional section. Would you have time to write it?

Thanks for the feedback!

acburkhardt:
> Let me know what you think of the new section.

Looks great!

> Just for me to understand, who is getting the node IP address when using the TG bot, is it the Telegram server, i.e. Telegram?

AFAIK the bot server would know your node's IP and I guess that anyone snooping on your outbound traffic would see that connection too (as opposed to an outgoing tor connection).

> Regarding using torify, did you try this yourself?

I've been using the suggestion in the linked comment for a while with no issue; all bot commands work.

> Is there any downsides to using this and is there any risk associated with setting AllowOutboundLocalhost to 1?

I also don't really grasp the security impact. I just understand that it allows torify/torsocks to proxy outbound connections through tor and I can only find general warnings that it might weaken security. I've experienced no issues but can't really advise

> Finally, if we want to use this, how the service file should be modified to torify the bos command..? I suppose it's a modification of the ExecStart=

Exactly! `ExecStart=/usr/bin/torify /home/bos/.npm-global/bin/bos telegram --connect <CONNECTIONCODE>`

Either way, you can set a budget in sats when starting the bot if you think it might be worth adding as an optional safety step:
 
 `ExecStart=/home/bos/.npm-global/bin/bos telegram --connect <CONNECTIONCODE> --budget 200000`
 
The bot can only spend up to the budget until the service is restarted—you can make a cron job if you want to restart at intervals:
- Edit sudoers file and add a line to the bottom to allow admin to restart the bot service with no password.

  ```sh
  $ sudo visudo
  ```
  ```
  admin ALL=NOPASSWD:/usr/bin/systemctl restart bos-telegram.service
  ```

- Edit admin user's crontab and add a line to restart the service whenever you want.

  ```sh
  $ crontab -e
  ```
  ```
  # Example: 5am everday
  0 5 * * * sudo systemctl restart bos-telegram.service
  ```

asvdf:
Having bos installed on a node is not really optional as of today, so I strongly vote for this to be added on the official guide.
Plus, newcomers will just love to get realtime updates on telegram, so that alone is a selling point.

VajraOfIndra:
Closing as outdated. Will open a new PR compatible with v3 and following the new PR template.

PR #752, opened 2021-08-10
Update raspibolt_40_lnd.md (many updates of the lnd/litd guide)

VajraOfIndra:
Summary: the updates aim at:
1. using the latest lnd and litd releases,
2. correcting typos and broken links,
3. adding useful options to the lnd/litd configuration files,
4. replacing outdated commands and examples and warning for outdated videos, and;
5. adding a few safety information for some commands.

The full list of changes and rationales for them:
1. Added warning above most videos: (Warning: the video below is outdated and does not correspond exactly to the commands above)
2. Updated the 'download and install' section with latest release (v0.13.1). Removed bitconner-related lines as not used for the verification process.
3. Updated lnd.conf content 
4. Removed protocol.anchor=true (now by default) (Fixes #748 )
5. Added AMP payments options (did not add keysend as AMP renders keysend obsolete with time)
6. Made minchansize option visible and changed it from default (20,000) to 100,000 sats
7. Made base fee and fee rate settings visible (kept default value)
8. Modified the default value of coop-close-target-confs to save on closing fees
9. Added the startup speed section from https://www.lightningnode.info/advanced-tools/lnd.conf (by Openoms)
10. Removed the 'sudo su - bitcoin' user login at the start of section "Run LND": not needed as we already are with user bitcoin after the lnd.conf section.
11. Removed "and admin.macaroon" when veryfing the TLS certs in folder /mnt/ext/lnd/
12. Added commands to allow safe reboots of the node (line 373)
13. Replaced the np2wkh address creation for funding by a bech32 p2wkh address
14. Replaced the nested segwit full address by the string bc1... (to avid risk of funds being sent to example address)
15. Added 'optional' to the autopilot mention as it is not by default
16. Added amboss.space together with 1ML as amboss despite being newer seems to be more actively developped than 1ml.
17. Default node to demonstrate lncli connect and lncli openchannel is down, replaced by ACINQ which should be a stable long-lasting node and useful for routing.
18. Removed the lncli openchannel command that does not use the optional fee selection --sat_per_byte (leads to needlessly overpaying fees)
19. Replaced deprecated sat_per_byte with preferred sat_per_vbyte in the openchannel command
20. Broken link converter satoshi to USD, fixed
21. Add comment and link to check mempool.space to select appropriate fee
22. In "More commands", added an example on how to send an invoice-less payment to a node using AMP
23. Added sat_per_vbyte option to cloeschannel command
24. Ask for selection of a new passwod [E] for the Lightning Terminal UI password
[...]

PR #756, opened 2021-08-20
anchors option deprecated

MG-ng:
I assume that `protocol.no-anchors=false` is the equivalent to the option `protocol.anchors=true`. 

I changed it because the protocol.anchors attribute is not supported anymore. 

See here [https://github.com/lightningnetwork/lnd/blob/951e2164fdbc77cac604881d689177a179577261/sample-lnd.conf#L1070](https://github.com/lightningnetwork/lnd/blob/951e2164fdbc77cac604881d689177a179577261/sample-lnd.conf#L1070)

### DIFF
--- a/raspibolt_40_lnd.md
+++ b/raspibolt_40_lnd.md
@@ -97,7 +97,7 @@ Now that LND is installed, we need to configure it to work with Bitcoin Core and
   debuglevel=info
   maxpendingchannels=5
   listen=localhost
-  protocol.anchors=true
+  protocol.no-anchors=false
   protocol.wumbo-channels=true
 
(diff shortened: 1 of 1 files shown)


VajraOfIndra:
In my PR #752, I removed the deprecated option. 
But having the new property explicitly written down and set to false is actually probably a better way to do it, as it allows the users to see that they could change this option if they don't want to use anchor channels.

MG-ng:
Okay, your Pull Request addresses way more aspects, that need also be done… I'll close this one

VajraOfIndra:
I added the property protocol.no-anchors=false to PR #752

PR #767, opened 2021-10-10
Fix Electrs config file

asierba:
First of all thanks a million for this amazing guide! It was quite useful for me to set-up a bitcoin node from scratch 😄 

I've downloaded latest electrs server (0.9.0) and wasn't working for me, also with v0.8.12. The issue is with the `cookie` variable in the config, should be `auth` instead.
That worked for me.
More details here: https://github.com/romanz/electrs/blob/master/doc/usage.md#configuration-files-and-priorities


asierba:
true! sorry I didn't see that one, closing this PR :)

PR #776, opened 2021-11-04
Monitoring channel.db size in welcome/motd script

VajraOfIndra:
**Goal:** This pull-request adds a few line to the `20-raspibolt-welcome` script to display the `channel.db` size in the LND section of the 'message of the day' (motd) system overview.

**Rationales:** When `channel.db` reaches ~1GB on 32-bit OS, LND experienced issues (e.g. [here](https://github.com/lightningnetwork/lnd/issues/4811)). Issues can also be met on 64-bit OSes when reaching larger sizes on low-spec devices (e.g. [here](https://github.com/lightningnetwork/lnd/issues/5705)). Therefore, monitoring the growth of the `channel.db` size is critical.

 **Ideas for future improvements:** For now the variable is a string. Getting an integer instead would allow to create a simple conditional statement to colour-code the line in green, orange and red when below  or above a certain size (e.g. <500MB green, 500-900MB orange, >900MB red)

### DIFF
--- a/resources/20-raspibolt-welcome
+++ b/resources/20-raspibolt-welcome
@@ -323,7 +323,7 @@ else
 fi
 
 #get channel.db size
-channel_db_size=$(du -h /mnt/ext/lnd/data/graph/mainnet/channel.db | head -c4)
+channel_db_size=$(du -h /mnt/ext/lnd/data/graph/mainnet/channel.db | awk '{print $1}')
 
 #electrs
 electrs_running=$(systemctl is-active electrs)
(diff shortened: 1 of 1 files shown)


asvdf:
I use a custom script to tell me the size of channel.db, welcome script would allow not-so-experienced users to just know that at a glance without the hassle of a `ls -alh` command in the shell

VajraOfIndra:
Closing this PR as it has become outdated. I'll open a new PR for the same bonus guide but that will use the new PR template and be compatible with v3.

PR #778, opened 2021-11-07
Update bitcoin rpc auth

Stadicus:
bitcoin: update rpc authentication

The rpcuser/rpcpassword method to authenticate RPC requests is
outdated. This commit explains how to switch to the rpcauth
method using a hashed password. Bitcoin Core will now also
create an authentication cookie, that can be used by other
serivces like ELectrs.

electrs: update authentication method

Together with the update of Bitcoin Core RPC authentication, this
commit removes the RPC user/password authentication and uses the
Bitcoin Core cookie authentication instead.

### DIFF
--- a/raspibolt_30_bitcoin.md
+++ b/raspibolt_30_bitcoin.md
@@ -28,7 +28,7 @@ After validation, the client can check all future transactions whether they are 
 
 The validated blocks are also the base layer for other applications, like Electrs (to use with hardware wallets) or LND (the Lightning Network client).
 
-Be already warned that the downloading and validation of all transactions since 2009, more than 600'000 blocks with a size of over 300 GB, is not an easy task.
+Be already warned that the downloading and validation of all transactions since 2009, more than 600'000 blocks with a size of over 400 GB, is not an easy task.
 It's great that the Raspberry Pi 4 can do it, even if it takes a few days, as this was simply not possible with the Raspberry Pi 3.
 
 ---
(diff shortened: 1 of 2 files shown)


raspibolt_30_bitcoin.md
```diff
@@ -37,41 +37,51 @@ It's great that the Raspberry Pi 4 can do it, even if it takes a few days, as th
...
   ```sh
-  cd /tmp
+  $ cd /tmp
   ```
 
-* Get the latest download links at [bitcoincore.org/en/download](https://bitcoincore.org/en/download){:target="_blank"} (ARM Linux 32 bit), they change with each update.
-  Then run the following  commands (with adjusted filenames) and check the output where indicated:
+* Get the latest download links at [bitcoincore.org/en/download](https://bitcoincore.org/en/download){:target="_blank"} (ARM Linux 64 bit), they change with each update.
 
   ```sh
```
VajraOfIndra, line 53: There are commands to download the binary hashes and signatures but no command to download the binary itself... just wondering if that's a typo or intended.

raspibolt_30_bitcoin.md
```diff
@@ -37,41 +37,51 @@ It's great that the Raspberry Pi 4 can do it, even if it takes a few days, as th
...
   # download Bitcoin Core binary
-  $ wget https://bitcoincore.org/bin/bitcoin-core-22.0/bitcoin-22.0-arm-linux-gnueabihf.tar.gz
+
```
VajraOfIndra, line 55: There are commands to download the binary hashes and signatures but no command to download the binary itself... just wondering if that's a typo or intended.

Stadicus: Thanks, good catch! I already updated this line to 64bit and didn't want to commit the change. Somehow I deleted the original line anyway. Fixed with lastest update.

Stadicus:
I also added `whitelist=download@127.0.0.1` to `bitcoin.conf`, otherwise Electrs runs into issues when syncing.

PR #780, opened 2021-11-08
Bonus installation guide: charge-lnd (with cron job, for automated fee policy management)

VajraOfIndra:
**Goal:** This is a bonus guide that explains how to install and set up charge-lnd, a simple policy-based fee manager for LND, together with a cron job to run it regularly and automate as much as possible the fee policy management.

**Rationales:** This guide belongs to a series of bonus installation guides to help the node operator to manage its LN node and channels (e.g. Ride The Lightning, Balance of Satoshis etc). Charge-lnd is a popular program used by the LN community to automate the management of channel fee policies. The guide shows how to install charge-lnd, gives an example of config file (i.e. a set of policies) and explains how to create a cron job a a log file to regularly run the program (every 12 hours to avoid spamming the gossip network) and how to upgrade and uninstall it.

### DIFF
--- a/raspibolt_80_charge-lnd.md
+++ b/raspibolt_80_charge-lnd.md
@@ -2,7 +2,7 @@
 layout: default
 title: charge-lnd
 parent: Bonus Section
-nav_order: 130
+nav_order: 140
 has_toc: false
 ---
 # Bonus guide: Charge-lnd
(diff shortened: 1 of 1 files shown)


asvdf:
there are discording opinions on the opportunity of using proportional fees based on local balance ratio, but one thing that charge-lnd is very good at, and is also extremely useful, is to apply 9999basefee policy when local is depleted.
In that case, a 12hrs cronjob would be quite ineffective. I'd say that, if you plan to use proportional fees, anything over 2 hours is fine. If instead you plan to use it only to stop flow with 9999basefee when local is drained, then I'd have it run every 15 minutes.

VajraOfIndra:
Thanks for the feedback @asvdf!

> there are discording opinions on the opportunity of using proportional fees based on local balance ratio
Absolutely. That's why I opted to use only static fee policies in this tutorial (in this case a simple dual policy of low and high static fees, on top of the 9999basefee policy) and did not mention the use of proportional fee policies as I also believe they are not as meaningful and also tends to encourage operators to run charge-lnd too frequently.

> If instead you plan to use it only to stop flow with 9999basefee when local is drained, then I'd have it run every 15 
Yes. But I feel it's useful to show at least one example of a static fee policy (other than the 9999basefee special case), and so we can't use such a short interval. Even 2 hours, I feel it sends the wrong message (like "it's ok to have frequent updates"), but maybe I'm needlessly over-cautious here... Would be great to use a minimum time interval that is somewhat based on gossip statistics, do you know if that exists? (assuming the worst case scenario of a large nodes with many channel state changes per hour)

asvdf:
Gossip is a neverending debate. No statistics that can be really leveraged.
I still stand by my previous opinion: 9999basefee is almost useless if applied with up to 12hrs delay, it should be applied immediately when it's needed. Fee tuning can and should be less often, but charge-lnd can't be set like it.
Or can it?
No idea, maybe two different crons with two different configs? One just for 9999basefee every 10 mins, the other for fees every N hours?
I don't use it so I wouldn't know.

VajraOfIndra:
Thanks @asvdf, your feedback is much appreciated!

> Gossip is a neverending debate. No statistics that can be really leveraged. 
Ok, well that's good to know :)

> I still stand by my previous opinion: 9999basefee is almost useless if applied with up to 12hrs delay, it should be applied  immediately when it's needed. Fee tuning can and should be less often, but charge-lnd can't be set like it. Or can it? No idea, maybe two different crons with two different configs? One just for 9999basefee every 10 mins, the other for fees every N hours? I don't use it so I wouldn't know.

I don't think charge-lnd allow for this unfortunately, although I'll have to double-check to be absolutely sure. But this would also become a bit too advanced/complex for this guide which is primarily about how to properly install the program on a node and how to setup the cron job and log.

I changed the delay from 12 hours to 2 hours for now.

asvdf:
If you use charge-lnd yourself, you should be able to confirm. It is very much possible you can launch it with `charge-lnd /path/to/config` (not at all advanced, just do the same steps twice), that way one crontab points to the 9999basefee conf, the other every 12 hours to the variable fee one (not at all advanced, just do the same steps twice)

PR #781, opened 2021-11-08
Remove "Shango", "Zap iOS" and "Electrum" bonus guides

VajraOfIndra:
**Goal:** This pull-request proposes to remove the bonus guides for the Shango Mobile wallet, Zap iOS and Electrum Personal Server.

**Rationales:** 
1. **Zap iOS**: the app is not actively maintained at the moment (see their [announcement](https://github.com/LN-Zap/zap-iOS#unmaintained) on their Github page). Therefore for security reasons, it might be better to remove the installation guide.
2. **Shango**: The Shango app does not seem to be developped/maintained anymore. The last commit in the [Github repo](https://github.com/neogeno/shango-lightning-wallet) was in June 2019, so almost 2.5 years ago... Also, the website shangoapp.com is now dead and does not redirect to anything.
3. **Electrum Personal server**:  This guide seems a bit redundant as the main guide already contains the installation of Electrs. I remember as a beginner to have been confused by this bonus guide, not understanding how it related to Electrs that I had already installed and I needed to also install this one. The last section explaining how to connect Electrum Desktop to one server only (our node's Electrs) is very useful and missing in the main guide. So this part could be saved and added at the end of the Electrs page. If the entire guide should be kept to offer an alternative to Electrs, then the introduction could be re-written to clarify how it's an alternative to the main guide Electrs.


Stadicus:
I agree on removing Shango.

With Zap iOS, I'm a bit torn, as the Android application is still under active development. I actually use it myself to manage other LND installations (not the RaspiBolt). Do you think the guide still works in principle?

For EPS, I dont' think we should just remove it. Their repo still points to this guide. Could also be removed with the version 3 of the RaspiBolt guide, while keeping the old raw version in the repo.

VajraOfIndra:
> With Zap iOS, I'm a bit torn, as the Android application is still under active development. I actually use it myself to manage other LND installations (not the RaspiBolt). Do you think the guide still works in principle?

Then we could just add a warning at the start stating it's no longer maintained and therefore potentially risky to use. Or else we could just update the guide to make it a 'Zap Android' guide instead (since this one is still actively maintained).. Or some sort of combination of both solutions?
 
> For EPS, I dont' think we should just remove it. Their repo still points to this guide. Could also be removed with the version 3 of the RaspiBolt guide, while keeping the old raw version in the repo.

Sure, it'd be great to keep it as an alternative to Electrs. We could maybe just update it and also explain in the introduction that it should be run _instead of_ Electrs and not in addition to it? (I remember I was a bit confused by this when I started).
We could even add a small comparison of Electrs and Electrum Personal Server and highlights the pros and cons of both and one might prefer to use one rather than the other..?

I'll undo the two deletions of Electrum and Zap iOS.

VajraOfIndra:
I don't know how to undo the two commits that deleted Zap iOS and EPS, so I'll close this PR and create a new one just for Shango.

PR #786, opened 2021-11-13
explorer: update to version 0.3.2

Stadicus:
Update instructions to install v0.3.2 with NodeJS 14


bmml1:
Hello,

I did not know where to put this so I will just write it down here.

With an eye towards the upcoming V3 of this guide and switch to cookie file instead of rpc user and password authentication there might be an issue here.

In the .env file there is an option to set the cookie file
`#BTCEXP_BITCOIND_COOKIE=/path/to/bitcoind/.cookie`
however user btcrpcexplorer can not access this file because it has mode 600 and belongs to the bitcoin user.

There is a workaround running the bitcoin service with umask 0027 and add user btcexplorer to the bitcoin group which I found [here](https://www.reddit.com/r/Bitcoin/comments/9y1rtn/correct_way_to_use_cookie_auth_on_server/)

> In the service unit add a UMask=0027 line under [Service]. Also add sysperms=1 to your conf file. Now when started the default system perms are 640 and the cookie is created as desired. The only thing is other files have same perms (like mempool.dat and peers.dat). I'm not too worried about that as they are read only for the bitcoin group and still blocked to others.

I do not know if the above is a good idea security wise and maybe it is better to run btcexplorer from user bitcoin direclty?

Stadicus:
> switch to cookie file instead of rpc user and password authentication there might be an issue here.

I'm aware of that. We'll probably use the "RPC auth" method that allows to use rpcuser/pw + cookie authentication in parallel. Adding `sysparms=1` to the `bitcoin.conf` gives the "bitcoin" group read-only access to the cookie, but this only works with the wallet in Bitcoin Core disabled.

Thanks for pointing this out. I think for this update here this should not yet be relevant.

Stadicus:
This pull request was resolved with the update to RaspiBolt version 3.

PR #789, opened 2021-11-15
add 404 error page

Stadicus:
With the move to the new organization and upcoming page
reorganization, a proper 404 "page not found" error page is
helpful.

The main advantage is that it includes the navigation
sidebar, so even if a specific page is not available, visitors
can quickly find what they're looking for.

### DIFF
--- a/404.md
+++ b/404.md
@@ -1,0 +1,13 @@
+---
+permalink: /404.html
+title: RaspiBolt 404
+nav_exclude: true
+---
+
+# 404: Page not found
+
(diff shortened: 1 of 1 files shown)


Stadicus:
You can see this exact same 404 page live on my fork:   
https://stadicus.github.io/raspibolt-remote/xxxxxxxx-not-found.html

VajraOfIndra (approved):
I tested it on my Github Page and could see the 404 page and the menu on the left, works great.

tACK (as part of trial run :)

PR #799, opened 2021-11-27
Electrs: add remark to wait until synced

Stadicus:
Users must wait until Electrs fully indexed the blockchain and compacted
its database before proceeding to configure wallets.

This was not explicitly mentioned, the added note should make that clear.

### DIFF
--- a/electrs.md
+++ b/electrs.md
@@ -297,6 +297,16 @@ Note that the remote device needs to have Tor installed as well.
 
 ---
 
+💡 Electrs must first fully index the blockchain and compact its database before you can connect to it with your wallets.
+This can take a few hours.
+Only proceed with the next section once Electrs is ready.
+
+* To check if Electrs is still indexing, you can follow the log output
+
+  ```sh
+  $ sudo journalctl -f -u electrs
+  ```
+
 ## Connect Electrum wallet
 
 [Electrum wallet](https://electrum.org){:target="_blank"} is a well-established, feature-rich Bitcoin wallet for power-users that supports most hardware wallets.


electrs.md
```diff
@@ -297,6 +297,12 @@ Note that the remote device needs to have Tor installed as well.
 
 ---
 
+💡 Electrs must first fully index the blockchain and compact its database before you can connect to it with your wallets.
+This can take a few hours.
+Only proceed with the next section once Electrs is ready.
+
+---
```
VajraOfIndra, line 304:
```suggestion
Only proceed with the next section once Electrs is ready.

* To check if Electrs is still indexing, open the log:

  ```sh
  $ sudo journalctl -f -u electrs
  ```

* If Electrs has completed the indexing, the log will show this kind of messages, with the height corresponding to the latest mined block:
  
  ```sh
  > Nov 28 08:54:06 indra64 electrs[1892]: [2021-11-28T08:54:06.126Z INFO  electrs::                                 index] indexing 1 blocks: [711640..711640]
  > Nov 28 08:54:06 indra64 electrs[1892]: [2021-11-28T08:54:06.247Z INFO  electrs::                                 chain] chain updated: tip=0000...xyz, height=711640
  ```

* The blockchain height can be checked using bitcoin-cli:

  ```sh
  $ bitcoin-cli getblockchaininfo | grep blocks
  >  "blocks": 711640,
  ```

---
```

VajraOfIndra (approved):
LGTM! This is a useful warning to add indeed.

I guess that users will then be wondering _how_ they can check that the indexing is completed and might be uncertain how many hours they have to wait. 

Maybe we could add a few lines to explain how to check this? There is no Electrs client that can return a true/false response when asked is the indexing is completed, correct? If not, then I suppose we can only rely on the log.. The comment below is a suggestion using the log, but there might be a cleaner way to do it!

Stadicus:
Thanks! I merged your suggestion into the same commit and force-pushed it. 

PTAL

PR #800, opened 2021-11-27
Avoid unnecessary nginx / tor restarts, reload is enough

kristapsk:
There is no need to restart neither nginx nor tor on each configuration change.


Stadicus (approved):
TIL about `systemctl reload`, thanks!

tACK

PR #802, opened 2021-11-28
Allowing external access to BTC RPC Explorer is not needed

kristapsk:
Because nginx reverse proxy already provides that.

Also, nowhere user is later directed to even try connecting to port 3002.


Stadicus (approved):
Good point, thanks! The default values are exactly what's needed and there's no need to change them:

```sh
BTCEXP_HOST=127.0.0.1
BTCEXP_PORT=3002
```

LGTM

PR #807, opened 2021-11-30
bonus: boot from microsd card

### LINKED ISSUE #806: SSD or Sdkarte
hallo people,
I'm 53 years old and tried to rebuild after your lead.
first problem what I had is that I tried it with a ssd and a memory card ds it was broken off with the time since the place was too little, clear.
the second problem is that you write you should connect to an external ssd the image of os raspberypi then to one of the usb and start, the raspberrypi can not just start from an usb without an entry in the boot.txt please help what I don't understand here thank you.

Stadicus:
#### What

This change adds a bonus guide on how to boot from a microSD card and mount an external drive to /data/ to use for data storage. This method should work with the broadest range of devices.

### Why

The new RaspiBolt version 3 assumes that all external drives can be used to boot the Raspberry Pi.

This seems to hold true for most modern drives, but if this does not work with a specific drive, the reader is stuck without any fallback instructions.

#### How

The instructions from the v2 guide are simplified, and the external drive mounted to `/data/`. Making the directory itself immutable with `chattr +i` prevents accidental data storage on the microSD card.

_Note:_ this change needs to be rebased after the new Bonus Guide structure is in place.

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

Fixes #806 

#### Test & maintenance

This change can be tested by writing a vanilla RasPiOS image to a microSD card and following the guide with a spare drive to format and mount.

#### Animated GIF (optional)

old school


bonus/raspberry-pi/boot-from-microsd-card.md
```diff
@@ -0,0 +1,174 @@
...
+
+---
+
+### Steps required
+
+To boot from a microSD card and store the data on an external drive, there are a few additional steps compared to the default RaspiBolt guide:
+
+1. [Operating system](../../operating-system.md):
+  * write the operating system to the microSD card instead of the external drive
+1. [System configuration](../../system-configuration.md):
```
VajraOfIndra, line 39:
```suggestion
2. [System configuration](../../system-configuration.md):
```

Fix typo in numbered list

bonus/raspberry-pi/boot-from-microsd-card.md
```diff
@@ -0,0 +1,174 @@
...
+{: .text-delta }
+
+1. TOC
+{:toc}
+
+---
+
+### Steps required
+
+To boot from a microSD card and store the data on an external drive, there are a few additional steps compared to the default RaspiBolt guide:
```
VajraOfIndra, line 35:
```suggestion
To boot from a microSD card and store the data on an external drive, there are a few additional steps compared to the default RaspiBolt guide. Below is a summary of the main differences. Detailed guidance is given in the following sections.
```
I added a few more words just to clarify a bit what this section is about vs the following section, as I was a bit confused at first how this section related to the next two sections. My wording is poor so could be reworked and improved for sure, but just wanted to give an example of clarification that I think could be useful.

bonus/raspberry-pi/index.md
```diff
@@ -0,0 +1,33 @@
...
+---
+
+## Security
+* **[Hardware-based (Trezor) SSH authentification](trezor-agent.md)** - a safer way to SSH to your device using your hardware wallet
+
+---
+
+## Hardware
+* **[Boot from microSD card](boot-from-microsd-card.md)** - if your Raspberry Pi is not able to boot directly from the external drive, these instructions explain how to boot from a microSD card and attach the external drive for data storage
+
```
VajraOfIndra, line 27:
```suggestion
```

operating-system.md
```diff
@@ -89,6 +89,11 @@ It should display a "Success" message after.
 * If you did not configure the Wifi settings: connect the Pi to your network with an ethernet cable
 * Start the Pi by connecting it to the power adapter using the USB-C cable
 
+## Fallback: if you can't boot from your external drive
+
+We assume that you can boot directly from the external drive.
+This method works for most modern drives, but if it doesn't work in your case, please follow the fallback instructions 'Boot from microSD card instead of external drive](bonus/raspberry-pi/boot-from-microsd-card.md).
```
VajraOfIndra, line 95:
```suggestion
This method works for most modern drives, but if it doesn't work in your case, please follow the fallback instructions [Boot from microSD card instead of external drive](bonus/raspberry-pi/boot-from-microsd-card.md).
```

system-configuration.md
```diff
@@ -80,8 +80,8 @@ Let's check if your drive works well as-is, or if additional configuration is ne
   $ sudo apt install hdparm
   ```
 
-* Your external drive should be connected as `/dev/sda1`.
-  Check if this is the case and look for the mountpoint `/`
+* Your external disk should be connected as `/dev/sda`.
```
VajraOfIndra, line 83: No change required but just curious and to learn :) Why changing from 'drive' to 'disk'? I thought disk was the outdated term between the two...

VajraOfIndra (changes requested):
I won't have time to test this guide with a SD card in the coming days/weeks so I just leave an untested review.

It's a really useful addition to the guide and a timely one, awesome to see :)

Just a few typo corrections and suggestions for very minor changes.

bonus/raspberry-pi/boot-from-microsd-card.md
Stadicus: Markdown ignores the numbers and will show them consecutively in any case. No need to do the actual numbering:
https://www.markdownguide.org/basic-syntax#ordered-lists

bonus/raspberry-pi/boot-from-microsd-card.md
Stadicus: Ack, makes sense.

bonus/raspberry-pi/index.md
Stadicus: I usually put a blank line between paragraphs, headers etc. Only simple lists are squashed together. It does not matter for web layout, but I feel it's easier to read the MD file. 

Happy to talk if we want to do this in general, or not.

operating-system.md
Stadicus: Good catch

system-configuration.md
Stadicus: Yep, I changed everything to "external drive", but the command here shows `disk` in the output, so I thought this might be more consistent.

bonus/raspberry-pi/boot-from-microsd-card.md
VajraOfIndra: It appears like this to me for now when using 'View file':
image

Is that normal and the number 2 will appear when published? Otherwise it could be an issue with the asterix items..? not sure

bonus/raspberry-pi/index.md
VajraOfIndra: I don't have strong opnions on this at all, I had just assumed it was conventional to not have blank lines between list items.
I'll do the same as you outlined above for now on (i.e. no blank line for short list items, a blank line for list with longer items).

system-configuration.md
VajraOfIndra: ah ok, makes sense, thanks!

bonus/raspberry-pi/boot-from-microsd-card.md
Stadicus: Hm, on the [branch published to github.io](https://stadicus.github.io/raspibolt-dev/bonus/raspberry-pi/boot-from-microsd-card.html#steps-required) it looks as expected. Probably an issue with the MD preview:

image

Stadicus:
Fixed the requested changes, PTAL

bonus/raspberry-pi/boot-from-microsd-card.md
VajraOfIndra: Ah great, good to know if it happens again.

VajraOfIndra:
> Fixed the requested changes, PTAL

LGTM!

VajraOfIndra (approved):

PR #810, opened 2021-12-01
Recommend optional DIY steel backup for LND 24 words

VajraOfIndra:
#### What

This PR aims at adding an optional recommendation to not only backup the LND wallet 24 mnemonic words on paper but also on steel to have a much more resilient backup. In the DIY spirit of RaspiBolt, the recommendation is to use a DIY solution with off-the-shelf items only (no bitcoin merchants used) to reduce privacy/security risks (merchant client DB leaks etc): stamped steel washers held on a bolt.

### Why

In case of node failure, the 24 words / mnemonics are indispensible to initiate a SCB recovery and recover all the user funds onchain. These words could secure a significant amount of bitcoins and therefore a resilient storage method must be recommended. The solution proposed is very aligned with the RaspiBolt ethos.

#### How

Added an extra paragraph after the paper backup paragraph that explains the solution, rationales and point towards an external guide. Slightly modified the format of the paragraph to make the steel recommendation a level 2 list item vs the compulsory paper backup which is now a level 1 list item.

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

No need for tests (this is a well-known and tested technique).
The only maintenance needed would be to occasionally check that the external link is not dead. If the link becomes dead, a bonus guide to be written instead to explain the method.

washers

### DIFF
--- a/lnd.md
+++ b/lnd.md
@@ -284,7 +284,7 @@ This piece of paper is all an attacker needs to completely empty your on-chain w
 Do not store it on a computer.
 Do not take a picture with your mobile phone.
 **This information should never be stored anywhere in digital form.**
-  * (optional, but recommended) Record these words in steel to render the backup resistant to fire, water and other hazards that could easily destroy the paper backup. In the spirit of the RaspiBolt DIY approach, a good DIY solution for this is to stamp the words into steel washers and store the washers on bolt secured by a wingnut. This solution uses only off-the-shelf components and therefore does not reveal to any merchants that you might own bitcoins. For a full guide, check out the following article: [https://www.econoalchemist.com/post/backup](https://www.econoalchemist.com/post/backup){:target="_blank"} *(note: the jig is optional, no need to buy it if you don't have a 3D printer)*.
+  * (optional) Record these words in steel to render the backup resistant to fire, water and other hazards that could easily destroy the paper backup. In the spirit of the RaspiBolt DIY approach, a good DIY solution for this is to stamp the words into steel washers and store the washers on bolt secured by a wingnut. This solution uses only off-the-shelf components and therefore does not reveal to any merchants that you might own bitcoins. For a full guide, check out the following article: [https://www.econoalchemist.com/post/backup](https://www.econoalchemist.com/post/backup){:target="_blank"} *(note: the jig is optional, no need to buy it if you don't have a 3D printer)*.
 
 * Exit the user "lnd" user session, and then exit the second SSH session altogether
 
--- a/lnd.md
+++ b/lnd.md
@@ -286,7 +286,7 @@ For this, the Static Channel Backup stored at `/data/lnd-backup/channel.backup` 
 
 🚨 This information must be kept secret at all times.
 * **Write these 24 words down manually on a piece of paper and store it in a safe place.** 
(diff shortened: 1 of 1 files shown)


Stadicus:
I really like the picture, it says more than 1000 words. But it's not really suited for the regular guide, as IMHO the metal stamping is not really required (and would detract too much from the core guide with the picture). 

How about giving the picture the place to shine in a bonus guide, linking to it from the main guide, and maybe even elaborating more on the necessary parts before linking to the external guide?

For example:

-----

**Write these 24 words down by hand**  
You can use a simple piece of paper, [print out a proper backup card](https://shiftcrypto.ch/backupcard/backupcard_print.pdf), or even [stamp the seed words into metal (bonus guide)](tbd).

⚠️ **This information should never be stored anywhere in digital form!**
  * This piece of paper is all an attacker needs to completely empty your on-chain wallet!
  * Do not store it on a computer.
  * Do not take a picture with your mobile phone.

----

WDYT?

VajraOfIndra:
> I really like the picture, it says more than 1000 words. But it's not really suited for the regular guide, as IMHO the metal stamping is not really required (and would detract too much from the core guide with the picture).
> 
> How about giving the picture the place to shine in a bonus guide, linking to it from the main guide, and maybe even elaborating more on the necessary parts before linking to the external guide?
> 
> For example:
> 
> **Write these 24 words down by hand** You can use a simple piece of paper, [print out a proper backup card](https://shiftcrypto.ch/backupcard/backupcard_print.pdf), or even [stamp the seed words into metal (bonus guide)](tbd).

Nice card btw :)

Yes simply linking to a bonus guide sounds great.

Actually rather than linking to an external guide, we could just simply explains how to do it in the bonus guide as this is now a well-known technique used by many. We could still acknlowedge the original article that first came with the idea (the one linked above I believe but could be wrong do you know who first published this idea?)

Stadicus:
> Actually rather than linking to an external guide, we could just simply explains how to do it in the bonus guide

Sounds like a good plan. Do you want to do this? In this pull request or later?

> Nice card btw :)

Thanks, my work as well... :)

VajraOfIndra:
> Sounds like a good plan. Do you want to do this? 

I wouldn't mind, should be fun, but if you'd like to write it, please do! (I'm off for 10 days now so I won't be able to work on it for a while anyway).

> In this pull request or later?

What about we link to the external guide for now and when the bonus guide is ready we update the link to it?

lnd.md
```diff
@@ -279,7 +279,8 @@ The current state of your channels, however, cannot be recreated from this seed.
 For this, the Static Channel Backup stored at `/data/lnd-backup/channel.backup` is updated continuously.
 
 🚨 This information must be kept secret at all times.
-**Write these 24 words down manually on a piece of paper and store it in a safe place.**
+* **Write these 24 words down manually on a piece of paper and store it in a safe place.** 
+You can use a simple piece of paper ([or print out a proper backup card](https://shiftcrypto.ch/backupcard/backupcard_print.pdf){:target="_blank"}), or even stamp the seed words into metal (e.g. [check this DIY guide](https://www.econoalchemist.com/post/backup){:target="_blank"}).
```
Stadicus, line 283: Nit: I find the wording with the partial sentences in the brackets a bit confusing. Maybe something like this?

> You can use a simple piece of paper, write them on a proper [backup card](xxx), or even stamp the seed words into metal (see this [DIY guide](xxx)).

Stadicus (approved):
I think the wording could be a bit improved, but otherwise LGTM.

lnd.md
VajraOfIndra: > Nit: I find the wording with the partial sentences in the brackets a bit confusing. Maybe something like this?

Sure, I updated accordingly.

PR #811, opened 2021-12-02
Update lntop bonus guide for RaspiBolt v3

kristapsk:
#### What

Fixes guide for RaspiBolt v3. Changes download from armv6 to arm64<del> and adds instructions about copying necessary LND credentials</del>.

#### Scope

- [X] independent bonus guide
- [X] simple bug fix

Note - there is also recent v0.2.0 release of lntop, but it does not provide precompiled binaries.

### DIFF
--- a/bonus/lightning/lntop.md
+++ b/bonus/lightning/lntop.md
@@ -28,12 +28,16 @@ Difficulty: Easy
 
 ### Install lntop
 
+* As user "admin", download and install application
+
 ```bash
-$ wget https://github.com/edouardparis/lntop/releases/download/v0.1.0/lntop_Linux_armv6.tar.gz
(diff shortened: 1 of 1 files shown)


Stadicus:
Thanks! I also just added https://github.com/raspibolt/raspibolt/pull/812, which gives the user "admin" access to lnd by default. That should make most additional authentication steps obsolete.

kristapsk:
> Maybe we can add #812 first into the main guide, and then this one without the additional copy steps after?

I think that's a good idea.

kristapsk:
Will modify this PR afer #812 is merged.

Stadicus (approved):
LGTM! 

I noticed that v0.2.0 is available, but only the source code. I asked if binary releases are planned. Not a blocker for this pull request, however, can be updated later.

PR #814, opened 2021-12-02
Fixed NodeJS version typo

VajraOfIndra:
#### What

Fixes a small typo for the NodeJS version used for RTL.

### Why

To avoid confusion for new users.

#### How

Fixed typo "12" to "14"

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

N/A

#### Additional remarks

NodeJS version 14 is describd as "the most recent stable one". However when checking [https://nodejs.org/en/about/releases/](https://nodejs.org/en/about/releases/) it shows v18 as "Current", v16 as the "Active LTS" and v14 and v12 as "Maintenance LTS". Wouldn't v16 be the most recent stable one? not sure how 'stable' matches the NodeJS terminology there! :)

Shoudl we reword "the most recent stable one"..?

NodeJS

### DIFF
--- a/rtl.md
+++ b/rtl.md
@@ -27,7 +27,7 @@ We install [Ride The Lightning](https://github.com/Ride-The-Lightning/RTL/blob/m
 ### Install NodeJS
 
 Starting with user “admin”, we switch to user “root” and add the Node JS package repository.
-We’ll use version 14 which is the most recent stable one.
+We’ll use version 14 which is well-tested with all web applications used in this guide.
 If you installed BTC RPC Explorer, then you've already accomplisehd this step.
 When finished, exit the “root” user session.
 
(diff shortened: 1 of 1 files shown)


Stadicus (changes requested):
> the most recent stable one

You're right, that's not correct. And would be quickly outdated even if it was.

Maybe we could rephrase to "which is well-tested with all web applications used in this guide."?

Stadicus:
BTW: I tested v16 and ran into some issues, but now I'm pretty sure that was just me. If we can set up and test the RaspiBolt with v16, that could be a good thing to do in the near future as well.

Stadicus (approved):
LGTM!

VajraOfIndra:
> > the most recent stable one
> 
> You're right, that's not correct. And would be quickly outdated even if it was.
> 
> Maybe we could rephrase to "which is well-tested with all web applications used in this guide."?

Perfect, I updated the sentence accordingly.

VajraOfIndra:
> BTW: I tested v16 and ran into some issues, but now I'm pretty sure that was just me. If we can set up and test the RaspiBolt with v16, that could be a good thing to do in the near future as well.

That'd be a nice thing to do. Is there a way to keep track of things to work on later on?

PR #816, opened 2021-12-03
Add passwords [D] & [E] for Explorer and RTL

VajraOfIndra:
#### What

Passwords used at various steps of the guide are setup at the start in the 'Preparation' section and given a letter name (A to C for now). 

An optional password is discussed for the Explorer and a compulsory password is required for RTL. These two passwords are not mentionned yet in the preparation section.

For consistency, this PR proposes to add password [D] and [E] in the preparation section for respectively the Explorer and RTL. The PR also refers to passwords [D] and [E] in the Explorer and RTL guides (as is done for passwords A to C in the other guides that use them).

### Why

Improves guide consistency. Reduce the risk of users not recording the passwords for Epxlorer and RTL in a proper way.

#### How

Added references to passwords [D] and [E]

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

Test: just checking that the right password letters are used in the right section of the guides
Maintenance: None

password

### DIFF
--- a/preparations.md
+++ b/preparations.md
@@ -56,6 +56,8 @@ They should be unique and very secure, at least 12 characters in length. Do **no
 [ A ] Master user password
 [ B ] Bitcoin RPC password
 [ C ] LND wallet password
+[ D ] BTC-RPC-Explorer (optional)
+[ E ] Ride The Lightning
 ```
 
 ![xkcd: Password Strength](images/preparations_xkcd.png)
--- a/preparations.md
+++ b/preparations.md
@@ -56,8 +56,8 @@ They should be unique and very secure, at least 12 characters in length. Do **no
 [ A ] Master user password
 [ B ] Bitcoin RPC password
 [ C ] LND wallet password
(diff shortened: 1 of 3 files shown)


preparations.md
```diff
@@ -56,6 +56,8 @@ They should be unique and very secure, at least 12 characters in length. Do **no
 [ A ] Master user password
 [ B ] Bitcoin RPC password
 [ C ] LND wallet password
+[ D ] BTC-RPC-Explorer (optional)
+[ E ] Ride The Lightning
```
Stadicus, line 60: For consistency, use the following?

```
[ D ] BTC-RPC-Explorer password (optional)
[ E ] Ride-The-Lightning password
```

Stadicus (approved):
One little nit, otherwise LGTM :+1:

preparations.md
VajraOfIndra: Good catch! I changed accordingly, PTAL

PR #820, opened 2021-12-04
Tor: move web endpoints to TLS/SSL

Stadicus:
#### What

Login passwords transmitted without TLS/SSL over the Tor network 
are a security risk if some of the nodes routing the traffic are
malicious.

### Why

While the Tor network is encrypted in itself, malicious actors can run
nodes that can snoop on traffic. A password transmitted in cleartext
could allow to log in to RTL and steal your Bitcoin (if no 2FA enabled).

### How

This change moves all web endpoint exposed as Tor .onion addresses
to the NGINX ports, adding TLS/SSL for additional encryption on top of
the Tor network.

The downside is that this will show a browser warning due to the
self-signed certificate.

### Scope

- [ ] independent bonus guide
- [x] simple bug fix

### Test & maintenance

Test this pull request by adding the new services and connecting with the Tor browser. No ongoing maintenance needed.

### DIFF
--- a/btcrpcexplorer.md
+++ b/btcrpcexplorer.md
@@ -291,20 +291,22 @@ You can easily do so by adding a Tor hidden service on the RaspiBolt and accessi
 
   ```sh
   ############### This section is just for location-hidden services ###
-  HiddenServiceDir /var/lib/tor/hidden_service_btcrpcexplorer/
+  HiddenServiceDir /var/lib/tor/hidden_service_btcrpcexplorer_ssl/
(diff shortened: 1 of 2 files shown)


Stadicus:
Converted to DRAFT due to some online discussions. I need to research this topic a bit more first.

Stadicus:
Closing this PR, as it seems Tor v3 hidden services are just fine over HTTP, Potential issues arise when leaving the Tor network through exit nodes.

PR #821, opened 2021-12-04
lntop: update to v0.2.0

Stadicus:
#### What

Update to latest release now that binaries are available.

### Why

Version 0.2.0 of lntop adds some new features and bugfixes:  
https://github.com/edouardparis/lntop/releases/tag/v0.2.0

This change also adds binary signature verification as best practice.

### How

Simple update of download links, plus added instructions for checksum signature validation. See https://github.com/edouardparis/lntop/issues/49 for some context.

Deleting downloaded files is removed as these are small and the /tmp folder is purged on every reboot anyway.

### Scope

- [ ] independent bonus guide
- [x] simple application update

### Test & maintenance

Follow instructions and run `lntop`

### DIFF
--- a/bonus/lightning/lntop.md
+++ b/bonus/lightning/lntop.md
@@ -68,7 +68,7 @@ Table of contents
 * Verify the signed checksum against the actual checksum of your download
 
   ```sh
-  $ $ sha256sum --check checksums-lntop-v0.2.0.txt --ignore-missing
+  $ sha256sum --check checksums-lntop-v0.2.0.txt --ignore-missing
   > lntop-v0.2.0-Linux-arm64.tar.gz: OK
   ```
 
@@ -83,10 +83,12 @@ Table of contents
 
 ### Run lntop
 
-```bash
+Depending on the size of your LND channel database, lntop can take quite a while to start.
+
+```sh
 $ lntop
 ```
 
-------
+---
 
 << Back: [+ Lightning](index.md)
--- a/bonus/lightning/lntop.md
+++ b/bonus/lightning/lntop.md
@@ -7,42 +7,86 @@ nav_exclude: true
 has_toc: false
 ---
 
-## Bonus guide: lntop: LNTOP terminal dashboard
+## Bonus guide: LNTOP terminal dashboard
 {: .no_toc }
+
+---
+
+[lntop](https://github.com/edouardparis/lntop){:target="_blank"} is an interactive text-mode channels viewer for Unix systems.
 
 Difficulty: Easy
 {: .label .label-green }
 
-<details open markdown="block">
-  <summary>
-    Table of contents
-  </summary>
-  {: .text-delta }
-1. TOC
-{:toc}
-</details>
-
-[lntop](https://github.com/edouardparis/lntop) is an interactive text-mode channels viewer for Unix systems.
+Status: Tested v3
+{: .label .label-green }
 
 ![lntop](../../images/74_lntop.png)
 
+---
+
+Table of contents
+{: .text-delta }
+
+1. TOC
+{:toc}
+
+---
+
 ### Install lntop
 
-* As user "admin", download and install application
+* As user “admin”, download the application, checksums, and the corresponding signature file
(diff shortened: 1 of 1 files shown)


bonus/lightning/lntop.md
```diff
@@ -7,42 +7,86 @@ nav_exclude: true
...
+  > gpg: Good signature from "Edouard (Personal) <m@edouard.paris>" [unknown]
+  > gpg: WARNING: This key is not certified with a trusted signature!
+  > gpg:          There is no indication that the signature belongs to the owner.
+  > Primary key fingerprint: A8BA 5205 BFCB C668 853D  5602 47EE BB01 4DD8 0918
+  ```
+
+* Verify the signed checksum against the actual checksum of your download
+
+  ```sh
+  $ $ sha256sum --check checksums-lntop-v0.2.0.txt --ignore-missing
```
VajraOfIndra, line 71:
```suggestion
  $ sha256sum --check checksums-lntop-v0.2.0.txt --ignore-missing
```
Fixed typo

bonus/lightning/lntop.md
```diff
@@ -7,42 +7,86 @@ nav_exclude: true
...
+
+---
 
 ### Run lntop
 
 ```bash
 $ lntop
 ```
+
 ------
```
VajraOfIndra, line 90: Nitpicking here, but is there any reason why there is 6 - there and not 3 as usual?

bonus/lightning/lntop.md
```diff
@@ -7,42 +7,86 @@ nav_exclude: true
...
+  $ sudo install -m 0755 -o root -g root -t /usr/local/bin release/lntop
+  ```
+
+---
 
 ### Run lntop
 
 ```bash
 $ lntop
 ```
```
VajraOfIndra, line 88:
```suggestion
### Run lntop

```bash
$ lntop
```

* Note: If you have a large LND database, it might take a while (up to 10-20 seconds) for the dashboard to appear. You'll see a blank screen while it's loading, do not worry!
```

I thought I had an issue with the install because I got a black screen when I ran lntop. The first couple of times I Ctrl+C before it had time to load! I assume it's dependent on` channel.db` size... So I suggest we add a little note to warn users it might  take a while for the dashboard to load. It takes about 10 seconds for me right now.

But if you think that's overkill, no problem, we can just let the user figure it out.

VajraOfIndra (changes requested):
Nice update! one more bonus guide which is up-to-date for v3 :1st_place_medal: 

A couple of typo fixes and one suggestion to reassure user that the dashbaord can take a while to load.

Other than that, I think it could be beneficial to add a section that explains how to navigate the dashboard with the keyboard, using F1 (help), F2 (left menu), Enter, q (quit), a (sort column, ascending), d (sort column, descending). I think that'd be super helpful, particulalry because it's not easy to find the explanations. But we could do this with another PR later.

Other than that, I'm puzzled regarding the Go requirement. I thought Go was _not_installed by default on RasPiOS Lite.. so how come we can use lntop without installing Go first?

bonus/lightning/lntop.md
Stadicus: No difference, 3x `-` have the same effect.

bonus/lightning/lntop.md
Stadicus: Ack, thanks!

bonus/lightning/lntop.md
Stadicus: Good point

Stadicus:
Addressed all your feedback, thanks!

> Other than that, I think it could be beneficial to add a section that explains how to navigate the dashboard with the keyboard, using F1 (help), F2 (left menu), Enter, q (quit), a (sort column, ascending), d (sort column, descending).

Good idea! I think that could be another pull-request, though. We could open an issue for this as a reminder, with the "help wanted" tag.

PTAL

VajraOfIndra (approved):
LGTM!

Issue #826, opened 2021-12-08
Blockchain Explorer cannot be updated v.3.3.0


Ubuntu51:
### Describe the issue
A clear and concise description of where you're stuck in the guide. What are you trying to do, and how is it not working?

### Location in guide
Link to the corresponding page and section on <https://raspibolt.org>

### Environment
 - Hardware platform: [e.g. Raspberry Pi 4, Odroid, ... ]
 - Operating system: [e.g, Raspberry Pi OS 64bit, Debian 11, ... ]
 - Version: [ output of `lsb_release -d` ]

Ubuntu51:
I try to make an update as described in the instructions 
https://raspibolt.org/btcrpcexplorer.html

git checkout v3.2.0 (old)
when I enter git checkout v3.3.0( New Version) comes the message 

**`error: Your local changes to the following files would be overwritten by checkout:
	npm-shrinkwrap.json
Please commit your changes or stash them before you switch branches.
Aborting`**

Do I enter npm install 

**npm WARN bootstrap@5.0.2 requires a peer of @popperjs/core@^2.9.2 but none is installed. You must install peer dependencies yourself.

audited 482 packages in 8.297s

43 packages are looking for funding
  run `npm fund` for details

found 15 vulnerabilities (11 moderate, 4 high)
  run `npm audit fix` to fix them, or `npm audit` for details**

Work on a RPI4

Thanks

Stadicus:
Try resetting the local repository before checking out the new version:

```sh
$ git reset --hard HEAD
$ git checkout v3.3.0
```

Stadicus:
I found another issue with the upgrade process: the command `git describe --tags --abbrev=0` does not return the latest tag because tag `v3.3.0` has no tag date. So if sorted, `v3.2.0` is still the newest tag... Never seen this before.

```sh
$ git for-each-ref --sort=-taggerdate --format '%(refname) %(taggerdate)' refs/tags

refs/tags/v3.2.0 Tue Aug 10 13:23:04 2021 -0400
refs/tags/v3.1.1 Tue Apr 20 19:00:30 2021 -0400
refs/tags/v3.1.0 Wed Apr 14 13:23:59 2021 -0400
refs/tags/v3.0.0 Thu Apr 8 11:47:27 2021 -0400
refs/tags/v2.2.0 Fri Jan 22 14:15:57 2021 -0500
refs/tags/v2.1.0 Tue Dec 15 14:32:46 2020 -0500
[...]

PR #829, opened 2021-12-08
BTC RPC Explorer: update to v3.3.0

Stadicus:
### What

This change updates BTC RPC Explorer to the lastest version and fixes the update instructions.

### Why

The latest release brings a ton of new features and is well worth the update.

Also, the current update guide has some issues that need to be addressed. The NPM install process changes files in the repository that block the checkout of a newer branch later on. The repo needs to be reset first, and local changes discarded.

Additionally, v3.3.0 was tagged without a date, not sure how that's even possible. But this breaks the 'git describe' command,
so at least temporarily we fall back to 'git tag'.

#### How

see above

fixes https://github.com/raspibolt/raspibolt/issues/826

### DIFF
--- a/btcrpcexplorer.md
+++ b/btcrpcexplorer.md
@@ -124,7 +124,7 @@ An attacker would not be able to do much within this user's permission settings.
 * Download the source code directly from GitHub and install all dependencies using the Node Package Manager (NPM).
 
   ```sh
-  $ git clone --branch v3.2.0 https://github.com/janoside/btc-rpc-explorer.git
+  $ git clone --branch v3.3.0 https://github.com/janoside/btc-rpc-explorer.git
   $ cd btc-rpc-explorer
   $ npm install
   ```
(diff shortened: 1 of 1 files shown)


btcrpcexplorer.md
```diff
@@ -322,13 +322,14 @@ Updating to a [new release](https://github.com/janoside/btc-rpc-explorer/release
...
-* Fetch the latest GitHub repository information, display the latest release tag (`v3.2.0` in this example), and update:
+* Fetch the latest GitHub repository information, display the release tags (use the latest `v3.3.0` in this example), and update:
 
   ```sh
   $ cd /home/btcrpcexplorer/btc-rpc-explorer
   $ git fetch
-  $ git describe --tags
-  $ git checkout v3.2.0
+  $ git reset --hard HEAD
+  $ git tag
```
VajraOfIndra, line 331: What does this command do? It seems to just list all the available tags including the latest fetched one. Is it to cdouble-check that the desired tag is there and know what version to use in the next command?

VajraOfIndra (approved):
LGTM!
tACK (updated my Explorer following this PR)

Just a question regarding the goal of git tag (see comment below).

PS: Would it be useful to add a line at the end, recommending the user to open their Explorer and scroll to the bottom of the home page to double-check that they are running the latst version? just an idea

btcrpcexplorer.md
Stadicus: `git describe` is much cleaner, as you can output just the latest tag. Unfortunately, that does not work with the current version:

> Additionally, v3.3.0 was tagged without a date, not sure how that's even possible. But this breaks the 'git describe' command, so at least temporarily we fall back to 'git tag'.

Here, the goal would just be to manually spot the biggest version and use it.

PR #831, opened 2021-12-09
Fixed typo in /etc/fstab

bavarianledger:
replaced /mnt/extnew with /data

just a typo...


Stadicus (approved):
Thanks for spotting this copy/pasta mistake. LGTM!

PR #838, opened 2021-12-12
Fix: lnd arm version from arm64 to armv7 for RPi

vindard:
#### What

`lnd` compatibility with Raspberry Pi devices. The correct version for RPi devices is `armv7` build.

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

### DIFF
--- a/lnd.md
+++ b/lnd.md
@@ -32,7 +32,7 @@ We'll download, verify and install LND.
 
   ```sh
   $ cd /tmp
-  $ wget https://github.com/lightningnetwork/lnd/releases/download/v0.14.1-beta/lnd-linux-arm64-v0.14.1-beta.tar.gz
+  $ wget https://github.com/lightningnetwork/lnd/releases/download/v0.14.1-beta/lnd-linux-armv7-v0.14.1-beta.tar.gz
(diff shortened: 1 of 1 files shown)


Stadicus:
The RaspiBolt v3 guide assumes you run a Raspberry Pi 4 with the latest Raspberry Pi OS 64-bit:   
https://raspibolt.org/operating-system.html

The `armv7` binaries are compiled for 32-bit operating systems. The `arm64` binaries (also called `armv8`) are the correct ones when used on a 64-bit system.

vindard:
Ah ok gotcha, I need to double-check my setup then because I thought had Raspberry Pi OS on my Pi 4 but still ran into this issue 🙈

vindard:
Raspbian Buster 10, looks like I set my Pi up quite a while back, hadn't realized 64-bit was now the standard 🙈

```
PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
NAME="Raspbian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
```

PR #839, opened 2021-12-12
Add Zeus as part of the main guide

VajraOfIndra:
#### What

This PR moves the Zeus bonus guide to the main part of the guide under the 'Lighnting' section.
It also update the guide to the latest versions of lndconnect and Zeus and to RaspiBolt v3 standards.

### Why

The Zeus mobile app connects directly to the node and allows to make on-chain and LN payments and operate the LN node while on-the-move. This use case is important enough to be part of the main guide rather than the bonus guide. 

Zeus is actively maintained, constantly adds new features as the LN space develops. It is also cross-platform and can be used by both Android and iOS users.

#### How

- Moved bonus guide to 'Lightning' section of the main guide
- Updated guide: lndconnect v0.2, illustrations etc
- Updated the 'Index' and 'Lightning' pages to make reference to Zeus
- Updated the 'RTL' page to points to 'Zeus' as the next page of the guide
- Updated the '+ Lightning' page to remove the refreence to the Zeus bonus guide

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

Test: Follow the guide and test that Zeus is properly connecting to the node.

Maintenance: Low. Keep an eye on updates for lndconnect (although not maintained anymore)

#### Animated GIF (optional)

zeus


VajraOfIndra:
> According to Zeus, they'll have the v0.6.0 in the iOS and Google stores early January

Do you want to wait until January before merging?

Stadicus:
> Do you want to wait until January before merging?

No, no. Just a general FYI because I mentioned I ask them. But it makes it also more clear that we don't need to explain APK and Testflight install. We can just start with the current version and update then.

VajraOfIndra:
> > Do you want to wait until January before merging?
> 
> No, no. Just a general FYI because I mentioned I ask them. But it makes it also more clear that we don't need to explain APK and Testflight install. We can just start with the current version and update then.

:+1:

VajraOfIndra:
I close this PR as it has conflicts. This PR is superceded by #921

PR #842, opened 2021-12-14
Fix #793: Add screenshot to show symlinks colour in LND guide?

### LINKED ISSUE #793: Expected output from ls -ln from lnd directory
May you here probably put in a screenshot with the expected output? As a reference how it should look like if it’s all well configured.
Screenshot 2021-11-24 at 12 02 04

VajraOfIndra:
#### What

Fix #793 : a PR to add a screenshot that shows the non-red colour of the symlinks in the lnd user home directory.

This is just a proposal to discuss how it looks and if this is something which should be in the guide or not...

PROs: 

* shows the expected ouput with colours

CONs: 

* Not all terminals will have the same colours so it might be confusing
* The guide does not usually use screenshots to show expected command results...

### Why

See above, to be discussed

#### How

Added a screenshot in /images and linked to it in the LND guide

#### Scope

- [x] small change to core configuration
- [ ] independent bonus guide
- [ ] simple bug fix

Fixes #793 

#### Test & maintenance

Test: Check that content of LND home directory is as expected
Maintenance: None


Stadicus:
Nice, I like it. Standard Ubuntu terminal looking good! :)

Stadicus (approved):

PR #846, opened 2021-12-17
bitcoind: use 'startupnotify' for group perms

### LINKED ISSUE #801: JoinMarket is currently incompatible with RaspiBolt v3
b845122d6e98b958d5aa6893fcef67599cdd0d57 disables Bitcoin Core wallet functionality, which is required for JoinMarket (watchonly, for address / transaction monitoring). So, if this isn't changed, link to JoinMarket guide in bonus section should be removed.

Stadicus:
### What

This pull request changes the method to ensure that the .cookie file is group accessible.

### Why

Bitcoin Core is accessible to applications run by different users through the authentication cookie. Users that are a member of the group "bitcoin" need read-only access to /data/bitcoin/.cookie and can then authenticate themselves without username and password.

By default, Bitcoin Core does not create the .cookie file with group permissions. As of now, the bitcoind argument `sysperms=1`
is used with UMask=0027 to override this. Bitcoind only allows this if `disablewallet=1` is set. This breaks compatibility with
applications that rely on the Bitcoin Core wallet like JoinMarket or Electrum Personal Server.

### How

This change uses a different method that allows to keep the wallet enabled: executing a shell command after startup of bitcoind, configured with the argument `startupnotify` and triggered by bitcoind itself.

### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

Fixes #801

### Test & maintenance

* Make sure Bitcoin Core is configured to use the `rpcauth` method [used in the RaspBolt v3](https://raspibolt.org/bitcoin-core.html#generate-access-credentials), and not the `rpcuser` / `rpcpassword` of older guides
* Disable the following options in the `bitcoin.conf` file:
  ```
  #sysperms=1
  #disablewallet=1
  ```
* Restart Bitcoin Core and make sure the `.cookie` file is generated. The `-rw-------` part means that only the user "bitcoin", but not the group "bitcoin" has (r)ead and (w)rite access to the file. For more context, see the comments above.
  ```
  $ sudo systemctl restart bitcoind
  $ $ ls -la /data/bitcoin | grep .cookie
  > -rw------- 1 bitcoin bitcoin      75 Dec 12 12:25 .cookie
  ```
* Update the file `/etc/systemd/system/bitcoind.service` according to this pull request.

* Reload the system unit and restart Bitcoin Core again. Then check if the group "bitcoin" now has (r)ead access to the file, as the additional `r` indicates in the output below (this can take several seconds):
  ```
  $ sudo systemctl daemon-reload
  $ sudo systemctl restart bitcoind
  $ watch ls -la /data/bitcoin/.cookie
  > -rw-r----- 1 bitcoin bitcoin      75 Dec 12 12:25 .cookie

  ```

### DIFF
--- a/bitcoin-core.md
+++ b/bitcoin-core.md
@@ -173,10 +173,6 @@ We'll also set the proper access permissions.
   server=1
   txindex=1
 
-  # File permissions, necessary for cookie authentication
-  sysperms=1
-  disablewallet=1
-
   # Network
   listen=1
   listenonion=1
@@ -265,7 +261,8 @@ We use “systemd“, a daemon that controls the startup process using configura
   ExecStart=/usr/local/bin/bitcoind -daemon \
                                     -pid=/run/bitcoind/bitcoind.pid \
                                     -conf=/home/bitcoin/.bitcoin/bitcoin.conf \
-                                    -datadir=/home/bitcoin/.bitcoin
+                                    -datadir=/home/bitcoin/.bitcoin \
+                                    -startupnotify="chmod g+r /home/bitcoin/.bitcoin/.cookie"
 
   # Process management
   ####################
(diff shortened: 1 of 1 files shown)


kristapsk (approved):
ACK 2aa4b7d24e64b48a61255bec62d406562b6f3afe

kristapsk:
Note to myself - probably need to open an issue in Bitcoin Core repo about this, to make it possible to do this in a less hackish way.

VajraOfIndra (approved):
LGTM :+1: 
tACK

Very nice that you solved this issue, thanks!

Stadicus:
Thanks @kristapsk, let's merge! :)

PR #862, opened 2021-12-29
Fix cmd in doc

shafan:
For make the directory immutable, the command parameter is : +i and not +1.

#### What

### Why

Why is this change important?

#### How

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

#### Animated GIF (optional)


Stadicus (approved):
Thanks!

PR #865, opened 2021-12-31
Update bitcoin-core.md

orenfromberg:
#### What

What is the reason of this change?

Following the guide here [at this point](https://raspibolt.org/bitcoin-core.html#verification-of-bitcoind-operations), I ran the command `systemctl status bitcoind.service` and saw the following results and warning:

```
admin@raspibolt3:~/.bitcoin $ systemctl status bitcoind.service
* bitcoind.service - Bitcoin daemon
     Loaded: loaded (/etc/systemd/system/bitcoind.service; enabled; vendor preset: enabled)
     Active: active (running) since Thu 2021-12-30 23:15:37 EST; 5min ago
    Process: 786 ExecStart=/usr/local/bin/bitcoind -daemon -pid=/run/bitcoind/bitcoind.pid -conf=/home/bitcoin/.bitcoin/bitcoin.conf -datadir=/home/bitcoin/.bitcoin -startupnotify=chmod g>
   Main PID: 787 (bitcoind)
      Tasks: 17 (limit: 4165)
        CPU: 6.324s
     CGroup: /system.slice/bitcoind.service
             `-787 /usr/local/bin/bitcoind -daemon -pid=/run/bitcoind/bitcoind.pid -conf=/home/bitcoin/.bitcoin/bitcoin.conf -datadir=/home/bitcoin/.bitcoin -startupnotify=chmod g+r /home>

Warning: some journal files were not opened due to insufficient permissions.
```

Next I noted that the cookie had the correct read permissions on the group.

Next I failed to view the debug log:
```
admin@raspibolt3:~/.bitcoin $ tail -f /home/bitcoin/.bitcoin/debug.log
tail: cannot open '/home/bitcoin/.bitcoin/debug.log' for reading: Permission denied
tail: no files remaining
```

Next I noticed that not all the files in the `~/.bitcoin` directory had the read permission for the bitcoin group.

I modified the group read permissions recursively on the directory with the command:
```
$ sudo chmod -R g+r /data/bitcoin
```

Then I was able to tail the log and the warning no longer appeared after restarting the bitcoind service.

### Why

Why is this change important?

All users in the bitcoin group should be able to access all the files in the bitcoin data directory.

#### How

Recursively modifying the group read permissions on all files in the bitcoin data directory.

#### Scope

- [ ] significant change to core configuration
- [ ] independent bonus guide
- [x] simple bug fix

Fixes #864 

#### Test & maintenance
[...]

PR #866, opened 2021-12-31
Update bitcoin-core.md: include warning and sudo when tailing debug.log

### LINKED ISSUE #864: [BUG] "Warning: some journal files were not opened due to insufficient permissions." and "cannot open debug.log for reading"
### Describe the bug
Following the guide here [at this point](https://raspibolt.org/bitcoin-core.html#verification-of-bitcoind-operations), I ran the command `systemctl status bitcoind.service` and saw the following results and warning:

```
admin@raspibolt3:~/.bitcoin $ systemctl status bitcoind.service
* bitcoind.service - Bitcoin daemon
     Loaded: loaded (/etc/systemd/system/bitcoind.service; enabled; vendor preset: enabled)
     Active: active (running) since Thu 2021-12-30 23:15:37 EST; 5min ago
    Process: 786 ExecStart=/usr/local/bin/bitcoind -daemon -pid=/run/bitcoind/bitcoind.pid -conf=/home/bitcoin/.bitcoin/bitcoin.conf -datadir=/home/bitcoin/.bitcoin -startupnotify=chmod g>
   Main PID: 787 (bitcoind)
      Tasks: 17 (limit: 4165)
        CPU: 6.324s
     CGroup: /system.slice/bitcoind.service
             `-787 /usr/local/bin/bitcoind -daemon -pid=/run/bitcoind/bitcoind.pid -conf=/home/bitcoin/.bitcoin/bitcoin.conf -datadir=/home/bitcoin/.bitcoin -startupnotify=chmod g+r /home>

Warning: some journal files were not opened due to insufficient permissions.
```

Additionally, I was unable to tail the log file:
```
admin@raspibolt3:~/.bitcoin $ tail -f /home/bitcoin/.bitcoin/debug.log
tail: cannot open '/home/bitcoin/.bitcoin/debug.log' for reading: Permission denied
tail: no files remaining
```

### Location in guide
https://raspibolt.org/bitcoin-core.html#verification-of-bitcoind-operations

### Expected behavior
I expect to not see the warning or the error when tailing debug.log

### Environment
 - Hardware platform: Raspberry Pi 4
 - Operating system: Raspberry Pi OS 64bit
 - Version: Debian GNU/Linux 11 (bullseye)

orenfromberg:
#### What

Adding an expected warning as well as prepending a command to use sudo.

This PR fixes #864 and succeeds #865 based on [this comment](https://github.com/raspibolt/raspibolt/issues/864#issuecomment-1003272244) by @c-mannix 

### Why

This change will prevent future confusion by users inspecting the bitcoind service and tailing the debug.log file.

#### How

* Now including the warning in the output
* Explicitly directing use of `sudo` with tail command
* 
#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

Fixes #864 

#### Test & maintenance

Test the command by running `tail -f /home/bitcoin/.bitcoin/debug.log` successfully.

### DIFF
--- a/bitcoin-core.md
+++ b/bitcoin-core.md
@@ -337,6 +337,7 @@ After rebooting, "bitcoind" should start and begin to sync and validate the Bitc
   >      CGroup: /system.slice/bitcoind.service
   >              `-2317 /usr/local/bin/bitcoind -daemon -pid=/run/bitcoind/bitcoind.pid -conf=/home/bitcoin/.bitcoin/bitcoin.conf > -datadir=/home/bitcoin/.bitcoin
   >
+  > Warning: some journal files were not opened due to insufficient permissions.
   ```
 
 * Check if the permission cookie can be accessed by the group "bitcoin".
@@ -351,7 +352,7 @@ After rebooting, "bitcoind" should start and begin to sync and validate the Bitc
   Exit with `Ctrl-C`
 
   ```sh
-  $ tail -f /home/bitcoin/.bitcoin/debug.log
+  $ sudo tail -f /home/bitcoin/.bitcoin/debug.log
   ```
 
 * Use the Bitcoin Core client `bitcoin-cli` to get information about the current blockchain
--- a/bitcoin-core.md
+++ b/bitcoin-core.md
@@ -220,6 +220,12 @@ Still logged in as user "bitcoin", let's start "bitcoind" manually.
 
 * Once everything looks ok, stop "bitcoind" with `Ctrl-C`
 
+* Grant the "bitcoin" group read-permission for the debug log file:
+  
+  ```sh
+  $ sudo chmod g+r /data/bitcoin/debug.log
+  ```
+
 * Exit the “bitcoin” user session back to user “admin”
 
   ```sh
@@ -337,7 +343,6 @@ After rebooting, "bitcoind" should start and begin to sync and validate the Bitc
   >      CGroup: /system.slice/bitcoind.service
   >              `-2317 /usr/local/bin/bitcoind -daemon -pid=/run/bitcoind/bitcoind.pid -conf=/home/bitcoin/.bitcoin/bitcoin.conf > -datadir=/home/bitcoin/.bitcoin
   >
-  > Warning: some journal files were not opened due to insufficient permissions.
   ```
 
 * Check if the permission cookie can be accessed by the group "bitcoin".
--- a/bitcoin-core.md
+++ b/bitcoin-core.md
@@ -332,7 +332,7 @@ After rebooting, "bitcoind" should start and begin to sync and validate the Bitc
   Exit with `Ctrl-C`
 
   ```sh
-  $ systemctl status bitcoind.service
+  $ sudo systemctl status bitcoind.service
   > * bitcoind.service - Bitcoin daemon
   >      Loaded: loaded (/etc/systemd/system/bitcoind.service; enabled; vendor preset: enabled)
   >      Active: active (running) since Thu 2021-11-25 22:50:59 GMT; 7s ago
--- a/bitcoin-core.md
+++ b/bitcoin-core.md
@@ -223,7 +223,7 @@ Still logged in as user "bitcoin", let's start "bitcoind" manually.
 * Grant the "bitcoin" group read-permission for the debug log file:
   
   ```sh
-  $ sudo chmod g+r /data/bitcoin/debug.log
+  $ chmod g+r /data/bitcoin/debug.log
   ```
 
 * Exit the “bitcoin” user session back to user “admin”
@@ -357,7 +357,7 @@ After rebooting, "bitcoind" should start and begin to sync and validate the Bitc
   Exit with `Ctrl-C`
 
   ```sh
-  $ sudo tail -f /home/bitcoin/.bitcoin/debug.log
+  $ tail -f /home/bitcoin/.bitcoin/debug.log
   ```
 
 * Use the Bitcoin Core client `bitcoin-cli` to get information about the current blockchain


VajraOfIndra (changes requested):
Thanks for following-up on this, it'd be useful to sort it out.

For the first change: LGTM. We could either make this change (adding the warning output) or else we could run the command with `sudo` and add the log outputs, WDYT?

The second change is not necessary I believe. The Bitcoin `debug.log` is owned by the "bitcoin" user and the user invoking the `tail` command is "admin" which is a member of the "bitcoin" group and therefore has read privilege. On my node, the `tail` command (without `sudo`) works fine.

Nice gif! :)

c-mannix:
> For the first change: LGTM. We could either make this change (adding the warning output) or else we could run the command with `sudo` and add the log outputs, WDYT?
> 

 Personally I don't like see warnings at all so perhaps we could show the command `systemctl status bitcoind.service` with `sudo`, if we leave the warning it might be good to also explain that  using `sudo` will remove the warning.

> The second change is not necessary I believe. The Bitcoin `debug.log` is owned by the "bitcoin" user and the user invoking the `tail` command is "admin" which is a member of the "bitcoin" group and therefore has read privilege. On my node, the `tail` command (without `sudo`) works fine.

 Having recently gone through the guide myself I ran into the same permission issue. Even though `admin` is part of the `bitcoin` group, invoking `tail -f /home/bitcoin/.bitcoin/debug.log` on my node without `sudo`  results in:
 
```
admin@raspibolt:~ $ tail -f /home/bitcoin/.bitcoin/debug.log
tail: cannot open '/home/bitcoin/.bitcoin/debug.log' for reading: Permission denied
tail: no files remaining
```

VajraOfIndra:
> Personally, I don't like seeing warnings at all so perhaps we could show the command systemctl status bitcoind.service with sudo, if we decide to leave the warning and not use sudo it might be good to also explain that using sudo will remove the warning.

Adding sudo sounds good to me (we'll need to add the few lines of output corresponding), @Stadicus WDYT?

> Having recently gone through the guide myself I ran into the same permission issue.

Ok, that's weird, I wonder where is the issue coming from.. 

Could you paste the output of the following 2 commands (from user "admin"):
1. `ls -la /home/bitcoin/.bitcoin/debug.log`
1. `grep bitcoin /etc/group`

c-mannix:
> Could you paste the output of the following 2 commands (from user "admin"):

>1. ls -la /home/bitcoin/.bitcoin/debug.log
```
admin@raspibolt:~ $ ls -la /home/bitcoin/.bitcoin/debug.log
-rw------- 1 bitcoin bitcoin 10292952 Dec 31 17:22 /home/bitcoin/.bitcoin/debug.log
```

> 2. grep bitcoin /etc/group
```
admin@raspibolt:~ $ grep bitcoin /etc/group
bitcoin:x:1002:admin,electrs,btcrpcexplorer
debian-tor:x:114:bitcoin
```

VajraOfIndra:
> bitcoin:x:1002:admin,electrs,btcrpcexplorer

So your "admin" user is indeed a member of "bitcoin", so that's good.

> -rw------- 1 bitcoin bitcoin 10292952 Dec 31 17:22 /home/bitcoin/.bitcoin/debug.log

But for whatever reason, your `debug.log` can only be read/write by the bitcoin user and there is no permissions at all for any member of the bitcoin group. In comparison, here is how my file looks:
`-rw-r--r-- 1 bitcoin bitcoin 2715693 Jan  1 10:46 /home/bitcoin/.bitcoin/debug.log`  

As you can see, on my node, the file can also be read by the "bitcoin" group and also by any other users.

I'm not sure what we've done differently though... maybe someone else will be able to find out.

c-mannix:
Yeah, not sure. I followed the guide twice and came up with the same result.

RooSoft (commented):
I'd rather work to change the file's permission to allow read access to the bitcoin group... or try to access it with the bitcoin user, rather than using root to do this task.

orenfromberg:
Does anyone have an idea if this is related to [this recent change](https://github.com/raspibolt/raspibolt/commit/b845122d6e98b958d5aa6893fcef67599cdd0d57#diff-fce54494da861cdd676def495ef166aa3d6309ad930fe58975dd83ecb3aadf01R281) that replaces `Group=bitcoin` with `UMask=0027`?

VajraOfIndra:
Could it be that the latest version(s) of Bitcoin Core set up user-only permissions on this file? If so, that could explain the discrepency between newly created nodes (user-only permissions) vs older v2 nodes migrated to v3 (not sure if @Stadicus re-used its node's bitcoind data when writing v3..).

I can't weight on which is the best solutions though... (using `sudo` _vs_ changing the permissions for this file)

Stadicus:
> Does anyone have an idea if this is related to this recent change that replaces Group=bitcoin with UMask=0027?

Very interesting. I think that's a side effect of the recent change https://github.com/raspibolt/raspibolt/pull/846

Before:
* we used Bitcoin Core argument `sysperms` to set UMask `0027` (read-permissions for group) on newly created files, primarily the authentication cookie
* but this only works with the Bitcoin Core wallet disabled

After:
* to enable the Bitcoin Core wallet (for EPS, JoinMarket etc), this argument was removed. Bitcoin Core now creates new files with access rights for the user only. I did not think about the log file here.
* to make the authentication cookie group-accessible, we now execute a "chmod" command using `startupnotify` in the systemd unit.

Setting the logfile permissions once should be enough:

```sh
$ sudo chmod g+r /data/bitcoin/debug.log
```

That should be added to the guide after Bitcoin Core is started the first time, for example around here:
https://github.com/raspibolt/raspibolt/blob/32bdcbe4bd37116836ca31996de31fbc6afb4e4c/bitcoin-core.md?plain=1#L222

Stadicus:
> Adding sudo sounds good to me (we'll need to add the few lines of output corresponding), @Stadicus WDYT?

Sounds good!

RooSoft (approved):

bitcoin-core.md
```diff
@@ -220,6 +220,12 @@ Still logged in as user "bitcoin", let's start "bitcoind" manually.
 
 * Once everything looks ok, stop "bitcoind" with `Ctrl-C`
 
+* Grant the "bitcoin" group read-permission for the debug log file:
+  
+  ```sh
+  $ sudo chmod g+r /data/bitcoin/debug.log
```
VajraOfIndra, line 226:
```suggestion
  $ chmod g+r /data/bitcoin/debug.log
```
Since the command is invoked with the "bitcoin" user, we don't need to use `sudo` (not even sure we can). Simply using `chmod` should work there.

bitcoin-core.md
```diff
@@ -351,7 +357,7 @@ After rebooting, "bitcoind" should start and begin to sync and validate the Bitc
   Exit with `Ctrl-C`
 
   ```sh
-  $ tail -f /home/bitcoin/.bitcoin/debug.log
+  $ sudo tail -f /home/bitcoin/.bitcoin/debug.log
```
VajraOfIndra, line 360:
```suggestion
  $ tail -f /home/bitcoin/.bitcoin/debug.log
```
Since debug.log is now readable by admin, `sudo` is not needed anymore. I tested it on my new node.

VajraOfIndra (changes requested):
Coming back to this pull request as I'm creating a new node, so very appropriate :)

The sudo are not needed I think.

Also, despite the changes, the systemctl status still shows the `Warning: some journal files were not opened due to insufficient permissions.` error message that I believe we wanted to get rid of or warn about: I suggest simply adding sudo in front of the `systemctl status bitcoind` command, wdyt?

VajraOfIndra:
@orenfromberg Last change needed before we can merge is adding sudo to the `systemctl status bitcoind` command (the first command in the "Verification of bitcoind operations" sub-section) to get rid of the error message due to a lack of permissions (at least it still needed on my new node despite changing the permissions of the debug.log, so I assume  there are other log files)

orenfromberg:
@VajraOfIndra I'll defer to you since you are actually running the steps. :smile:

VajraOfIndra (approved):

VajraOfIndra:
> @VajraOfIndra I'll defer to you since you are actually running the steps. smile

tACK, LGTM! 
Let's merge it :)

PR #869, opened 2022-01-02
Update rtl.md

shafan:
Fix  "Add the NodeJS software repository"

#### What

### Why

Why is this change important?

#### How

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

#### Animated GIF (optional)

### DIFF
--- a/rtl.md
+++ b/rtl.md
@@ -34,7 +34,7 @@ When finished, exit the “root” user session.
 * Add the NodeJS software repository
 
   ```sh
-  $ curl -sL https://deb.nodesource.com/setup_14.x | bash -E -
+  $ curl -sL https://deb.nodesource.com/setup_14.x | sudo bash -E -
   ```
 
 * Install NodeJS using the apt package manager
--- a/rtl.md
+++ b/rtl.md
@@ -26,10 +26,9 @@ We install [Ride The Lightning](https://github.com/Ride-The-Lightning/RTL/blob/m
 
 ### Install NodeJS
 
-Starting with user “admin”, we switch to user “root” and add the Node JS package repository.
+Starting with user “admin”, we add the Node JS package repository.
 We’ll use version 14 which is well-tested with all web applications used in this guide.
 If you installed BTC RPC Explorer, then you've already accomplisehd this step.
-When finished, exit the “root” user session.
 
 * Add the NodeJS software repository
 
--- a/rtl.md
+++ b/rtl.md
@@ -33,7 +33,7 @@ If you installed BTC RPC Explorer, then you've already accomplisehd this step.
 * Add the NodeJS software repository
 
   ```sh
-  $ curl -sL https://deb.nodesource.com/setup_14.x | sudo bash -E -
+  $ curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash -
   ```
 
 * Install NodeJS using the apt package manager


Stadicus (approved):

VajraOfIndra:
We are using the command as the "root" user in the guide, so I don't think there is a need to add "sudo", correct?

Logging in as the "root" user is suggested in the comment above the command but not explicitly indicated with the command `sudo su -`... To avoid missing it out, maybe we shoud add this command?

VajraOfIndra (changes requested):
We are using the command as the "root" user in the guide, so I don't think there is a need to add "sudo", correct?

Logging in as the "root" user is suggested in the comment above the command but not explicitly indicated with the command `sudo su -`... To avoid missing it out, maybe we shoud add this command?

shafan:
If I'm not mistaken, all commands are always with sudo and not as root.
By the way, the command just after to install nodejs is also with sudo :
$ sudo apt-get install nodejs
In my opinion, it is rather the comment that should be changed, no?

VajraOfIndra:
> If I'm not mistaken, all commands are always with sudo and not as root. By the way, the command just after to install nodejs is also with sudo : $ sudo apt-get install nodejs In my opinion, it is rather the comment that should be changed, no?

I'm not sure why using "root" was choosen in the first place... there could be a good reason but I don't know what it would be!

If we could stay with "admin" and use `sudo`, I agree that it would be a better option as it's simpler.

@Stadicus : any reason to use "root" rather than "admin" to add he Node.js pasckage repo?

RooSoft:
I agree with @VajraOfIndra about running task using the `admin` and `sudo`.

shafan:
Yes I agree with that.At first reading I agree with you.
The two times where NodeJS installations appear is for "Blockchain Explorer" and "Ride The Lightning". These are two optional elements.
But what I like about this tutorial is that it's very easy to read and you can follow the steps without having to switch from one place to another.
For only two times, I'm finally of the opinion to leave it as it is.

Stadicus:
I think mentioning user "root" is a relict from the past, it should never be used (only use `sudo` for this). The user is mentioned, but not reflected by the commands.

I'd change

> Starting with user “admin”, we switch to user “root” and add the Node JS package repository

to 

> > Starting with user “admin”, we add the Node JS package repository

My bad, sorry for the confusion

rtl.md
```diff
@@ -26,15 +26,14 @@ We install [Ride The Lightning](https://github.com/Ride-The-Lightning/RTL/blob/m
...
+Starting with user “admin”, we add the Node JS package repository.
 We’ll use version 14 which is well-tested with all web applications used in this guide.
 If you installed BTC RPC Explorer, then you've already accomplisehd this step.
-When finished, exit the “root” user session.
 
 * Add the NodeJS software repository
 
   ```sh
-  $ curl -sL https://deb.nodesource.com/setup_14.x | bash -E -
+  $ curl -sL https://deb.nodesource.com/setup_14.x | sudo bash -E -
```
VajraOfIndra, line 36: Shouldn't it be `sudo -E bash -` rather than `sudo bash -E -` ? (see Node.js [instructions](https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions))

VajraOfIndra (changes requested):

rtl.md
shafan: It is the same but I can change to be more faithful to the nodejs doc

RooSoft: Unfortunately it's not... in the first version, -E is applied to the `sudo` command whereas in the second would be a `bash` argument that apparently doesn't exist.

shafan: Oops sorry you are right

VajraOfIndra: LGTM!

VajraOfIndra (approved):

VajraOfIndra:
LGTM, let's merge it.
Thanks @shafan for the contribution!

PR #873, opened 2022-01-09
Update RTL guide to v0.12.0 (+ new sample configuration file location)

VajraOfIndra:
#### What

The PR updates the RTL guide to the newest version (v0.12.0) in the download and install section. It also fixes a breaking change: the location of the sample configuration file has been moved from `~/RTL/docs` to `~/RTL/.github/docs`

### Why

* To fix a breaking change
* To keep the guide up-to-date with latest version

#### How

Updated `rtl.md`

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

Test: If possible, install RTL from scratch using the updated guide and check that the updated command and outputs are all ok.
Check that the new configuration file location is ok.

### DIFF
--- a/rtl.md
+++ b/rtl.md
@@ -104,12 +104,12 @@ We do not want to run Ride the Lightning alongside bitcoind and lnd because of s
   $ cd RTL
 
   $ git describe --tags --abbrev=0
-  > v0.11.2
+  > v0.12.0
 
-  $ git checkout v0.11.2
+  $ git checkout v0.12.0
 
-  $ git verify-tag v0.11.2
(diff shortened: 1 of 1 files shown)


AsyncLuck:
Hi @VajraOfIndra ,
I think we are already on 12.1.
Can be great to have the version control for RTL integrated in the raspibolt home script... top, to be able to check this version too when we ssh into our node. If I find sometimes I will try to do it.

AsyncLuck:
For info, #882 submited but noob to render a proper output at the end ;)

PR #882, opened 2022-01-18
System overview : add RTL

AsyncLuck:
Check RTL version and if it's running on the System overview screen

#### What

Add section in System overview to show if RTL is running and if the node is on the lastest vs

### Why

RTL is a standard part of the raspibolt v3 guide

#### How

Modify existing system overview script.
Someone needs to modify the render output section, I don't know how to do that properly. For the moment, it is dirty at the end of the output wihtout color and version control, but normally all the configs are in place to do something fancy ;)

#### Scope

- [X] independent bonus guide
- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

Nothing, simple fun stuff...


AsyncLuck:
I close it to clean up and push in the v3 branch

PR #886, opened 2022-01-21
os: enable "boot from usb" configuration

Stadicus:
#### What

Raspberry Pis are not always configured to boot from USB devices.
It has to be enabled by booting once from a microSD card containing this setting.

#### Why

This is very confusing for new users and needs to be addressed.

#### How

This change

* explains this issue
* gives simple instructions
* links to a detailed external guide

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

#### Test & maintenance

Check if the instructions make sense. No ongoing maintenance is required.


VajraOfIndra (approved):
LGTM! ACK

PR #897, opened 2022-02-03
electrs: fix update instructions

Stadicus:
#### What

Simple fix for an error in the Electrs update instructions.

### Why

The current guide does not work:

image

#### How

There's no need to specify the repository here, the command works without it.

#### Scope

- [ ] independent bonus guide
- [X] simple bug fix

#### Test & maintenance

Update Electrs using these instructions


RooSoft (approved):
Agreed, only the version is necessary as a checkout parameter

PR #899, opened 2022-02-04
Update 64bit raspiOS link to official page

asvdf:
#### What

As per title

### Why

Because Stadicus and Indra asked for it in the telegram group, obviously. Also, it is extremely important because it's my first PR.

#### How

With extreme skill.

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix
- [ x] updated link

#### Test & maintenance

Just click the link, you'll see.

#### Animated GIF (optional)

feelsgoodman.gif


asvdf:
Oh, that one, I've never used the imager before but I'll try updating the PR for it.
Just don't expect me to squash the commits as I'm doing everything from the webUI 😄

asvdf:
...done

Stadicus (changes requested):
If you want to keep the bullet point, the following lines (that should also be part of that bullet paragraph) need to be indented with two spaces as well. 

As of now, the bullet is not shown. Which is ok, I think, but then we should remove it completely.

image

Also, I'd rather use double-quotes " instead of `code` formatting (which is commonly used for stuff you need to type).

Stadicus:
Another thing that might be worthwhile to update: instead of using the keyboard shortcut `Ctrl`+`Shift`+`X`, the Image now has a simple button (the "cogwheel") for this:

image

asvdf:
...and done

VajraOfIndra:
There is a minor conflict with the instructions below. I think, we could simply add the bullet points there actually:
image

* Instead of 'select "Use custom"' and 'Select the downloaded RasPi OS disk image' we could simply repalce it with 'select "Raspberry Pi OS (Other)" and then "Raspberry Pi OS Lite (64-bit)"' )

And then in the '# Get Raspeberry Pi OS' section above, we could simply the sentence starting with "Unfortunately,..." and the first bullet point "Download the latest..."

Wdyt?

operating-system.md
```diff
@@ -33,18 +33,14 @@ As a result, it should work smoothly with Raspberry Pis while still being compat
...
-* Download the latest [RasPi OS Lite (64-bit)](https://downloads.raspberrypi.org/raspios_lite_arm64/images/){:target="_blank"} disk image.
-  Simply get the ZIP file named similarly to `2021-10-30-raspios-bullseye-arm64-lite.zip`.
-  Its size should be around 400 MB.
+## Configure boot options
 
-* Download and install the [Raspberry Pi Imager](https://www.raspberrypi.com/software/){:target="_blank"} for your regular operating system.
+Start the Raspberry Pi Imager and open the "Advanced options" by pressing the cogwheel in the bottom right corner of the application window:
 
-## Configure boot options
+![image](https://user-images.githubusercontent.com/19550140/152545918-1cfb4869-87f5-4b4f-bcdb-5da6fb7d148b.png)
```
VajraOfIndra, line 42: The image could be uploaded in "Images" directory in the root folder of the repo and then this URL can be used in the link.

Nitpicking here.., but at this stage the OS is likely not yet selected, so maybe the screenshot should be one with no OS selected yet.

VajraOfIndra (changes requested):

operating-system.md
Stadicus: We definitely must not use the image hosted on GitHub, but upload and reference it ourselves.

Stadicus:
BTW: I feel we're pinning @asvdf down with all our additional change requests. 

I think it's fine if you just do the simple change first, and someone else can then add all other things afterwards. We don't want to scare you away... :-P

Stadicus (approved):
tACK, let's go to master! :)

VajraOfIndra (approved):

PR #900, opened 2022-02-04
Remove qrencode package installation from core guide as not used

VajraOfIndra:
#### What

This PR is a proposal to remove the installation of the `qrencode` package in the core guide.

Note: what about `jq`, do we use it in the core guide..? I can't find where.

### Why

The package is not used at any point in the core guide, so no reason to install it by default (security risk)

#### How

Deletedt the package name from the command

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

#### Test & maintenance

Just need to make sure that we indeed do not need this package anywhere in the core guide

#### Animated GIF (optional)

QR

### DIFF
--- a/system-configuration.md
+++ b/system-configuration.md
@@ -142,7 +142,7 @@ The “Advanced Packaging Tool” (apt) makes this easy.
 * Make sure that all necessary software packages are installed:
 
   ```sh
-  $ sudo apt install wget curl gpg git htop jq qrencode --install-recommends
+  $ sudo apt install wget curl gpg git htop jq --install-recommends
   ```
 
 ---


RooSoft:
A quick search for `qrencode` reveals

```
bonus/raspberry-pi/upgrade-external-drive.md
74:$ sudo apt install htop git curl bash-completion jq qrencode dphys-swapfile hdparm --install-recommends

system-configuration.md
145:  $ sudo apt install wget curl gpg git htop jq qrencode --install-recommends
```

It definitely isn't called directly... could it be a dependency for something else? My guess is that it's almost impossible to know unless someone creates a brand new raspibolt from scratch without `qrencode`.

AsyncLuck:
it's maybe used by lnconnect and for Zeus wallet config ? But maybe not... no mention in the guide or on lnconnect github. (but it's in the bonus section anyway)

RooSoft:
Why has this dependency been added in the first place?

Stadicus:
I think `qrencode` was used in the past to encode the LND connection strings and connect e.g. Zap Wallet. Not used atm, and can always be installed in the section where it's actually used.

Stadicus (approved):
utACK :rocket:

PR #901, opened 2022-02-04
Fixed path to RTL sample config file

VajraOfIndra:
#### What

Fix the path of RTL sample config file.

### Why

The sample file has once again moved location (it was recently put in the `/docs` directory which does not exist anymore and it's back in the root directory).

#### How

fixed path

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

#### Test & maintenance

Check that the corrected path is the right one

### DIFF
--- a/rtl.md
+++ b/rtl.md
@@ -149,7 +149,7 @@ Now we take the sample configuration file and add change it to our needs.
 * Copy the sample config file, and open it in the text editor.
 
   ```sh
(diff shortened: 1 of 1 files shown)


Stadicus (approved):
utACK :+1:

RooSoft:
Confirmed, the file now resides in RTL's root folder

PR #902, opened 2022-02-05
Updated list of default packages to install

VajraOfIndra:
#### What

This is a proposal to update the list of packages we install in the "System configuration" of the guide: probably only `git` need to be installed.

### Why

Presently, the command installs the following packages: `wget` `curl` `gpg` `git` `htop` `jq` 

* Based on the [list of packages included in RaspiOS Lite 64-bit](https://downloads.raspberrypi.org/raspios_lite_arm64/images/raspios_lite_arm64-2022-01-28/2022-01-28-raspios-bullseye-arm64-lite.info), `wget`, `curl`, `gpg` and `htop` are already shipped with the OS and therefore do probably not need to be installed (moreover, `htop` is never used in the core guide)... or do they..?

* `jq` is not part of the OS but is not used in the core guide, only in bonus guides (albeit several of them: system overview, white paper, lnchannels, lnbalance). Therefore, `jq` should probably be installed only by users that want to install one of the bonus guide.

* `git`: git is used in the core guide to install electrs, the block explorer and RTL and therefore needs to be installed

Wdyat?

#### How

* Updated the command to install `git` only and updated the overlying comment accordingly.
* Added a `jq` installation step for the bonus guides that depends on it

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

#### Test & maintenance

Make sure the removed packages are indeed already part of the OS or never used in the core guide.

#### Animated GIF (optional)

packages

### DIFF
--- a/bonus/bitcoin/white-paper.md
+++ b/bonus/bitcoin/white-paper.md
@@ -51,6 +51,13 @@ This guide explains how to reconstruct the Bitcoin white paper PDF using your ow
   $ mkdir bitcoin-white-paper
   $ cd bitcoin-white-paper
   ```
+  
+* Install jq, a JSON processor that will be used to parse the transaction data
+
+  ```sh
+  $ sudo apt update
+  $ sudo apt install jq --install-recommends
+  ```
 
 * Use `bitcoin-cli` to download and create the PDF
 
--- a/bonus/bitcoin/white-paper.md
+++ b/bonus/bitcoin/white-paper.md
@@ -52,7 +52,7 @@ This guide explains how to reconstruct the Bitcoin white paper PDF using your ow
   $ cd bitcoin-white-paper
   ```
   
-* Install jq, a JSON processor that will be used to parse the transaction data
+* Install `jq`, a JSON processor that will be used to parse the transaction data
 
   ```sh
   $ sudo apt update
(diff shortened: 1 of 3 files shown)


Stadicus:
I don't think there's much merrit to reduce the initial installation of packages for tools that are actually used in the core repository. While some packages are already installed in RasPiOS Lite, this is not necessarily true for other distributions, even if they're Debian-based. 

If a package is already installed, there's no downside to running `apt install` for it again, just to make sure.

* Standard tools like `curl`, `wget`, `git` or `gpg` should be installed upfront
* Ack to install `jq` only when needed 
* Ack on removing `htop`, it's something I use a lot personally, but is not part of the guide

Stadicus (approved):
utACK

VajraOfIndra:
> I don't think there's much merrit to reduce the initial installation of packages for tools that are actually used in the core repository. While some packages are already installed in RasPiOS Lite, this is not necessarily true for other distributions, even if they're Debian-based.
> 
> If a package is already installed, there's no downside to running `apt install` for it again, just to make sure.
> 
>     * Standard tools like `curl`, `wget`, `git` or `gpg` should be installed upfront
> 
>     * Ack to install `jq` only when needed
> 
>     * Ack on removing `htop`, it's something I use a lot personally, but is not part of the guide

Makes sense. Amended accordingly!

PR #841, opened 2021-12-14
Split the 'Additional scripts' bonus guide into three bonus guides: Aliases, lnbalance and lnchannels

VajraOfIndra:
#### What

This PR aims at three objectives:

* Split the 'Additional scripts' bonus guide into three separate guides as it contains guides that belong to both the +Raspberry Pi and + Lightning bonus sections
  * aliases.md in the '+Raspberry Pi' section -> Explains how to set up aliases to simplify command entry
  * lnbalance.md in the '+Lightning' section -> Explains how to install the lnbalance script
  * lnchannels.md in the '+Lightning' section -> Explains how to install the lnchannels script
* Updates the content of the guides to match v3 styling and remove outdated references (e.g. testnet)
* Add aliases suggestions by MarcosDub in this [Telegram post](https://t.me/raspibolt/4466)

### Why

This change of structure makes the three sub-guides much more visible and placed in their proper logical sections of the Bonus Section.
Also the guides were outdated and needed some updating both in content and style.
It also allows to add a contribution by MarcosDub re. additional aliases

#### How

Split the 'Additional scripts' guide intro three new bonus guide and deleted the original guide.
Tested on a v3 node (some but not all aliases have been tested)

#### Possible improvements

I placed the aliases.md guide in the 'System administration' heading within '+ Raspberry Pi' section... but I'm not too sure this sub-heading is the best, any better idea what kind of sub-heading to use under which setting up aliases would fall under..?

#### Scope

- [ ] significant change to core configuration
- [x] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

Test: Follow the three new bonus guides

Maintenance: None

#### Animated GIF (optional)

chop

PR #903, opened 2022-02-06
Split the 'Additional scripts' bonus guide into three bonus guides: Aliases, lnbalance and lnchannels (new PR without conflicts)

VajraOfIndra:
Identical to #841 but without merging conflicts.


Stadicus:
Seems like the following two images are missing:

* images/alias-example.png
* images/script-lnchannels.png

Otherwise, all seems in order. Although I didn't do a proper diff... :)

bonus/raspberry-pi/aliases.md
```diff
@@ -0,0 +1,189 @@
...
+  ##################
+  # GENERAL STATUS #
+  ################## 
+  
+  alias livehealth='sudo watch -n 1 "vcgencmd measure_clock arm; vcgencmd measure_temp"'
+ 
+  alias showversion='echo The installed versions of the services are as follows: ; \
+    bitcoind --version ; \
+    lnd --version ; \
+    echo "BTC RPC Explorer: $(jq '.version' < /home/btcrpcexplorer/btc-rpc-explorer/package.json)" ; \
```
RooSoft, line 61: The `/home/btcrpcexplorer` folder doesn't exist on my raspibolt... am I missing something?

RooSoft: Oh, the alias requires some btc explorer to be installed... might be a good idea to explain dependencies those alias have...

RooSoft (approved):
All good to me aside from a minor comment that could end up creating a different PR...

RooSoft:
> Seems like the following two images are missing:
> 
> * images/alias-example.png
> * images/script-lnchannels.png
> 
> Otherwise, all seems in order. Although I didn't do a proper diff... :)

Good catch... I concur.

bonus/raspberry-pi/aliases.md
Stadicus: Good point, if some application is not installed, an error is thrown. This can be avoided by at least piping the error message into nirvana (/dev/null):
```
cat /home/btcrpcexplorer/btc-rpc-explorer/package.json 2>/dev/null | jq '.version
```

VajraOfIndra:
> Seems like the following two images are missing:
> 
>     * images/alias-example.png
> 
>     * images/script-lnchannels.png
> 
> 
> Otherwise, all seems in order. Although I didn't do a proper diff... :)

Good catch! I added them. I also fixed the path for the picture for lnbalance.

bonus/raspberry-pi/aliases.md
VajraOfIndra: I added this solution to the Explorer and RTL lines. For Electrs, not sure how to do it so I'll let ou add it as a separate PR if you wish so.

VajraOfIndra:
Thanks all for the reviews, I'm now merging the PR!

PR #904, opened 2022-02-07
/data owned by admin, and bitcoin user creation belongs in the Bitcoin section

RooSoft:
/data owned by admin, and bitcoin user creation belongs in the Bitcoin section

#### What

Postpone the bitcoin user creation until it is being used

### Why

Considering writing a testnet version of the bitcoin section... it would be pretty similar except for the user creation. So to keeps things consistent, might be a good idea to create the bitcoin user in the Bitcoin section.

#### How

Removed the bitcoin user creation in the `System Configuration` section, moved it to the `Bitcoin` section. A tad below, in the former section, the `/data` folder was created and owned by the bitcoin user. This was part of another discussion, wondering if that's the right thing to do. As that user simply doesn't exist at that point anymore, it makes much more sense to give that folder to the `admin` user.

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

I fear it might be a good idea to create a brand new Raspibolt from scratch to make sure it works.

### DIFF
--- a/bitcoin-core.md
+++ b/bitcoin-core.md
@@ -95,7 +95,7 @@ You can read more on [How to securely install Bitcoin](https://medium.com/@luked
 The Bitcoin Core application will run in the background as a daemon and use the separate user “bitcoin” for security reasons.
 This user does not have admin rights and cannot change the system configuration.
 
-* Create the user "bitcoin"
+* With user "admin", create the Bitcoin data folder
 
   ```sh
   $ sudo adduser --gecos "" --disabled-password bitcoin
--- a/bitcoin-core.md
+++ b/bitcoin-core.md
@@ -95,7 +95,7 @@ You can read more on [How to securely install Bitcoin](https://medium.com/@luked
 The Bitcoin Core application will run in the background as a daemon and use the separate user “bitcoin” for security reasons.
 This user does not have admin rights and cannot change the system configuration.
 
-* With user "admin", create the Bitcoin data folder
+* Create the user bitcoin
 
   ```sh
   $ sudo adduser --gecos "" --disabled-password bitcoin
@@ -116,7 +116,7 @@ Instead of creating this directory, we create a data directory in the general da
 
   ```sh
   $ mkdir /data/bitcoin
-  $ chown bitcoin. /data/bitcoin
+  $ chown bitcoin:bitcoin /data/bitcoin
   ```
 
 * Switch to user "bitcoin"
@@ -126,7 +126,7 @@ Instead of creating this directory, we create a data directory in the general da
   ```
 
 * Create the symbolic link `.bitcoin` that points to that directory
-`
+
   ```sh
   $ ln -s /data/bitcoin /home/bitcoin/.bitcoin
   ```
--- a/bitcoin-core.md
+++ b/bitcoin-core.md
@@ -90,11 +90,40 @@ This is a precaution to make sure that this is an official release and not a mal
 🔍 *Verifying signed software is important, not only for Bitcoin.
 You can read more on [How to securely install Bitcoin](https://medium.com/@lukedashjr/how-to-securely-install-bitcoin-9bfeca7d3b2a){:target="_blank"} by Luke-Jr.*
 
+### Create the bitcoin user
+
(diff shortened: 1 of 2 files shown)


bitcoin-core.md
```diff
@@ -90,26 +90,49 @@ This is a precaution to make sure that this is an official release and not a mal
...
+  $ sudo adduser admin bitcoin
+  ```
 
 ### Create data folder
 
 Bitcoin Core uses by default the folder `.bitcoin` in the user's home.
 Instead of creating this directory, we create a data directory in the general data location `/data` and link to it.
 
-* Switch to user "bitcoin"
+* Switch to user "admin"
```
Stadicus, line 115: I don't think that's nessacary: you're already in the "admin" user session

More high-level: users always log in with "admin", so there never should be the need to switch to this session (just quit a child session with `exit`). It's not even possible, as you need "sudo" to switch sessions, and user "admin" is the only user with that privilege.

bitcoin-core.md
```diff
@@ -90,26 +90,49 @@ This is a precaution to make sure that this is an official release and not a mal
...
 
   ```sh
-  $ sudo su - bitcoin
+  $ sudo su - admin
   ```
 
 * Create the Bitcoin data folder
```
Stadicus, line 121: To make sure the right user session is used, I usually prefix a critical command like this:

> With user "admin", create the Bitcoin data folder

system-configuration.md
```diff
@@ -48,21 +48,6 @@ We will use the primary user "admin" instead of "pi" to make this guide more uni
   $ sudo adduser admin sudo
   ```
```
Stadicus, line 50: Idea, but can also be a follow-up pull request: How about rebooting here, and logging in with user "admin"?

* No need to `export LC_ALL=C` twice (first instance can be deleted)
* No need to change ownership of `/data/` below, as it's already created by user "admin"

Stadicus (commented):
I like this change, it makes sense and improves modularity.

Seemingly small changes like this can lead to follow-up issues, so it's important to test it well and look for hidden caveats. 

I left a few comments.

bitcoin-core.md
RooSoft: Removed the statement...

bitcoin-core.md
RooSoft: Fixed

system-configuration.md
RooSoft: Not a fan of reboots, personally... can't hurt if nothing else could be disrupted on the system, but it's a good practice in Linux land to have as much uptime as possible.

RooSoft: Removed the second `export LC_ALL=C`

RooSoft: Removed the `chown`, good catch

Stadicus: Sorry, I didn't mean a physical reboot. I wanted to say "exit the `pi` session and log in again as `admin`". That's necessary at some point anyway.

RooSoft: Ok, thus, the command `exit`

RooSoft: Not even, sorry, at that point, we're in the `pi` user's shell and need to close it and reopen a session with the new `admin` user. Is that what you mean?

RooSoft: Added two steps... let me know if it makes sense

Stadicus (changes requested):
consistency fixes

system-configuration.md
* move "logout" / "login" instructions up from section System update
* Remove first config of LC_ALL for user "pi"
* Re-add second config of LC_ALL for user "admin"
* Re-add 'sudo' explanation
* minor wording adjustments

bitcoin-core.md
* use chown format "user:group"
* wording fixes

Stadicus:
consistency fixes

system-configuration.md
* move "logout" / "login" instructions up from section System update
* Remove first config of LC_ALL for user "pi"
* Re-add second config of LC_ALL for user "admin"
* Re-add 'sudo' explanation
* minor wording adjustments

bitcoin-core.md
* use chown format "user:group"
* wording fixes

Stadicus (approved):
I think this is ready to merge :)

PTAL @RooSoft as a final check

PR #905, opened 2022-02-09
Removed outdated password reference in LND guide

VajraOfIndra:
#### What

Removes an outdated reference to "Password [D]" in the LND guide (a relict from v2)

#### Why

Outdated reference that can confuse new users.

#### How

* Updated `lnd.md`

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

#### Test & maintenance

Double-check that the modified sentence is correct.

### DIFF
--- a/lnd.md
+++ b/lnd.md
@@ -281,7 +281,7 @@ Once LND is started, the process waits for us to create the integrated Bitcoin w
   ...
   ```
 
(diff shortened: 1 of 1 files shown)


Stadicus (approved):
I think we could even remove the "(combined with your optional passphrase)", but it's also ok to keep it.

PR #909, opened 2022-02-09
Sudo for file creation

RooSoft:
#### What

Can't change user to bitcoin without using sudo

### Why

The guide simply doesn't work as is

#### How

Add sudo

#### Scope

- [ ] independent bonus guide
- [X] simple bug fix

Fixes # (link issue)


RooSoft:
Doesn't apply to the current context... closing.

PR #913, opened 2022-02-11
Update privacy.md

the-jade-rabbitt:
#### What

When following the v3 RaspiBolt guide to (optionally) SSH in to the raspibolt via Tor I hit a problem when following the advice pointed to here:
https://gitlab.torproject.org/legacy/trac/-/wikis/doc/TorifyHOWTO/Putty

### Why

I was unable to connect to my pi node following the linked-to guide.
The advice on the linked-to page was to connect to port 9050 which didn't work for me.

#### How

Because I was SSHing via Putty through Tor on windows with Tor browser running, I think Tor browser used Tor Bundle with port 9150 rather than 9050?
I think this is just Tor browser-bundle specific. Tor itself still listens on 9050

#### Scope

- [x] simple bug fix

#### Test & maintenance

None

### DIFF
--- a/privacy.md
+++ b/privacy.md
@@ -124,7 +124,7 @@ A few examples:
 
 * **Windows**: configure PuTTY as described in this guide [Torifying PuTTY](https://gitlab.torproject.org/legacy/trac/-/wikis/doc/TorifyHOWTO/Putty){:target="_blank"} by the Tor Project.
 
-  * **Note:** If you are using Putty and fail to connect to your pi by setting port 9050 in the Putty proxy settings, try setting port 9150 instead.
+  * **Note:** If you are using PuTTy and fail to connect to your Pi by setting port 9050 in the PuTTy proxy settings, try setting port 9150 instead.
   * Also make sure to run Tor in the background by installing and running the Tor browser.
 
 * **MacOS and Linux**: use `torify` or `torsocks`.
--- a/privacy.md
+++ b/privacy.md
@@ -124,8 +124,7 @@ A few examples:
 
 * **Windows**: configure PuTTY as described in this guide [Torifying PuTTY](https://gitlab.torproject.org/legacy/trac/-/wikis/doc/TorifyHOWTO/Putty){:target="_blank"} by the Tor Project.
 
-  * **Note:** If you are using PuTTy and fail to connect to your Pi by setting port 9050 in the PuTTy proxy settings, try setting port 9150 instead.
-  * Also make sure to run Tor in the background by installing and running the Tor browser.
+  * **Note:** If you are using PuTTy and fail to connect to your Pi by setting port 9050 in the PuTTy proxy settings, try setting port 9150 instead. When Tor runs as an installed application instead of a background process it uses port 9150.
 
 * **MacOS and Linux**: use `torify` or `torsocks`.
   Both work similarly; just use whatever you have available:
--- a/privacy.md
+++ b/privacy.md
@@ -124,6 +124,9 @@ A few examples:
 
 * **Windows**: configure PuTTY as described in this guide [Torifying PuTTY](https://gitlab.torproject.org/legacy/trac/-/wikis/doc/TorifyHOWTO/Putty){:target="_blank"} by the Tor Project.
 
+  * **Note:** If you are using Putty and fail to connect to your pi by setting port 9050 in the Putty proxy settings, try setting port 9150 instead.
+  * Also make sure to run Tor in the background by installing and running the Tor browser.
+
 * **MacOS and Linux**: use `torify` or `torsocks`.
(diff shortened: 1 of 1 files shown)


privacy.md
```diff
@@ -124,6 +124,9 @@ A few examples:
 
 * **Windows**: configure PuTTY as described in this guide [Torifying PuTTY](https://gitlab.torproject.org/legacy/trac/-/wikis/doc/TorifyHOWTO/Putty){:target="_blank"} by the Tor Project.
 
+  * **Note:** If you are using Putty and fail to connect to your pi by setting port 9050 in the Putty proxy settings, try setting port 9150 instead.
+  * Also make sure to run Tor in the background by installing and running the Tor browser.
```
VajraOfIndra, line 128: IMO, this is a bit redundant as the guide already states just above: "You also need to have Tor installed on your regular computer where you start the SSH connection." wdyt?

privacy.md
```diff
@@ -124,6 +124,9 @@ A few examples:
 
 * **Windows**: configure PuTTY as described in this guide [Torifying PuTTY](https://gitlab.torproject.org/legacy/trac/-/wikis/doc/TorifyHOWTO/Putty){:target="_blank"} by the Tor Project.
 
+  * **Note:** If you are using Putty and fail to connect to your pi by setting port 9050 in the Putty proxy settings, try setting port 9150 instead.
```
VajraOfIndra, line 127:
```suggestion
  * **Note:** If you are using PuTTy and fail to connect to your Pi by setting port 9050 in the PuTTy proxy settings, try setting port 9150 instead.
```
Fixed a few typos.

Also, is 9050 for when Tor is running as a background service and 9150 for when it's not but the Tor Briwser is opened? If that's the case we could simply explains this, wdyt?

VajraOfIndra (commented):
Thanks for your proposal!

I'm wondering when is 9050 and 9150 needed, From what I remembered from another section of the guide, the difference is whether Tor is running in the background or not, correct? if that's the case, we could maybe explains this in your bullet point instead. I'm not sure what's the simplest approach here.

+ a couple of fixes for spelling consistency

the-jade-rabbitt:
Hi @VajraOfIndra, thanks for the typo fixes - added them to my request.
Yes, I think this bit is redundant "* Also make sure to run Tor in the background by installing and running the Tor browser."
But it would have helped me to know about port 9150 because I ran Tor on Windows as an application rather than as a background service, maybe others will too - I think you're right, port 9150 is used by Tor browser when installed and run as an application rather than as a background service.
I will update the pull request to reflect this.

VajraOfIndra:
> Hi @VajraOfIndra, thanks for the typo fixes - added them to my request. Yes, I think this bit is redundant "* Also make sure to run Tor in the background by installing and running the Tor browser." But it would have helped me to know about port 9150 because I ran Tor on Windows as an application rather than as a background service, maybe others will too - I think you're right, port 9150 is used by Tor browser when installed and run as an application rather than as a background service. I have added a new commit to my request to request to reflect this.

I think your proposal is a useful addition, I approved your changes.

Ultimately, I feel the guide would benefit from a clearer way to explain how to use the various Tor hidden services: i.e. clarifying the difference between installing the Tor Browser vs a background service, how to install them on Windows/MAC/Linux, the implications for which port to use (9050|9150), what are proxy servers used for in some of the apps etc etc. I'm not sur what would be the best way to do it though. Maybe a separate guide dedicated to Tor could help really help lay down the fundamentals and would make the user better understands what choice they have to make later on at various steps of the guide. But that's just blue sky thinking for later work! :) For now, this addition could help users who are stuck because of the wrong port. 

@Stadicus Are you happy with this addition to be merged to the main guide?

VajraOfIndra (approved):

the-jade-rabbitt:
@VajraOfIndra I'm a bit new to this, should I rebase this pull request on top of all the new commits others are making to master?
Or just leave it as it is now?

VajraOfIndra:
I think you're ok for now. Your PR can be merged as is (as you can see below there is no conflicts). But you could fetch the new commits from master to your branch if you want to to keep it updated anyways... Tbh, I'm failry new to this as well so you might get some better advice from more experienced git users :)

Although the PR is ready to bemerged as it has been approved, we still need a review/approval from @Stadicus as this PR pertains to the core guide.

Stadicus (approved):

VajraOfIndra:
Thanks a lot for your contribution @the-jade-rabbitt! We're going to master! :tada:

the-jade-rabbitt:
> Thanks a lot for your contribution @the-jade-rabbitt! We're going to master! 🎉

Thank you @VajraOfIndra, I've really enjoyed learning from the guide.  Very happy that I can give back.
:sparkles::rabbit2::sparkles:

PR #929, opened 2022-02-17
Fixes bug with SCB repo cloning destination

### LINKED ISSUE #928: [BUG] Guide: lightning backup via github: Wrong Path
### Describe the bug
In the section:
https://raspibolt.org/static-channel-backup.html#test
The user should switch to a path:
`cd /data/lnd/remote-lnd-backup`
But this path does not exist.

### Possible solution
When the git clone command is executed in the home directory you can change the command in the guide like this:
`cd remote-lnd-backup/`
Otherwise you have to state in the guide, that this branch needs to be created beforehand.

VajraOfIndra:
#### What

This PR fixes a bug related to the location of the local GitHub repository for the remote backup of the SCB. As it is, the repo is cloned in `/home/lnd` instead of `/data/lnd`

### Why

The remote backup guide does not work as is.

#### How

I propose a simple `cd /data/lnd` just before cloning the repo + some minor related changes due to this added step

An alternative would be to add a destination argument to the `git clone` command. However, I encountered some issues with this technique. So I wonder if moving first to the desired directory might not be more fool-proof..

Wdyat?

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

Fixes #928 

#### Test & maintenance

Try out the new commands and see if it works as intended.

### DIFF
--- a/static-channel-backup.md
+++ b/static-channel-backup.md
@@ -129,10 +129,10 @@ We create a shell script to monitor `channel.backup` and make a copy to our back
   REMOTE_BACKUP_ENABLED=false
 
   # Locations of source SCB file, formatted backup files and Git repo
-  SOURCEFILE="/home/lnd/.lnd/data/chain/bitcoin/mainnet/channel.backup"
+  SOURCEFILE="~/.lnd/data/chain/bitcoin/mainnet/channel.backup"
   LOCAL_BACKUP_FILE="/mnt/static-channel-backup-external/channel-$(date +"%Y%m%d-%H%M%S").backup"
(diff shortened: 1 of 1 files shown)


RooSoft (approved):
Makes a lot of sense

VajraOfIndra:
> Makes a lot of sense

I propose we use relative paths rather than absolute ones for the LND data folder in order to make the guide a bit more universal, i.e. easier to adpat to other directory structures...

Wdyt? We just need to make sure it does not break anything..

RooSoft:
I was wondering why we're using absolute paths everywhere...

Stadicus (approved):
utACK

PR #940, opened 2022-02-24
Harmonize left menu headings with generic names (#938)

Stadicus:
Follow-up change to https://github.com/raspibolt/raspibolt/pull/938#event-6131425079

Original change description by @VajraOfIndra 

---

#### What

See discussion in #930 

_tl;dr_: The PR updates the left menu headings to make them generic rather than program-specific.

### Why

* Consistency/harmonization
* Makes it easier to switch a program with another one in the future if desired, without changing the guide's main structure

### How

* Changed file names: `*.md`
* Changed yml titles
* Changed h1 titles
* Changed the title/path of the "Next" links at the bottom of the page

### Scope

- [x] significant change to core configuration
- [ ] independent bonus guide
- [ ] simple bug fix

Fixes #930 

#### Test & maintenance

Make sure there is no broken links and that all relevant files have been updated with the proper name.

names

Issue #930, opened 2022-02-18
[DISCUSSION]: Standardization of Bitcoin and Lightning sub-headings: generic names vs program names


VajraOfIndra:
## Introduction

As stated by @Stadicus in this [comment](https://github.com/raspibolt/raspibolt/pull/921#issuecomment-1044668250), it is preferrable to have generic names for the left menu headings:

> Regarding "Generic" vs. "Application-specific" titles: I strongly prefer the generic titles, at least when there's a choice. This also affects the filenames, and we can replace individual applications in the core guide easier if necessary.

It does indeed make the guide more future-proof. However, presently, there are still a couple of headings that derive their names from the chosen program to install on the node, namely:
* Bitcoin Core
* LND
* Ride The Lightning

NB: Interestingly I assume "Electrum server" has now become a generic name! E.g. Fulcrum could be categorized as an Electrum server, correct? I looked around to see if there was a more generic name without the term 'Electrum' in it but coulnd't find anything at all. Any thoughts on this? Feel freet to bring this in the discussion below and propose alternative names.

NB 2: In this post, I'll focus on the "Bitcoin" and "Lightning" chapters and will ignore the "Raspberry Pi" chapter of the guide, although the heading "Raspberry Pi" is more specific than generic... Feel freet to bring this in the discussion if you wish so.

---

## New names proposals

I propose below a few ideas to homogenize the left menu headings with the following goals:
* *Generic rather than specific*: All headings are have generic names and not program names
* *Industry standard*: The generic name are well-known and accepted categories in the bitcoin community
* *Short*: The names are kept short for ease of display and navigation

### Program: Bitcoin Core

* _Existing_: **Bitcoin Core**
* _Proposed_:  **Bitcoin Node**
* _Alternative_: **Bitcoin Node Implementation** -> Quite long (also not an implementation per se as there is no specifications, correct?.. but that's pedantry now :)
* Alternative: **Bitcoin Client** -> but it's more than just a client...
* Alternative: **Bitcoin Core** -> we keep it as is, see [comment](https://github.com/raspibolt/raspibolt/pull/921#issuecomment-1044668250). IMHO, I don't think it's needed to keep the program name as the generic names above looks suitable.

### Program: Electrs

* _Existing_: **Electrum server** -> 
* _Proposed_: **Electrum Server** -> Keep existing name. Unless someone can come up with a more generic name that does not use the term 'Electrum', if that even exists.

### Program: BTC RPC Explorer

[...]

PR #941, opened 2022-02-25
New .md file structure

VajraOfIndra:
#### What

See discussion in #940 and #930 
A new file structure that is recommended for .md files in GitHub repos (e.g. [Just The Docs](https://pmarsceill.github.io/just-the-docs/docs/navigation-structure/#pages-with-children))

#### How

* changed file name paths
* updated internal link paths
* + extra bug fixes

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

There is a need for a thorough review of the navigation, left menu bar, and also identifying dead links.


Stadicus:
I like the idea to organize the different sections in different directories, but I'm not a fan of putting the whole guide in a generic `/docs` folder. This is a very standard folder name commonly used with software projects that have some documentation attached to them as an afterthought. But for the RaspiBolt, the documentation is the whole project.

Looking at the current folder structure, I realized that we don't really need the `/raspibolt` directory. Using this could be an option for me. Or we could even use a more explicit name, like `/raspibolt-guide`.

WDYT?

RooSoft:
This could be an unpopular take here but... as much as I agree with the fact that this is a documentation project... I tend to avoid as much as possible having root folders packed with files and folders with different use cases.

VajraOfIndra:
> I like the idea to organize the different sections in different directories, but I'm not a fan of putting the whole guide in a generic `/docs` folder. This is a very standard folder name commonly used with software projects that have some documentation attached to them as an afterthought. But for the RaspiBolt, the documentation is the whole project.
> 
> Looking at the current folder structure, I realized that we don't really need the `/raspibolt` directory. Using this could be an option for me. Or we could even use a more explicit name, like `/raspibolt-guide`.
> 
> WDYT?

I don't know enough about the open-source world best practices to have a valuable opinion here... but your suggestion of replacing `/docs` by `/raspibolt-guide` sounds good as well. I don't have a strong opinion here tbh.


> This could be an unpopular take here but... as much as I agree with the fact that this is a documentation project... I tend to avoid as much as possible having root folders packed with files and folders with different use cases.

What do you mean by "with different use cases"? I'm not sure I understand what you mean in the context of the new proposed file structure and question over the name of the `/docs` folder.. What'd be your favourite option(s)?

RooSoft:
`/raspibolt-guide` seems a tad longer that it has to be... IMO `guide` would be enough if `doc` is a controversial choice.

VajraOfIndra:
> `/raspibolt-guide` seems a tad longer that it has to be... IMO `guide` would be enough if `doc` is a controversial choice.

I like `guide`, it's a nice compromise as it's short and neat while still keeping a lot of meaning in term of what it contains... I'll wait for @Stadicus to confirm his preferred choice before amending the PR.

I assume I can also safely delete the `/raspibolt` folder and what it contains (a `README.md` duplicate), correct?

RooSoft:
> I assume I can also safely delete the `/raspibolt` folder and what it contains (a `README.md` duplicate), correct?

I tend to agree with this statement

RooSoft:
As far as images are concerned, I have a proposition....

Since folders will tend to be less crowdy than the current structure in root... I suggest we directly put images in the same directory as their `.md` file. In short, I'd nuke the `/images` folder...

That's based on my programmer background. I like having `js` and `css` files as near as possible to their `html` file when it's the sole user of it. Seems fitting to do the same here with docs and images...

Would not make sense in the current context, but since we're creating a cleaner directory structure, that's what I'd do.

RooSoft:
Would make sense to keep images in their own folder if they were intended to be reused between docs. Probably is the case for some of them, but I suspect that most image only have one purpose... wonder why sending them farther than they need to be.

Stadicus:
> I suggest we directly put images in the same directory as their .md file.

Not sure about this. I can understand it in a programming context, but I'd rather keep the `/guide` folder optimized to be read by humans. The new document structure will look very neat. Now imagine putting a dozen images in the same folder... 

image

I thought about putting `./img/` folders either in the `/guide` folder (so that it's like a www root) or into the individual folders. But with what goal? I find the single image repository in `/images` a pretty pragmatic solution.

VajraOfIndra:
> I thought about putting ./img/ folders either in the /guide folder (so that it's like a www root) or into the individual folders. But with what goal?

Putting the images folder inside `/guide` would shorten the image links inside the .md files (_i.e._ removing one level of `../`)... so we would gain a bit of readability and less chance of typing wrong links maybe, but then again, it's a lot of work for a minor gain.

I'll go with whatever you decide guys! But I'll wait until a decision has been made to do all the fixes at once!

Stadicus:
> I'll go with whatever you decide guys

I personally would leave the `/image` repository. If you view the MD files rendered, the images are shown. If you read the raw MD text, that's what the alternative text description would be for. To shorten the image URL, we could also use absolute links like `/images/...` instead of relative paths.

Stadicus:
I just moved all MD files to the `/guide` folder, I think that's a huge pain in the web interface. Let me know if I should do some additional batch moving/renaming, I'm happy to assist.

VajraOfIndra:
> I just moved all MD files to the `/guide` folder, I think that's a huge pain in the web interface. Let me know if I should do some additional batch moving/renaming, I'm happy to assist.

You just saved me an hour of work, thank you! Yes, in the web interface it has to be done one file at a time. Reminds me to move to desktop editing and git!

Stadicus:
FYI: I'm currently wrestling with the `include_metatags.md`, as this gives a Jekyll error when publishing to Github Pages.

Stadicus:
I tried to replace all relative image links (e.g. `../../../image/whatever.png`) with an absolute link. If we want to keep the `/images` folder, this would make the overall linking much more robust. 

The problem is that when you fork the repository and publish to your own GitHub Pages site, the whole thing is put in a sub-folder, e.g. https://stadicus.github.io/raspibolt-dev/ And that means, that an absolute path starting from the root folder does not work... not sure how to solve that. Just FYI.

Stadicus:
The current state is now published to https://stadicus.github.io/raspibolt-dev/, which can be used to run it against broken link checkers.

Stadicus:
I fixed all broken links and added back two missing images in old the "EPS" bonus guide (probably deleted by mistake). This PR is published here, PTAL
https://stadicus.github.io/raspibolt-dev/

VajraOfIndra:
> I fixed all broken links and added back two missing images in old the "EPS" bonus guide (probably deleted by mistake). This PR is published here, PTAL https://stadicus.github.io/raspibolt-dev/

Looking great, LGTM. 
Shall we merge this branch into the `generic-names` branch and take it from there for the final checks before merging into `master`?

PR #942, opened 2022-02-26
Update static-channel-backup.md

the-jade-rabbitt:
#### What

The recommended scb-backup script in v3 RaspiBolt guide fails with "No such file or directory"
I am still at the point in the guide of running this script with both local and remote backup methods disabled.

### Why

tilde character in the path does not work inside quotes.

#### How

I'm suggesting replacing ~ with $HOME which is more universal and avoids the error in this case.
You could just remove the quotes but that wouldn't work where there are spaces in the string assignments.

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

#### Test & maintenance

I no longer get the error when I run the script as user lnd with my proposed changes.
The scb-backup.service status now shows as running (it was getting 'run too frequently' before)

### DIFF
--- a/guide/lightning/channel-backup.md
+++ b/guide/lightning/channel-backup.md
@@ -119,7 +119,7 @@ We create a shell script to monitor `channel.backup` and make a copy to our back
   # -u causes the bash shell to treat unset variables as an error and exit immediately.
   set -eu
 
-  # The script waits for a change in ~/.lnd/data/chain/bitcoin/mainnet/channel.backup.
+  # The script waits for a change in /data/lnd/data/chain/bitcoin/mainnet/channel.backup.
   # When a change happens, it creates a backup of the file locally
   #   on a storage device and/or remotely in a GitHub repo
 
@@ -128,26 +128,26 @@ We create a shell script to monitor `channel.backup` and make a copy to our back
   LOCAL_BACKUP_ENABLED=false
   REMOTE_BACKUP_ENABLED=false
 
-  # Locations of source SCB file, formatted backup files and Git repo
-  SOURCEFILE="~/.lnd/data/chain/bitcoin/mainnet/channel.backup"
-  LOCAL_BACKUP_FILE="/mnt/static-channel-backup-external/channel-$(date +"%Y%m%d-%H%M%S").backup"
-  GITREPO="~/.lnd/remote-lnd-backup"
-  REMOTE_BACKUP_FILE="~/.lnd/remote-lnd-backup/channel-$(date +"%Y%m%d-%H%M%S").backup"
-
+  # Locations of source SCB file and the backup target directories (local and remote)
+  SCB_SOURCE_FILE="/data/lnd/data/chain/bitcoin/mainnet/channel.backup"
+  LOCAL_BACKUP_DIR="/mnt/static-channel-backup-external"
+  REMOTE_BACKUP_DIR="/data/lnd/remote-lnd-backup"
 
   # Local backup function
   run_local_backup_on_change () {
     echo "Copying backup file to local storage device..."
-    cp "$SOURCEFILE" "$LOCAL_BACKUP_FILE"
+    echo "$1"
+    cp "$SCB_SOURCE_FILE" "$1"
     echo "Success! The file is now locally backed up!"
   }
 
   # Remote backup function
   run_remote_backup_on_change () {
     echo "Entering Git repository..."
-    cd $GITREPO || exit
+    cd $REMOTE_BACKUP_DIR || exit
     echo "Making a timestamped copy of channel.backup..."
-    cp "$SOURCEFILE" "$REMOTE_BACKUP_FILE"
+    echo "$1"
+    cp "$SCB_SOURCE_FILE" "$1"
     echo "Committing changes and adding a message"
     git add .
     git commit -m "Static Channel Backup $(date +"%Y%m%d-%H%M%S")"
@@ -161,17 +161,20 @@ We create a shell script to monitor `channel.backup` and make a copy to our back
   run () {
     while true; do
 
-        inotifywait $SOURCEFILE
+        inotifywait $SCB_SOURCE_FILE
         echo "channel.backup has been changed!"
+
+        LOCAL_BACKUP_FILE="$LOCAL_BACKUP_DIR/channel-$(date +"%Y%m%d-%H%M%S").backup"
+        REMOTE_BACKUP_FILE="$REMOTE_BACKUP_DIR/channel-$(date +"%Y%m%d-%H%M%S").backup"
 
         if [ "$LOCAL_BACKUP_ENABLED" == true ]; then
           echo "Local backup is enabled"
-          run_local_backup_on_change
+          run_local_backup_on_change "$LOCAL_BACKUP_FILE"
         fi
 
         if [ "$REMOTE_BACKUP_ENABLED" == true ]; then
           echo "Remote backup is enabled"
-          run_remote_backup_on_change
+          run_remote_backup_on_change "$REMOTE_BACKUP_FILE"
         fi
 
     done


VajraOfIndra (commented):
tACK, I tested it by using `$HOME` and simulated a SCB change, all good.

I'll let someone more expert in shell script review and approve the change, but as far as I can tell this solves a bug in the guide! 

Thanks @the-jade-rabbitt !

the-jade-rabbitt:
Thanks @VajraOfIndra 
I came across another bug in this script when I enabled remote and local backups.
The file names were not being updated with current timestamps because they were only being set once at script startup.  This caused the script to bomb out during the remote backup and resulted in the system service restarting.  The effect of this was only half the backups were making it to the remote backup! 
I've pushed an amend to the existing commit this time.
This will need further review from somebody since I've changed the PR since you last tested it above.
I tested it by touching the watched file while watching the journal log (as recommended by the guide) and also checked the changes to the remote and local backups and I think I can see it working as intended with my fixes in there.

VajraOfIndra:
> Thanks @VajraOfIndra I came across another bug in this script when I enabled remote and local backups. The file names were not being updated with current timestamps because they were only being set once at script startup. This caused the script to bomb out during the remote backup and resulted in the system service restarting. The effect of this was only half the backups were making it to the remote backup! I've pushed an amend to the existing commit this time. This will need further review from somebody since I've changed the PR since you last tested it above. I tested it by touching the watched file while watching the journal log (as recommended by the guide) and also checked the changes to the remote and local backups and I think I can see it working as intended with my fixes in there.

@Stadicus @RooSoft: I don't feel I know enough about bash scripting, could you review these two changes? The SCB guide does not work as it is right now so it's be great to merge a fix asap.

RooSoft:
`$HOME` and `~` are pretty much synonyms...

IMO `~` is way easier to read. I personally never use `$HOME`.

RooSoft:
One thing I have to mention, `~` is a relative path...

In my experience, `systemd` and relative paths don't bode well together... maybe that's why @the-jade-rabbitt proposed $HOME, as that variable can be set in environment variables of the systemd config file, while I doubt it's possible with `~`.

the-jade-rabbitt:
Hi @RooSoft, the script produces an error if you use a ~ inside quotes because it is then treated as a literal.
Here is a stackoverflow that mensions this issue.
https://unix.stackexchange.com/questions/151850/why-doesnt-the-tilde-expand-inside-double-quotes
It does work if you remove the quotes but I don't think it can be avoided where there are spaces within the string assignments.

Using an absolute path would work - but I think it causes a problem with permissions unless you go through the lnd user path with the simlinks.

RooSoft (changes requested):
I like the way that variables are now defined closer to where they're useful.

To further improve to this regard, functions such as `run_local_backup_on_change` and `run_remote_backup_on_change` should accept variables... as code is laid out actually, these functions are defined before the variables they use. It doesn't matter much as they'll probably be called once the variables exist, but function arguments would simplify things up and avoid errors on future changes.

Unsure about this, but I think being more specific by setting up the $HOME environment variable in the systemd config file might help on execution.

RooSoft:
> Using an absolute path would work - but I think it causes a problem with permissions unless you go through the lnd user path with the simlinks.

Regarding permission rights, absolute and relative paths are the same... the difference is that in certain circumstances, relatives paths don't translate to absolute. As far as I know, systemd requires absolute paths, even when the user is specified...

Stadicus:
Updating the timestamps shortly before these are needed is certainly a good thing. :) Initiating them at the beginning can't work of course. Thanks for bringing this up.

But I also think all "configuration" variables should be defined at the top of a script. For example, I'd avoid hardcoding the `/mnt/static-channel-backup-external...` path in the middle of the execution code.

That brings me to another question: why would we want to go through the user's home directory in the first place? We can keep it simple and simply specify three absolute directories in the beginning, easy to change to any system configuration:

* `SCB_SOURCE_FILE="/data/lnd/data/chain/bitcoin/mainnet/channel.backup"`
* `LOCAL_BACKUP_DIR="/mnt/static-channel-backup-external/"`
* `REMOTE_BACKUP_DIR="/data/lnd/remote-lnd-backup/"`

Then, during runtime, the file variables can be created from these, without hardcoding any specific directories. I feel explicit is usually better and easier to understand.

the-jade-rabbitt:
Yes - much better. PR updated.  Thanks!

I've really enjoyed going through the guide.  Got everything up and running on my pi now.
I've also added a service to run testnet too.
Next steps for me will be Learning Bitcoin from the Command Line with Christopher Allen :)

the-jade-rabbitt:
oops, removing the $HOME now ...

RooSoft:
@Stadicus I agree about not relying on users homes for that kind of job...

static-channel-backup.md
```diff
@@ -129,16 +129,16 @@ We create a shell script to monitor `channel.backup` and make a copy to our back
...
   # Locations of source SCB file, formatted backup files and Git repo
-  SOURCEFILE="~/.lnd/data/chain/bitcoin/mainnet/channel.backup"
-  LOCAL_BACKUP_FILE="/mnt/static-channel-backup-external/channel-$(date +"%Y%m%d-%H%M%S").backup"
-  GITREPO="~/.lnd/remote-lnd-backup"
-  REMOTE_BACKUP_FILE="~/.lnd/remote-lnd-backup/channel-$(date +"%Y%m%d-%H%M%S").backup"
-
+  SCB_SOURCE_FILE="/data/lnd/data/chain/bitcoin/mainnet/channel.backup"
+  LOCAL_BACKUP_DIR="/mnt/static-channel-backup-external"
+  REMOTE_BACKUP_DIR="/data/lnd/remote-lnd-backup"
+  GITREPO="/data/lnd/remote-lnd-backup"
```
Stadicus, line 135: I think we could get rid of `GITREPO`, as it's redundant:

* Delete this line here
* Replace it with `REMOTE_BACKUP_DIR` for the `cd` command in line 148

static-channel-backup.md
```diff
@@ -161,17 +162,20 @@ We create a shell script to monitor `channel.backup` and make a copy to our back
...
+        inotifywait $SCB_SOURCE_FILE
         echo "channel.backup has been changed!"
 
+        LOCAL_BACKUP_FILE="$LOCAL_BACKUP_DIR/channel-$(date +"%Y%m%d-%H%M%S").backup"
+        REMOTE_BACKUP_FILE="$REMOTE_BACKUP_DIR/channel-$(date +"%Y%m%d-%H%M%S").backup"
+
         if [ "$LOCAL_BACKUP_ENABLED" == true ]; then
           echo "Local backup is enabled"
-          run_local_backup_on_change
+          run_local_backup_on_change $LOCAL_BACKUP_FILE
```
Stadicus, line 173: Being pedantic, the variable should be in double quotes:
```
run_local_backup_on_change "$LOCAL_BACKUP_FILE"
```
https://github.com/koalaman/shellcheck/wiki/SC2086

static-channel-backup.md
```diff
@@ -129,16 +129,16 @@ We create a shell script to monitor `channel.backup` and make a copy to our back
   REMOTE_BACKUP_ENABLED=false
 
   # Locations of source SCB file, formatted backup files and Git repo
```
Stadicus, line 131: Could be slightly rephrased to reflect the new variables:

```
# Locations of source SCB file and the backup target directories (local and remote)
```

Stadicus:
> Next steps for me will be Learning Bitcoin from the Command Line with Christopher Allen :)

That's an awesome resource! Never quite made it through it, though... :smile: 

> Would it have been better if I pushed a sequence of commits to this PR as opposed to ammending each time?

Doesn't really matter, no worries. I configured the repository so that it automatically squashes all commits on merge anyway. Thanks for contributing in the first place, much appreciated! :pray:

the-jade-rabbitt:
> I configured the repository so that it automatically squashes all commits on merge anyway 

👍

If the automatic squash loses the 1st commit message I can update it if you like.
Or does it keep them all?

static-channel-backup.md
the-jade-rabbitt: Today I leaned something :)

Stadicus (approved):
tACK

Stadicus:
> If the automatic squash loses the 1st commit message I can update it if you like.
> Or does it keep them all?

It just keeps all the commit titles as a list, without the individual commit messages. If you want, you can phrase a commit message here so that I can copy/paste it when I'll merge after @RooSoft's review

the-jade-rabbitt:
The titles will be fine :-)

RooSoft:
Seems to me there's a conflict... maybe I could review the PR once it's resolved...

the-jade-rabbitt:
> Seems to me there's a conflict... maybe I could review the PR once it's resolved...

Thanks @RooSoft, I've rebased it now.

RooSoft (approved):
That's ok to me, but I'd still make arguments instead of using environment variables in functions.

Might propose a PR later to change this if this idea proves to be popular. Would make things more obvious in my opinion.

Stadicus:
@the-jade-rabbitt, your change is now live. Thanks for contributing! :pray:

PR #946, opened 2022-03-02
charge-lnd: fix internal links

Stadicus:
#### What

This change fixes two internal links in the "charge-lnd" bonus guide.

### Why

The bonus guide used full URLs including the raspibolt.org domain, which we should avoid. I did not catch these broken links when testing the restructured guide in my own fork, as the links pointed to the not-yet-restructured public guide. 

#### How

This change fixes these two links by using relative paths, pointing to the local markdown files.

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

#### Test & maintenance

Check the links of the "charge-lnd" bonus guide.

### DIFF
--- a/guide/bonus/lightning/charge-lnd.md
+++ b/guide/bonus/lightning/charge-lnd.md
@@ -240,7 +240,7 @@ Then exit the charge-lnd user.
   $ exit
   ```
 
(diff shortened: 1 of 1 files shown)


VajraOfIndra (approved):
LGTM, thanks!
utACK

PR #948, opened 2022-03-03
Zeus guide: fixes mobile commands font + adds an APK intall section

VajraOfIndra:
#### What & Why

This PR proposed to:
* adds an APK install section for Zeus for Android users that cannot (e.g. can't find Zeus in Google Play) or do not want (e.g. for privacy goals) use Google Play
* fixes the font of a couple of commands to tap on the app (capitalized vs non capitalized) to match what the user sees on their phone
* switches the term "click" for "tap" in the section using the phone

#### How

* updated mobile-app.md

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

#### Test & maintenance

If desired, can download the APK and install it following the instructions in the guide to check that they are valid

andoid

### DIFF
--- a/guide/lightning/mobile-app.md
+++ b/guide/lightning/mobile-app.md
@@ -87,20 +87,21 @@ lnconnect generates a URI and displays it as a QR code that Zeus can read.
 ### Install the Zeus app
 
 Download the Zeus app for your mobile phone.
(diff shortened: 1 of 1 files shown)


Stadicus (approved):
utACK

PR #950, opened 2022-03-04
Add a "Resources" section to the Bonus section with 3 guides (Bitcoin Only, Bitcoin Minds, white paper)

VajraOfIndra:
#### What

This is a proposal to add a new "Resources" section to the Bonus section. This section could contain bonus guides related to resources compilations, learning etc.

The PR also already includes three guides in the section:
* [Bitcoin Only](https://bitcoin-only.com/): a self-hosted website with a compilation of Bitcoin and Lightning resource links
* [Bitcoin Minds](https://bitcoinminds.org/): similar to above
* Download the white paper: This guide was previously in the "Bitcoin" section of the Bonus section but would probably better fit in this section?

### Why

IMO, the addition of such a section and the self-hosted resources compilations has two benefits (in comparison to browsing the public websites of these two projects): _promotion of learning and more privacy_
* It shed lights/bring attention to the various articles, books and other resources listed in the websites and promote their discovery
* It allows for a private search of a relevant resource. Even though the reading/watching/listening of one of the listed resource is done on clearnet and is not private, the discovery process is private and not done in a search engine or on the public websites of Bitcoin Only or Bitcoin Minds

#### How

* Created a new directory: /guide/bonus/resources
* Created two guides for Bitcoin Only and Bitcoin Minds
* Moved the "Download the white paper" guide to this new directory and removed it from the Bitcooin bonus guide index.
* Created a "Resources" index

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

* If possible, the two new guides on a node; but a visual inspection/utACK might be good enough for these small bonus guides
* Verify that all the links between the guides and external are ok

#### WDYAT

* Is "Resources" a meaningful heading here? there might be a better category. "Others" or "Miscellaneous" could work well as well but I find the last one too academic and both of them too generic to the point of being meaningless... But happy to change "Resources" for them or another suggestion! Edit: addition ideas: "Learn" or "Education"  
* Bitcoin Minds has a page promoting their services as expert... It might be necessary to add a warning in our guide that we do not vouch for them and solely publish the guide for the resources compilation (or something like that)

Always be learning


VajraOfIndra:
Closing this proposal for the time being due to:
* possible unresolved issues with nginx/systemd services timing issues
* no show of interest

Might reopen in the future if needed

PR #956, opened 2022-03-07
Fix for installing to correct path

flexo82:
Fixed issue where pip3 install fails to install to ~/.local/bin


Stadicus (approved):
Thanks for this fix!
utACK

PR #961, opened 2022-03-11
BoS: use built-in flag `--use-proxy` instead of torify

blckbx:
#### What

Connect BoS to Telegram via proxy (Tor/VPN/other proxy) using built-in flag `--use-proxy <file>` instead of `torify`.

### Why

Proxy BoS to Telegram via Tor or VPN or custom proxy.

#### How

Use built-in flag `--use-proxy` which is available since BoS version 11.50.1.

#### Scope

- [x] simple section changes in `balance-of-satoshis.md`

#### Test & Maintenance

Please review, implement and test. Unfortunately I'm not using a rpi setup, so this was translated. It should be set and forget once installed.


VajraOfIndra (approved):
tACK (switching from torify to proxy)

Thanks a lot @blckbx , this is much neater than using `torify`.

PR #964, opened 2022-03-13
patch: bos TG first setup via proxy

blckbx:
#### What

patch: bos TG first setup via proxy
fix: some typos


VajraOfIndra (approved):
LGTM, good catch!

PR #967, opened 2022-03-15
Split "Electrum server" guide into an "Electrum server" (Electrs) and "Desktop wallet" (Sparrow) guides

### LINKED ISSUE #896: [FEATURE REQ] Split "Electrum Server" guide into two guides: "Electrum Server" & "Desktop wallet" + focus on one desktop app only (e.g. Sparrow)
### What & why

Presently, the Electrum server guide contains instructions on:
1. How to install and setup Electrs
2. How to connect the Electrum wallet and BitBoxApp to Electrs

However, the Electrums server is used for other apps than Electrum and BitBoxApp: e.g. the BTC RPC Explorer and possibly other wallet apps.
I'm noticing that new users often have some issues connecting their desktop wallet (e.g. Electrum) to their Electrum server. This could be partly due to the concise nature of the existing guidelines.

Therefore, to make things structurally clearer and to have a bit more space to explain how to connect a wallet to the server, I think that separating the two sections could make a lot of sense, i.e.:
1. An "Electrum server" guide that focuses solely on the installation and configuration of Electrs (and pointing to alternatives in the bonus section, e.g. ElectrumX etc)
2. Followed by a "Desktop wallet" guide that clearly explains how to install and connect a desktop wallet to Electrs

So the new layout for the "Bitcoin" section of the guide could be:
1. Bitcoin Core*
2. Electrum server
3. Desktop wallet
4. Blockchain Explorer

\* _Note_: as the other sub-headings are all generic names rather than program names.. we should maybe use "Bitcoin node implementation" or something more generic like this rather than "Bitcoin Core".. wdyt?

For the "Desktop wallet" guide, I think we should focus on explaining how to install and connect only _one_ popular desktop wallet. All other apps could be as bonus guides linked to in this section. IMHO, the two main candidates would be either:
1. Electrum
2. Sparrow

My preference would go to Sparrow as its focus on showing how things are under the hood really resonates with the whole "RaspiBolt" philosophy (+ its native coinjoining capabilities that could be highlighted)

### How

- Restrict the existing "Electrum Server" guide to installation and configuration of Electrs only (remove the Electrum and BitBoxApp sections)
- Create a new "Desktop wallet" guide that explains how to install only one popular app (e.g. Sparrow) and how to connect it to Electrs. Links to the bonus section for guides on how to connect other apps (Electrum, BitBoxApp, Specter, Bisq etc).

Wdyat? :)

VajraOfIndra:
#### What

This PR is a proposal to split the existing Electrs guide into two: 
* an "Electrum server" guide explaining how to install and configure Electrs
* a "Desktop wallet" guide explaining how to install and configure Sparrow

### Why

tl;dr: 
1. Clarity: The present-day guide mixes two different pieces of software, the Electrum server and desktop wallets. It can lead to some amount of confusion for new users and would benefit from being split into two separate guide to make it easier for the new users to understand what each program does, how they relate to each other and to give us more space to properly explain how to install and configure both pieces of software.
2. Sparrow is proposed over other desktop wallets (Electrum, BitBoxApp, Specter..) because:
  * it aims at showing what's going on under-the-hood at all stages, which is very aligned with the RaspiBolt spirit of education and understanding how things work
  * it iis feature-rich with a big focus on privacy, transparency and security: HW, mutli-sig, batch tx, PSBT, paynyms, coinjoin, explorer and many more
  * Although young, it is very popular and has an excellent reputation within the community
3. Will lead to less users encountering issues (many reported issues presently when using Electrum)

For more rationales and explanation, see discussion thread #896 

#### How

* Created a `desktop-wallet.md file`
* Removed the Electrum and BitBoxApp sections from `electrum-server.md` and made them as separate bonus guides `electrum.md` and `bitboxapp.md`
* Amended `electrum-server.md` to reflect the new focus on Sparrow
* Amended a few files to correct for new links and indexes

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

Fixes #896

#### Test & maintenance

Follow the Desktop wallet guide to test all the commands and links and see if the wallet connects to your Electrs!

#### To do

1. The "(Optional) Connect to Bitcoin Core" section did work on my node but sometimes led me to have a buggy and very slow node. Not entirely sure why but it seems to be linked to the modification of the bitcoin.conf file. Unless we pinpint what the issue is, we might need to remove this section entirely.. but quite a few users do not use an Electrum server so if it can be fixed and left that could be nice.
2. The bonus guides for Electrum, BitBoxApp and Specter Desktop would benefit from a complete revision and re-testing.  For now, I simply pasted the same text that was in the Electrs guide.

#### Animated GIF (optional)
Sparrows


RooSoft (changes requested):
Nothing much to change, really, great job!

Only thing is the Desktop Wallet section that is a bit confusing, as it mixes many use cases. We've talked about this with @VajraOfIndra so I will refrain from going in depth about it now.

VajraOfIndra:
> Only thing is the Desktop Wallet section that is a bit confusing, as it mixes many use cases. We've talked about this with @VajraOfIndra so I will refrain from going in depth about it now.

Very good point. So I removed the "Connection to Bitcoin Core" section entirely. This makes the guide significantly shorter, simpler and less confusing. If users would like to know how to do this, we could always write a bonus guide for that.

I also wrote an Introduction to try to explain how Sparrow uses Tor for various purposes. I might not always use the right termnilogy as I'm still not groking how it all works.. so if someone could update this intro to make it technicall sounder and also clearer, that would be awesome.

Do we have to enable the Tor Proxy when using a Tor onion address to connect to the server? or does Sparrow then default to Tor to connect to the external services? I'm still a bit confused about when does Sparrow uses its own in-built Tor vs when running Tor as an external service on the computer is required..!

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+{: .no_toc .text-delta }
+
+1. TOC
+{:toc}
+
+---
+
+## Introduction
+
+Sparrow can connect to either your own Electrum server, or directly to your Bitcoin Core, or to a public Electrum server. We will connect Sparrow to our own Electrum server as it is the most private and secure option. For more information about the privacy and security trade-offs of the various server connection options, read the following [article](https://www.sparrowwallet.com/docs/best-practices.html){:target="_blank"} by Craig Raw, the author of Sparrow.
```
RooSoft, line 28: Now that the Bitcoin Core use case isn't documented anymore, should we remove the reference to it here?

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+### Launch Sparrow
+
+* Launch Sparrow
+* Read carefully the introductory messages and click on "Next" several times
+* When you reach the "Connecting to a Private Electrum Server" message, click on "Configure Server"
+
+### Connect to Electrs
+
+* Click on the "Private Electrum" tab
+* On the "URL" line, paste `raspibolt.local` or your node IP (e.g. `192.168.0.20`) in the first box and `50002` in the second box
```
RooSoft, line 76: Is it common to refer to this specific `192.168.0.20` IP elsewhere in the docs?

Should `e.g.` be replaced by `ex:`? I think `e.g.` means that this is the actual value and thus not an example.

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+* Enable SSL by clicking on the slider (it should become blue)
```
RooSoft, line 77: That's my own opinion, but I'd remove some parenthesis in this docs, and here is a good example.

`Enable SSL by clicking on the slider. Then, make sure that it turns blue.`

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+* Enable the Tor proxy by clicking on the "Use proxy" slider
+* On the "Proxy URL" line, paste `127.0.0.1` in the first box and `9150` in the second box (or `9050` if you run Tor as a background service)
```
RooSoft, line 79: On the "Proxy URL" line, paste `127.0.0.1` in the first box and either `9150` or `9050` in the second box depending if you run Tor as a background service or not

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+
+  ```
+  $ sudo cat /var/lib/tor/hidden_service_electrs/hostname`
+  > abcd...1234.onion
+  ```
+
+### Sparrow configuration
+
+* Open Sparrow
+* Navigate to the server configuration page (`Ctrl`+`p` then click on "Server")
```
RooSoft, line 145: Navigate to the server configuration page by hitting `Ctrl`+`p` then click on "Server"

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+
+![Sparrow stages](../../images/sparrow-stages.png)
+
+We will set up Sparrow to connect to Electrs within your local network. There is also an optional section at the end that explains how to connect Sparrow to Electrs using Tor for when you're on the move. 
+
+Sparrow also connects to a couple of external services to get the Bitcoin price and communicate with the Whirlpool server during coinjoin operations. By default, it uses clearnet which leaks your computer IP address to these services and degrade your privacy. However, Sparrow can also connect to them using a Tor proxy. Whether you are using Tor or not to connect to your node, in both cases we will set up the Tor proxy for Sparrow to privately connect to these external services.
```
RooSoft, line 34: Arhe connections to Whirlpool and Bitcoin price services through tor described in this doc? I can't find any other reference to them from this point.

If it's automatically done once the proxy is set, maybe this fact should be more clearly explained here. "Create a proxy, set it up in the connection and those will be taken care of, nothing more to do..."

RooSoft (changes requested):
Some minor comments...

I'd remove most of the parenthesis and turn them into more complete phrases... specifically pointed out some examples, but refrained from commenting them all.

guide/bitcoin/desktop-wallet.md
VajraOfIndra: Well, I thought it was worth poitning out the different options.. but yes, can remove the first sentence for example. I think pointing to the linked article is a useful resource, but we could change the wording of the rest of the paragraph.

guide/bitcoin/desktop-wallet.md
VajraOfIndra: It's used in the guide yes, for example [here](https://raspibolt.org/guide/raspberry-pi/remote-access.html#find-your-raspberry-pi)
> You should now be able to reach your Pi, either with the hostname raspibolt.local or an IP address like 192.168.0.20.

Afaik, e.g. just means "for example"](https://www.merriam-webster.com/dictionary/e.g.) and is used in other parts of the guide.

Happy to use another notation though, as long as we apply it consistently throughtout the guide.. What would you have in mind?

guide/bitcoin/desktop-wallet.md
VajraOfIndra: Good point! I removed the section in parenthesis altogother as it's not that useful anyway.

guide/bitcoin/desktop-wallet.md
VajraOfIndra: Fixed

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+* If you have used Sparrow before on this computer, try to connect again after deleting the existing certificates that are stored within the `certs` folder:
+  * On Windows: `C:\Users\remi\AppData\Roaming\Sparrow\certs`
+  * On macOS: `~/.sparrow/cert`
+  * On Linux: `~/.sparrow/cert`
+
+Now, let's go back to the wallet and check that the server and proxy are working properly.
+
+* Close the server configuration window
+* Check the two icons in the bottom right corner of the wallet window ![status icons](../../images/sparrow-server-proxy-icons.png)
+  * the first icon should be grey (not red) and a mouse hover should display "External Tor proxy enabled"
```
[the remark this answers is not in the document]
VajraOfIndra: Absolutely, good catch. Fixed!

guide/bitcoin/desktop-wallet.md
VajraOfIndra: Fixed!

guide/bitcoin/desktop-wallet.md
VajraOfIndra: I'm not sure understand your point. This introductory paragraph aims at explaining the need to set up the Tor Proxy in Sparrow's network config to ensure that the connection to the external service are done privately via Tor.. It justifies/explains why later in the guide we require the user to install and launch the Tor Browser and set up the Tor Proxy in the configs.

Could you paste an amended paragraph with the modification you'd like to see?

VajraOfIndra:
> Some minor comments...
> 
> I'd remove most of the parenthesis and turn them into more complete phrases... specifically pointed out some examples, but refrained from commenting them all.

Thanks for the comments. I removed most of the parenthses from the guide.
The only thing I was unclear about is regarding the section about external services in the Intro.

I also added a little paragraph in  the Bitcoin index.md to explain what the user will install in the Bitcoin section. I used the same model as was done in the Lightning index file.

VajraOfIndra:
I'm unsure about one thing though.. I asked on the Sparrow TG group but didn't get any replies.

When using a Tor onion address  to connec to the server, Sparrow uses its own in-built Tor engine (not sure what the right word is there.. Tor instance? Tor service? anyways!). So.. in that case, is there still a need to toogle on the "Tor Proxy" and set it up to local host port 9050|9150..?! Wouldn't the connection to the external services also use the internal built-in Tor and NOT need the external Tor running on the computer?

guide/bitcoin/desktop-wallet.md
RooSoft: Oh, my bad on this one... seems like it's fine as it is.

guide/bitcoin/desktop-wallet.md
RooSoft: I mean, if we setup tor, will Sparrow automatically use tor-based whirlpool and bitcoin price services?

RooSoft (approved):
Looks good to me!

guide/bitcoin/desktop-wallet.md
VajraOfIndra: > I mean, if we setup tor, will Sparrow automatically use tor-based whirlpool and bitcoin price services?

That's exactly what I'm wondering as well! :) See post below. I don't know enough about Tor to guess if it's the case or not.. I  tried to ask on their TG group but got no reply so far.

VajraOfIndra:
> Looks good to me!

Thanks for the review and improvement suggestions.
Let's wait for @Stadicus review and opinion on this proposal now.

And we also need to sort out this Tor proxy thing, to make sure we use it right.

asvdf:
I agree.
On a general note, I find the way the guide is currently laid out is "confusing", with the two big sections "bitcoin" and "lightning" instead of a linear progression, which IMO would be bitcoind > lnd > electrum > explorer > wallet without subsections at all... but I'm digressing, and not the place for this.

In any case, I agree with the OP, as having a precompiled path of wallet software right in the Electrum guide, will give the impression there are no other wallets out there.

Instead, you just present the electrum guide, and THEN you add a reference to separate guides, "...for example one wallet you could use is Sparrow (see guide), but then there's electrum (see guide) or bitbox (see guide) or whatever else you prefer (set the electrum server ip:port in settings)

VajraOfIndra:
> ... but I'm digressing, and not the place for this.

Maybe open an Issue to discuss this.

> Instead, you just present the electrum guide, and THEN you add a reference to separate guides, "...for example one wallet you could use is Sparrow (see guide), but then there's electrum (see guide) or bitbox (see guide) or whatever else you prefer (set the electrum server ip:port in settings)

The present proposal is not doing that. It actually follow the same method as for the rest of the guide by giving guidelines for _one_ choosen program, in this case Sparrow. 
Sparrow here is a guide of its own under the title "Desktop wallet" that follows the "Electrum server" guide. Sure, we could suggest alternatives and point to bonus guides.. but then we should probably do the same with the other sections of the guide for consistency (which is not done right now).

RooSoft (approved):
Still good

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+title: Desktop wallet
+nav_order: 30
+parent: Bitcoin
+---
+<!-- markdownlint-disable MD014 MD022 MD025 MD033 MD040 -->
+
+# Desktop wallet
+{: .no_toc }
+
+We install the [Sparrow](https://github.com/romanz/electrs/){:target="_blank"} desktop wallet on a computer and connect it to your Electrum server on your node for private bitcoin on-chain transactions.
```
kristapsk, line 12:
```suggestion
We install the [Sparrow](https://github.com/sparrowwallet/sparrow){:target="_blank"} desktop wallet on a computer and connect it to your Electrum server on your node for private bitcoin on-chain transactions.  
```

VajraOfIndra: Thanks for spotting this typo, fixed!

guide/bitcoin/desktop-wallet.md
Stadicus: I'm also no expert in this. If setting the Tor proxy solves all issues, I agree that this paragraph could be removed. 

Personally, I'd even make the "Tor" section optional, with a (very) short explanation. I'm not sure if hiding your IP address when querying for some exchange rates is a MUST. But I don't have a strong opinion on this.

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+
+We will set up Sparrow to connect to Electrs within your local network. There is also an optional section at the end that explains how to connect Sparrow to Electrs using Tor for when you're on the move. 
+
+Sparrow also connects to a couple of external services to get the Bitcoin price and communicate with the Whirlpool server during coinjoin operations. By default, it uses clearnet which leaks your computer IP address to these services and degrade your privacy. However, Sparrow can also connect to them using a Tor proxy. Whether you are using Tor or not to connect to your node, in both cases we will set up the Tor proxy for Sparrow to privately connect to these external services.
+
+## Preparations
+
+### Tor
+
+For Sparrow to connect to the external services via Tor, Tor has to be running on your computer. You can run the Tor Browser and use port 9150 on localhost or else run Tor as a background service and use port 9050.
```
Stadicus, line 40: As mentioned above, maybe this section could be optional, with a very quick explanation here:

> Sparrow Wallet connects to external services to fetch exchange rates or when facilitating CoinJoins. This can reveal your IP address. If you care for maximum privacy, we recommend using Tor to obscure this information.

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+
+Here, we'll use the Tor Browser and port 9150 as it is easier to set up on Windows and macOS.
+
+* Install the Tor Browser on your computer by following [this guide](https://tb-manual.torproject.org/installation/){:target="_blank"}.
+* Launch the Tor Browser
+* Click on "Always connect automatically" and "Connect"
+* Keep the Tor Browser running while using Sparrow
+
+*Note:* For Linux Debian-based OS users, you might prefer to just have Tor running in the background and not have to open the Tor Browser each time you're using Sparrow. If so, instead of intalling the browser, simply install Tor with `sudo apt install tor` and check that it is running with `systemctl status tor`.
```
Stadicus, line 49: When using the Tor daemon, we should note that the proxy port is `9050` instead of `9150`

guide/bitcoin/desktop-wallet.md
Stadicus: If we want to split hairs, "e.g." requires a comma immediately after ;-)

> your node IP (e.g., `192.168.0.20`)

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+* Read carefully the introductory messages and click on "Next" several times
+* When you reach the "Connecting to a Private Electrum Server" message, click on "Configure Server"
+
+### Connect to Electrs
+
+* Click on the "Private Electrum" tab
+* On the "URL" line, paste `raspibolt.local` or your node IP (e.g. `192.168.0.20`) in the first box and `50002` in the second box
+* Enable SSL by clicking on the slider
+* Enable the Tor proxy by clicking on the "Use proxy" slider
+* On the "Proxy URL" line, paste `127.0.0.1` in the first box and either `9150` or `9050` in the second box depending if you run the Tor Browser or Tor as a background service.
```
Stadicus, line 79: It confuses me a bit that we connect to Electrs using a local connection, and enable Tor anyway.

I guess this comes from using the proxy for external services, but it's a bit out-of-scope for the section "Connect to Electrs". Especially in case you'd make the Tor section above optional. Maybe it could be split out of this section?

But maybe this is the simplest way to run through the configuration and we can all leave it as is. No strong opinion, just some thoughts, I'm happy to leave this up to you.

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+  * On Linux: `~/.sparrow/cert`
+
+Now, let's go back to the wallet and check that the server and proxy are working properly.
+
+* Close the server configuration window
+* Check the two icons in the bottom right corner of the wallet window ![status icons](../../images/sparrow-server-proxy-icons.png)
+  * The first icon should be grey, not red; and a mouse hover should display "External Tor proxy enabled"
+  * The second icon should be a blue slider button and a mouse over should display "Connected to [...] at height [...]"
+
+You're set! Sparrow is now configured to connect to your Electrum server from within your LAN.
```
Stadicus, line 103: I'd replace "LAN" with "local network" to simplify

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+* Coinjoin your coins for improved privacy
+* Analyse your tansactions with the built-in blockchain explorer
+
+For more information, tutorials and support, visit the [Sparrow documentation webpage](https://sparrowwallet.com/docs/){:target="_blank"} and their [Telegram group](https://t.me/sparrowwallet){:target="_blank"}.
+
+---
+
+## (Optional) Remote connection over Tor
+
+If you want to use Sparrow outside your LAN, when you're on the go, you can use a connection over Tor.
```
Stadicus, line 128: s/LAN/home network/

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,171 @@
...
+{: .no_toc .text-delta }
+
+1. TOC
+{:toc}
+
+---
+
+## Introduction
+
+We will connect Sparrow to our own Electrum server as it is the most private and secure option. For more information about the privacy and security trade-offs of the various server connection options, read the following [article](https://www.sparrowwallet.com/docs/best-practices.html){:target="_blank"} by Craig Raw, the author of Sparrow.
```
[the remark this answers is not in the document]
Stadicus: Here, I would also explain that Sparrow Wallet is a great choice, but that alternatives exist. Not every user wants to switch from the software wallet they're used to:

> You can also use the following alternatives instead of Sparrow Wallet, according to your preferences and needs: 
> * [Electrum](../bonus/bitcoin/electrum.md): a well-established poweruser wallet
> * [BitBoxApp](../bonus/bitcoin/bitboxapp.md): wallet for users of BitBox hardware wallets
> * [Trezor Suite](https://blog.trezor.io/connecting-your-wallet-to-a-full-node-edf56693b545){:target="_blank"}: wallet for users of Trezor hardware wallets

I'm not quite sure if you can use Ledger Live directly with an Electrum server. If so, I think we should add it too.

Stadicus (commented):
Great update, I think that makes the guide even more modular and well structured. 

In general, I'd remove the exclusive focus of Sparrow Wallet a bit. In the end, the RaspiBolt is a full node project, and a desktop wallet not an integral part of the node itself. It's good to have one option in the main guide, but I'd still point prominently to alternatives that users might already know and use.

Also, I'd call it "Sparrow Wallet" consistently, as it's used in their own project.

I also found an old issue that could be quickly fixed with this pr (can't comment directly on it, as it's not touched by this pull request):

* on line 39, there's an unnecessary `#prepare-nginx-reverse-proxy` outside the link (probably an old copy/paste error)

VajraOfIndra (commented):
Thanks @Stadicus for the review! 

I implemented all your suggestions. PTAL

I also updated all the pictures to reflect the change with the Tor proxy and also to dislay our default raspibolt.local instead of my own IP

craigraw:
Seeing this PR for the first time - honoured to have Sparrow recommended in my favourite node "package"!

Some notes re Sparrow and Tor:
1. Sparrow can be configured to use an internal (bundled) or external Tor proxy. To use the internal proxy, an onion Electrum server URL must be specified, and the 'Use Proxy' toggle must be off. The external proxy is configured as you have described.
2. If a Tor proxy is configured in Sparrow, all external connections use Tor. This includes rate fetching, coinjoin etc - even transaction broadcasting is then done via external service (like blockstream.info) over Tor for additional privacy. Where-ever possible, the onion URLs of these services are used.
3. It is possible (and I believe desirable in common situations) to use a local IP for the Electrum server, and an external Tor proxy for all other network connections. There is little point in connecting over slow and unreliable Tor when the server is on your local network. I believe you have described this approach well, and the alternative of an onion address when travelling.
4. Although this may well be due to the nature of the Raspibolt setup, it is technically not necessary to use SSL when communicating over Tor, as Tor traffic is already encrypted.

A couple of very minor nits I picked up:
1. The certificate folder is always named 'certs' - not 'cert'
2. The shortcuts in Sparrow follow the OS defaults as far as possible, so the Preferences shortcut is 'Cmd+P' on OSX.

Otherwise this is great work! Looking forward to seeing it published.

guide/bitcoin/desktop-wallet.md
Stadicus: Nice. If you think it's too much, I'm happy if you keep it simple. As always, these are only suggestions.

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,207 @@
...
+  > Connected to electrs/0.9.5 on protocol version 1.4
+  > [...]
+  ```
+
+![Sparrow connect to Electrs](../../images/sparrow-electrum-no-proxy.png)
+
+If there is a connection error message, try the following troubleshooting:
+* Make sure that your computer is not connected to any "guest" WiFi network at home. A "guest" WiFi network prevents the computer to connect to any other devices on the local network, such as your node.
+* If you have used Sparrow Wallet before on this computer, try to connect again after deleting the existing certificates that are stored within the `certs` folder:
+  * On Windows: `C:\Users\remi\AppData\Roaming\Sparrow\certs`
```
Stadicus, line 86: maybe replace username 'remi' with something generic, like 'username'?

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,207 @@
...
+  * On macOS: `~/.sparrow/cert`
+  * On Linux: `~/.sparrow/cert`
```
Stadicus, line 88: As mentioned by Craig, for both macOS and Linux:

s/cert/certs/

guide/bitcoin/desktop-wallet.md
```diff
@@ -0,0 +1,207 @@
...
+
+Now, let's go back to the wallet and check that the server and proxy are working properly.
```
Stadicus, line 90: It's less about the server working properly, and without Tor we don't need the proxy at this point. Maybe:

>  Let's go back to the wallet and check that it is connected to our own Electrum server.

guide/bitcoin/desktop-wallet.md
Stadicus: IIUC, the current guide captures Sparrow's capabilities well:

* Use clearnet by default in your own network
* Use built-in Tor proxy when on the go
* For additional privacy, use external Tor proxy for external services (at home or on the go)

guide/bitcoin/desktop-wallet.md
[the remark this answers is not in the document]
Stadicus: Perfect thanks. I think we wanted to list alternatives at some point (while still keeping one straight and simple implementation path through the guide), and here I think is a good point to start.

Stadicus (approved):
Added a few nits, but otherwise I think this is ready to be merged.

Great work @VajraOfIndra, thanks a lot!

guide/bitcoin/desktop-wallet.md
VajraOfIndra: fixed!

guide/bitcoin/desktop-wallet.md
VajraOfIndra: fixed

guide/bitcoin/desktop-wallet.md
VajraOfIndra: Good point, I forgot to change that sentence

VajraOfIndra:
> Seeing this PR for the first time - honoured to have Sparrow recommended in my favourite node "package"!

❤️ 

> Some notes re Sparrow and Tor:
> 
>     1. Sparrow can be configured to use an internal (bundled) or external Tor proxy. To use the internal proxy, an onion Electrum server URL must be specified, and the 'Use Proxy' toggle must be off. The external proxy is configured as you have described.
> 
>     2. If a Tor proxy is configured in Sparrow, all external connections use Tor. This includes rate fetching, coinjoin etc - even transaction broadcasting is then done via external service (like blockstream.info) over Tor for additional privacy. Where-ever possible, the onion URLs of these services are used.

This is great and clarify the doubts I had on the Tor proxy when using a Tor onion address to connect to the server 👍 

I reused your exact phrasing for the explanatory introduction of the Tor proxy section!

>     3. It is possible (and I believe desirable in common situations) to use a local IP for the Electrum server, and an external Tor proxy for all other network connections. There is little point in connecting over slow and unreliable Tor when the server is on your local network. I believe you have described this approach well, and the alternative of an onion address when travelling.

For now, we have the Tor proxy as an optional section when using clearnet.. for the sake of simplification.

But I added a recommendation to set it up at the end of the main section that explains how to connect to the Electrum server:

> For maximal privacy, we highly recommend that you set up the Tor proxy when using Sparrow within your local network over clearnet. Check the [optional section](#optional-set-up-a-tor-proxy-for-external-services) at the end of this guide.

>     4. Although this may well be due to the nature of the Raspibolt setup, it is technically not necessary to use SSL when communicating over Tor, as Tor traffic is already encrypted.

I'm not too good with this networking stuff. I believe we do that right now because we set up the hidden Tor service to redirect to the nginx reverse proxy for the Electrum server which then redirects to the Electrs server itself... Would it make more sense to redirect it to the Electrum server directly (skipping the nginx reverse proxy)?

> A couple of very minor nits I picked up:
> 
>     1. The certificate folder is always named 'certs' - not 'cert'

Well spotted, fixed!

>     2. The shortcuts in Sparrow follow the OS defaults as far as possible, so the Preferences shortcut is 'Cmd+P' on OSX.

I added this to the instrcutions.

> Otherwise this is great work! Looking forward to seeing it published.

Thanks! let's go to master now and get community feedback :) 🎉

VajraOfIndra:
@craigraw , would you like to join our [RaspiBolt Telegram group](https://t.me/raspibolt)? it'd be awesome to have you lurk there :)

Also, do you mind if I publish a link to the Sparrow guide in your Sparrow Telegram group? some users might find it useful to install on a new node or something..

craigraw:
> Would it make more sense to redirect it to the Electrum server directly (skipping the nginx reverse proxy)?

I believe so yes. Using a Tor hidden service means no firewall ports need to be opened either. Nginx is not adding any security here, and just adds additional load to the request.

Also, I made a mistake earlier with the shortcut - I was making the point that Cmd generally replaces Ctrl on OSX, but actually in the specific case of opening an application's preferences on OSX, the system appropriate way is `Cmd+,` - not Cmd+P. Sparrow follows this convention. Apologies for the mixup.

> @craigraw , would you like to join our [RaspiBolt Telegram group](https://t.me/raspibolt)? it'd be awesome to have you lurk there :)

Done :)

> Also, do you mind if I publish a link to the Sparrow guide in your Sparrow Telegram group? some users might find it useful to install on a new node or something..

Go ahead!

VajraOfIndra:
> > Would it make more sense to redirect it to the Electrum server directly (skipping the nginx reverse proxy)?
> 
> I believe so yes. Using a Tor hidden service means no firewall ports need to be opened either. Nginx is not adding any security here, and just adds additional load to the request.

Thanks, we'll probably change that soon.

> Also, I made a mistake earlier with the shortcut - I was making the point that Cmd generally replaces Ctrl on OSX, but actually in the specific case of opening an application's preferences on OSX, the system appropriate way is `Cmd+,` - not Cmd+P. Sparrow follows this convention. Apologies for the mixup.

Just made a PR to fix it, it will be corrected soon.

> > @craigraw , would you like to join our [RaspiBolt Telegram group](https://t.me/raspibolt)? it'd be awesome to have you lurk there :)
> 
> Done :)
> 
> > Also, do you mind if I publish a link to the Sparrow guide in your Sparrow Telegram group? some users might find it useful to install on a new node or something..
> 
> Go ahead!

ta 👍

PR #968, opened 2022-03-15
Add a web services dashboard (Homer) to the core guide

VajraOfIndra:
#### What

This PR is a proposal to add a web service dashboard (Homer) to the core guide.

### Why

Presently, each bitcoin and lightning self-hosted web services have their own local URL that have to be saved in bookmarks. Homer allows to gather all these URLs in one beautiful UI with a negligable resource footprint.

While the core guide only contains two web services (BTC RPC Explorer and RTL), users could potentially have many more with the existing and soon-to-be bonus guides: e.g. ThunderHub, Mempool, Lightning Terminal, LNDg, Bitcoin Only, Bitcoin Minds etc.

The dashboard also allows to display a number of external links that could be very useful to RaspiBolt node operator such as the guide itself, all the community groups and their node profile in the various LN explorers.

More background information, rationales and discussions in Issue #849 

The proposal seems to be quite well supported by the community (but 19 votes only so small population):
image

The key question is: is such a feature valuable enough to be part of the core guide or does it stray too much from the goals of the core guide? IMO, I think it's a useful addition but if judged not critical enough, I'll transform this guide into a bonus guide instead.

#### How

* Install Homer after the "Privacy" section in the "Raspberry Pi" chapter, under a generic heading (e.g. "Web dashboard")
  * Explains that if a user does not plan to use self-hosted web services they can skip this section

* The initial config would contain all the useful external links and logos and web services categories but no links to any self-hosted services -> `web-dashboard.md`

* We would then have "Homer" sections within any subsequent guides that provide a web service:
  * _In the core guide that would be:_
    * the BTC RPC Explorer guide -> `blockchain-explorer.md#homer`
    * the RTL guide -> `web-app.md#homer`
  * _In the bonus section, that would be an optional section in:_
    * Mempool -> `mempool.md#optional-homer`
    * the white paper download -> `white-paper.md#optional-homer`
    * and later once more PRs are merged, Ligthning Terminal, ThunderHub and many others...!

#### Scope

- [x] significant change to core configuration
- [ ] independent bonus guide
- [ ] simple bug fix

Fixes #849 

#### Test & maintenance

Test: go through the new guide and sections and test Homer on your node

#### To do

1. Can we use raspibolt.local without any port to serve the dashboard as was intended in the OP of #849 
2. Is the nginx configuration suitable? does it work after reboot or same problem as with Mempool? seee #919 
3. I'm not entirely sure the upgrade section works as is

#### Animated GIF (optional)

Homer bitcoin

PR #978, opened 2022-04-05
Bonus guide: Homer

VajraOfIndra:
#### What

A web service dashboard

### Why/How

See discussions in closed PR #968 and in Issue # 849

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix


guide/bonus/raspberry-pi/homer.md
```diff
@@ -0,0 +1,610 @@
...
+  $ sudo chown -R www-data:www-data /var/www/homer
+  ```
+
+By default, the logos for Homer are located in the "/var/www/homer/assets/tools" folder. 
+However, if you want to re-install Homer for whatever reason, you will have to re-download and move all the logos. Instead, we'll create a symlink to our Homer data folder created above.
+
+* Create a symlink to the logos folder and change its ownership
+
+  ```sh
+  $ sudo ln -s /data/homer/tools /var/www/homer/assets
```
RooSoft, line 130: The tools folder is being used by Homer, such as in this URL: `https://raspibolt:4091/assets/tools/raspibolt3.png`

Means, I think the command above should be changed to:

```bash
sudo ln -s /data/homer /var/www/homer/assets
```

... so that the tools folder ends up existing under the assets folder.

Otherwise, links will need to be changed in the front end to remove the tools folder.

guide/bonus/raspberry-pi/homer.md
```diff
@@ -0,0 +1,610 @@
...
+  $ sudo chown -h www-data:www-data /var/www/homer/assets/tools
```
RooSoft, line 131: Here, if we do as above, the assets should be chown, with a recursive flag so that the tools folder gets affected as well.

```bash
sudo chown -R -h www-data:www-data /var/www/homer/assets
```

guide/bonus/raspberry-pi/homer.md
```diff
@@ -0,0 +1,610 @@
...
+    - name: "LN+"
+      url: "https://lightningnetwork.plus/nodes/<node_pubkey>"
+      target: "_blank"
+    - name: "1ML"
+      url: "https://1ml.com/node/<node_pubkey>"
+      target: "_blank"
+
+
+  ################################
+  # Services                     #
```
RooSoft, line 370: I guess this section could be a whole document in itself, as services could differ wildly between instances of Raspibolt.

For instance, it took me a while to understand that the White paper link was referring to a section I personally haven't applied on mine, hence the resulting 404.

RooSoft (changes requested):
Very impressive document, I got to say... applied it as is from start to finish with very little problems aside from the tools folder.

Kudos to @VajraOfIndra!

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: The idea was to include only web services that have been created if somone follows the core guide, i.e. the BTC RPC Explorer and RTL. The white paper is actually also the BTC RPC Explorer (just a specific page/URL from it) so it should work for anyone who has installed the Explorer.

Maybe, I could add a warning at the start that the guide assumes that the user followed the main guide ans had installed the Explorer and RTL.

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: > ... so that the tools folder ends up existing under the assets folder.

I believe that's what the existing symlink provides. It create a "tools" symlink in `/var/www/homer/assets` that redirects to the tools folder in `/data/homer` so that when Homer looks for `/assets/tools` it sees all  the images inside `/data/homer/tools`.

> Means, I think the command above should be changed to:

> `sudo ln -s /data/homer /var/www/homer/assets`

This creates the following symlink inside /assets: `homer -> /data/homer` Afaik, the tools folder would then be missing and Homer wouldn't be able to display any images...

But I might be misunderstanding your point.

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: We could also remove the white paper download as it's not a web service per se but a special link from the Explorer. 

I just thought it would add a bit of content to the dashboard and also increase the visibility of this little known feature of the blockchain.

Shall we remove it?

VajraOfIndra: Ok, so I added a warning in the introduction: 
image

VajraOfIndra:
Thanks @RooSoft for taking the time to review this PR!

I replied to your reviews, PTAL.

(I also added a few missing hyperlinks and updated the top illustration to match the guide in its present-day form)

guide/bonus/raspberry-pi/homer.md
RooSoft: Problem is that at that point, I don't think the `/var/www/homer/assets` folder exists... so the `/data/homer/tools` folder ends up being linked to by `/var/www/homer/assets` directly. When the website tries to reach files in the `assets/tools` folder, it issues some 404 because that folder doesn't exist.

Try it by deleting the `/var/www/homer/assets` folder and running the above command. The contents of the assets folders should be what tools is supposed to be and thus the tools folder will be absent.

RooSoft (changes requested):
The `assets` and `tools` folders documentation still needs some correction

RooSoft:
In my review, I think I forgot about that 404 the page hits in the background... no idea what it is.

`https://raspibolt:4091/assets/manifest.json`

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: > Problem is that at that point, I don't think the `/var/www/homer/assets` folder exists..

Hum... it should exist afaik. It is being created due to the following command (which is done a bit before, like 2 commands before):
`sudo rsync -av --delete /home/homer/homer/dist/ /var/www/homer/`

The rsync dump the content of the dist folder into /var/www/homer, and it includes the tools folder...

--- 

But let me delete Homer and follow the guide again, I'll see if I can reproduce your issue.
I'll get back to you once I tested this again :+1:

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: Hum, I think I understand your point now.

I could maybe write a "Configuration" section, where we give the default config file (as it is now), followed by a series of short explanations on how to edit Homer and make it more personal: how to remove/add a web service, how to use logo vs icon, how to change the API-based message etc. Either in the middle of the guide (under a "Configuration" section, or towards the end, after the default install, under a "Personalize your dashboard" chapter.

Wdyt?

guide/bonus/raspberry-pi/homer.md
```diff
@@ -0,0 +1,612 @@
...
+  $ sudo systemctl stop homer
+  $ sudo su - homer
+  ```
+
+* Fetch the latest GitHub repository information (v9.99.9 in this example), and update:
+
+  ```sh
+  $ cd homer
+  $ git fetch
+  $ git describe --tags --abbrev=0
```
kristapsk, line 516:
```suggestion
  $ git tag | sort --version-sort | tail -n 1
```

kristapsk (changes requested):

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: Thanks :+1:

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: I deleted Homer frm my node and rebuilt it from scratch.

I think the issue was when creating the symlink for the "tools" folder. There is a need to first delete the existing folder which contains sample pics. I added a rm line to do this just before creating the symlink.

I also shuffled a few commands to simplify the guide.

I went thourgh the whole guide and can display the homepage well with all the logos in my browser.

PTAL

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: I now place the chown command at the very end just before the first launch. Instead of chowning several times, it is now done only once which simplifies the guide.

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: I deleted Homer frm my node and rebuilt it from scratch.

I think the issue was when creating the symlink for the "tools" folder. There is a need to first delete the existing folder which contains sample pics. I added a rm line to do this just before creating the symlink.

I also shuffled a few commands to simplify the guide.

I went thourgh the whole guide and can display the homepage well with all the logos in my browser.

PTAL

VajraOfIndra:
> The `assets` and `tools` folders documentation still needs some correction



> In my review, I think I forgot about that 404 the page hits in the background... no idea what it is.
> 
> `https://raspibolt:4091/assets/manifest.json`

Where do you see this error? in the browser or in some logs?

The` manifest.json` is part of the `assets` directory. Did you delete the file or folder maybe?
I can't replicate this issue. I can see the file in my asset folder:
image

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: I created a brand new chapter that introduce users on how to personalize their dashboards and add new web service buttons. I think the guide is quite complete now :)

RooSoft:
> > The `assets` and `tools` folders documentation still needs some correction
> 
> > In my review, I think I forgot about that 404 the page hits in the background... no idea what it is.
> > `https://raspibolt:4091/assets/manifest.json`
> 
> Where do you see this error? in the browser or in some logs?
> 
> The` manifest.json` is part of the `assets` directory. Did you delete the file or folder maybe? I can't replicate this issue. I can see the file in my asset folder: image

In the browser

guide/bonus/raspberry-pi/homer.md
RooSoft: Done my homework and it seems to work...

Figuring out that the `ln` command acts differently if there is a trailing slash or not.. with a slash, it copies the contents of the assets folder. When it's not there, makes a link to the same folder instead and that's what's wanted and documented.

Well done!

RooSoft (approved):
Incredibly well crafted piece of documentation right from the get go.

guide/bonus/raspberry-pi/homer.md
VajraOfIndra: Ok thanks, I'll try to keep in mind that trailing slash effect on symlink creation 👍

VajraOfIndra:
Thanks a lot @RooSoft and @kristapsk for your reviews! 👍

It's now in master and live 🎉

Issue #919, opened 2022-02-14
[BUG] Bonus guide: Mempool - Nginx fails to start on reboot.


ghgeiger:
### Describe the bug
After successful install and start of mempool.space on RPI4/2G - SD/SSD raspibolt 3, upon reboot, Nginx fails to start and prevents all web interfaces from operating.

### Location in guide
[Link to the corresponding page and section on <https://raspibolt.org>](https://raspibolt.org/bonus/bitcoin/mempool.html)

### Expected behavior
Nginx and all web services function after reboot.

### Screenshots
<img width="1079" alt="Screen Shot 2022-02-07 at 1 37 06 PM" src="https://user-images.githubusercontent.com/60176205/153786884-770f618c-bda2-4278-87b6-8197618adcdd.png">

### Environment
 - Hardware: Raspberry Pi 4 - 2GB, OS on 64 GB microSD, Data on 1TB SSD
 - Operating system: Raspberry Pi OS 64 bit
 - Version: Debian GNU/Linux 11 (bullseye)

### Additional context
It seems that this occurred because nginx was expecting mempool to be operational, but mempool had not started yet because it is waiting for bitcoind to be functional. 

Rectified by changing the file:
`/lib/systemd/system/nginx.service`
to include the lines:
```
Restart=on-failure
RestartSec=30s 

```
under the line:
`Type=forking`

RooSoft:
Can you paste the output of that command here please?

```bash
cat -n /etc/nginx/snippets/nginx-mempool.conf | sed -n 45,55p
```

RooSoft:
Mine contains that on line 49 and above... is yours similar?

```ini
    49		# mainnet API
    50		location /api/v1/donations {
    51			proxy_pass https://mempool.space;
    52		}
```

RooSoft:
`Host not found` can also be a DNS problem, as this URL seems to redirect to an external site... maybe also try this

```bash
ping mempool.space
```

and make sure it returns a valid IP

wiz:
@RooSoft this is my fault, we need to add a `resolver 8.8.8.8;` line in nginx.conf to enable dynamic on-demand DNS resolution, by default it tries to resolve names on startup which fails for most people because their network / internet connectivity is not working yet and it causes nginx to fail starting

RooSoft:
Is there a way to prevent this?

ghgeiger:
[...]

PR #980, opened 2022-04-09
Mempool nginx startup

RooSoft:
#### What

nginx cannot start on boot due to a bug introduced in the Mempool setup

Related to issue #919 

### Why

The configuration tries to reach a remote website while the DNS server isn't yet ready to receive requests

#### How

In the systemd nginx config, add a prerequisite step that makes sure the query goes through

#### Scope

- [X] simple bug fix

Fixes # (link issue)

#### Test & maintenance

Restart the raspibolt and make sure the mempool site works at https://raspibolt:4081

#### Animated GIF (optional)

https://media.giphy.com/media/XSVyi8jXJcv4ZEvOas/giphy-downsized-large.gif

### DIFF
--- a/guide/bonus/bitcoin/mempool.md
+++ b/guide/bonus/bitcoin/mempool.md
@@ -399,7 +399,7 @@ We now need to modify the nginx configuration to create a web server for the web
   $ sudo nano /lib/systemd/system/nginx.service
   ```
 
-  Add this line to the nginx systemd configuration file under the `[Service]` section
+  Add this line under the `[Service]` section
 
   ```ini
   ExecStartPre=/bin/bash -c 'until host mempool.space; do sleep 1; done;'
--- a/guide/bonus/bitcoin/mempool.md
+++ b/guide/bonus/bitcoin/mempool.md
@@ -393,11 +393,13 @@ We now need to modify the nginx configuration to create a web server for the web
   }
   ```
 
-* Add this line to the nginx systemd configuration file under the `[Service]` section
+* Modify the nginx systemd configuration file
 
   ```bash
   $ sudo nano /lib/systemd/system/nginx.service
   ```
+
+  Add this line to the nginx systemd configuration file under the `[Service]` section
 
   ```ini
   ExecStartPre=/bin/bash -c 'until host mempool.space; do sleep 1; done;'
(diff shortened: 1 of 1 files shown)


VajraOfIndra:
Fixes #919

Note: A [pull request to fix this bug](https://github.com/mempool/mempool/pull/1455) has been opened 16 days ago in the Mempool GitHub repository . We'll have to monitor if this PR is merged in Mempool. If it is, we could then remove our own fix from the guide.

I don't have Mempool installed at the moment so I can't test it with a reboot myself.

VajraOfIndra:
As you pointed out earlier, delaying the launch of nginx until a public website is reachable is not ideal as other core programs rely on nginx (Electrs, RTL etc) 

[Someone else](https://github.com/mempool/mempool/issues/1452#issuecomment-1079953442) made the Mempool systemd service sleeps for 30 seconds before starting, allowing for the DNS to be up and running:
`ExecStartPre=/usr/bin/sleep 30`

Sounds like a good solution, wdyt?

RooSoft:
> sleeps for 30 seconds before starting

I'm against it.

As a developer, you quickly discover that sleep based solutions are the root of all evil.

This PR actively checks if the website si available, wherever a sleep might come to an end before communication becomes possible. These problems are hard to debug when problems happen several months after installation.

The active check isn't ideal either. I'd prefer not depending on that external site at all.

omfgd00d:
@RooSoft the PR itself does include a check but it is non-blocking and nginx will continue to run so I believe it is more robust than any others. FYI I authored this PR because of my raspibolt having this issue.

Stadicus:
Wiz commented on this proposed change in https://github.com/mempool/mempool/pull/1455#issuecomment-1098053556

> We solve this in a different better way using the resolver statement. Take a look at the production nginx.conf we just need to copy that one line over

Would the mentioned change, adding the following lines to the Mempool NGINX file, solve this issue?

```
# DNS servers for on-demand resolution, change if desired
	resolver 8.8.8.8;
```

VajraOfIndra:
FYI, another reported issue due to this bug: #1059

RooSoft:
> Would the mentioned change, adding the following lines to the Mempool NGINX file, solve this issue?
> 
> ```
> # DNS servers for on-demand resolution, change if desired
> 	resolver 8.8.8.8;
> ```

Their production nginx config has changed significantly since this spring... will try it out and adapt it to our needs in hopes it fixes this issue.

PR #992, opened 2022-04-15
Fix charge lnd.md cronjob cmd

mktech-gh:
#### What

In the cron job configuration, the path to the script is wrong.

### Why

The cron job doesn't work

#### How

Correction of the typo

#### Scope

- [x ] independent bonus guide
- [x ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

I did my install after this change

#### Animated GIF (optional)

:-)


VajraOfIndra (approved):
Thanks for spotting this typo! ACK

I saw you also added a couple of lines in the output of `charge-lnd -h`, they must have been added since the creation of the guide -> tACK

Let's go to master! :tada:

PR #994, opened 2022-04-20
System overview: Get running version of BTC RPC Explorer and RTL from `npm` not `git`

kristapsk:
#### What

Get current local installed / running BTC RPC Explorer and RTL version info from `npm` not `git`.

### Why

1) `git fetch` does not mean it's actually installed (`npm install`) and running;
2) `git describe tags` will not always return latest version.

#### Scope

- [ ] independent bonus guide
- [X] simple bug fix


Stadicus (approved):
tACK, nice addition. We might use that method in the main guide as well for the update sections.

PR #987, opened 2022-04-13
Bump Electrs version and update command to fetch latest release tag

kristapsk:
#### What

Bump Electrs version to v0.9.6 and update command to fetch latest release tag for upgrade guide.

### Why

Resolves #951. Fixes #970.

#### Scope

- [ ] significant change to core configuration
- [ ] independent bonus guide
- [X] simple bug fix

PR #995, opened 2022-04-20
Bump versions and update instructions to get latest release tag from `git` (RTL, charge-lnd, BoS)

kristapsk:
#### What

Bump latest version fo RTL, charge-lnd and BoS and update commands to get latest release tags for upgrades.

Also updated sample `git verify-tag` outputs. Signature date/time will change witch each versions, checking fingerprint is more important.

### Why

See #987, same as was done there for Electrs.

#### Scope

- [ ] independent bonus guide
- [X] simple bug fix

P.S. Could we write some bot that automatically opens version bump PRs?


VajraOfIndra (approved):
Nice improvements and version updates :+1: 

LGTm, utACK

PR #1021, opened 2022-05-18
Bonus guide: Fulcrum

DanielPX1:
#### What

Installation guide for Fulcrum server, alternative to electrs.

### Why

Alternative to electrum because of better performance, ex. for deep wallets etc. 

#### How

Following developer installation recommendations, testing and fixing issues.

#### Scope

- [x]  Review guide and formatting
- [ ] Approve merging

#### Test & maintenance

Test by following steps stated in the guide on your system. Maintenance no issue for me.


VajraOfIndra:
Thanks so much for writing this guide. Fulcrum is an important guide to have as in some cases it seems the only indexer that works properly.

I wished we could have reviewed this PR earlier but we are all quite busy at the same time so it did not happen.. I'll try to give it a try in the coming couple of weeks! Hang in there! :)

twofaktor:
Some suggestions for the guide. 

Should we use "rpccookie" config? :

`#rpccookie = /path/to/bitcoind/datadir/.cookie`

+ make "fulcrum" user and add to "bitcoin" group 

+ use SSL (port 50002) instead of 50001 (for more wallet compatibility)

VajraOfIndra:
> Some suggestions for the guide.
> 
>     * use "rpccookie" config :
> 
> 
> `#rpccookie = /path/to/bitcoind/datadir/.cookie`
> 
>     * make "fulcrum" user and add to "bitcoin" group
> 
>     * use SSL (port 50002) instead of 50001 (for more wallet compatibility)

Good ideas! They all seem like very sensible suggestions to me :+1: 

@DanielPX1, would you be keen to implement them? For the new user you could check how we usually do it by looking at the Electrs guide for example (or any other server program on the node as they all use a separate user).

DanielPX1:
> > Some suggestions for the guide.
> > ```
> > * use "rpccookie" config :
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `#rpccookie = /path/to/bitcoind/datadir/.cookie`
> > ```
> > * make "fulcrum" user and add to "bitcoin" group
> > 
> > * use SSL (port 50002) instead of 50001 (for more wallet compatibility)
> > ```
> 
> Good ideas! They all seem like very sensible suggestions to me 👍
> 
> @DanielPX1, would you be keen to implement them? For the new user you could check how we usually do it by looking at the Electrs guide for example (or any other server program on the node as they all use a separate user).

Sure! Great suggestions, I am not available until Monday, but when I am back for sure will work on that!

DanielPX1:
> > Some suggestions for the guide.
> > ```
> > * use "rpccookie" config :
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `#rpccookie = /path/to/bitcoind/datadir/.cookie`
> > ```
> > * make "fulcrum" user and add to "bitcoin" group
> > 
> > * use SSL (port 50002) instead of 50001 (for more wallet compatibility)
> > ```
> 
> Good ideas! They all seem like very sensible suggestions to me 👍
> 
> @DanielPX1, would you be keen to implement them? For the new user you could check how we usually do it by looking at the Electrs guide for example (or any other server program on the node as they all use a separate user).

@VajraOfIndra  @twofaktor 
I have added the suggested fixes, tested on my device and no problems, would someone be able to check if all is ok? 

All I am unsure of are permissions to fulcrum folder, which command would you suggest for making all files and subfiles executable. readable and writable? I personally used chmod 777 as I am not very familiar with them yet, but will change them as it is not secure to my understading

twofaktor:
> > > Some suggestions for the guide.
> > > ```
> > > * use "rpccookie" config :
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > `#rpccookie = /path/to/bitcoind/datadir/.cookie`
> > > ```
> > > * make "fulcrum" user and add to "bitcoin" group
> > > 
> > > * use SSL (port 50002) instead of 50001 (for more wallet compatibility)
> > > ```
> > 
> > Good ideas! They all seem like very sensible suggestions to me 👍
> > 
> > @DanielPX1, would you be keen to implement them? For the new user you could check how we usually do it by looking at the Electrs guide for example (or any other server program on the node as they all use a separate user).
> 
> @VajraOfIndra  @twofaktor 
> I have added the suggested fixes, tested on my device and no problems, would someone be able to check if all is ok? 
> 
> All I am unsure of are permissions to fulcrum folder, which command would you suggest for making all files and subfiles executable,readable and writable? I personally used chmod 777 as I am not very familiar with them yet, but will change them as it is not secure to my understading

--> My proposals: 

sudo chown -R fulcrum:fulcrum /data/fulcrum

&

Put only exact the necessary lines to include in bitcoin.conf for fulcrum work fine

twofaktor:
I updated my suggestions!

DanielPX1:
> I updated my suggestions @DanielPX1 !!!

Did not notice, will look into it!

twofaktor:
txindex=1 is not necessary in bitcoin.conf, because fullcrum creates its own index database, anyway txindex=1 was already configured in Bitcoin Core section. I would deleted

DanielPX1:
> is necessary to touch bitcoin.conf for fulcrum to work properly? what is de line/lines strictly necessaries?



> > > > Some suggestions for the guide.
> > > > ```
> > > > * use "rpccookie" config :
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > `#rpccookie = /path/to/bitcoind/datadir/.cookie`
> > > > ```
> > > > * make "fulcrum" user and add to "bitcoin" group
> > > > 
> > > > * use SSL (port 50002) instead of 50001 (for more wallet compatibility)
> > > > ```
> > > 
> > > 
> > > Good ideas! They all seem like very sensible suggestions to me 👍
> > > @DanielPX1, would you be keen to implement them? For the new user you could check how we usually do it by looking at the Electrs guide for example (or any other server program on the node as they all use a separate user).
> > 
> > 
> > @VajraOfIndra  @twofaktor
> > I have added the suggested fixes, tested on my device and no problems, would someone be able to check if all is ok?
> > All I am unsure of are permissions to fulcrum folder, which command would you suggest for making all files and subfiles executable,readable and writable? I personally used chmod 777 as I am not very familiar with them yet, but will change them as it is not secure to my understading
> 
> --> My suggestions :
> 
>     * `sudo chown -R fulcrum:fulcrum /data/fulcrum` instead of `sudo chown bitcoin:bitcoin /data/fulcrum/*`
> 
>     * is necessary to touch bitcoin.conf for fulcrum to work properly? what is de line/lines strictly necessaries?
> 
>     * is necessary this line?
>       `sudo ufw allow 8332`
>       never i has seen this line in Electrs section...
> 
>     * delete line * Nginx of requirements --> we have `* Generate cert and key files for SSL` section
> 
>     * in tor connect section: ` HiddenServicePort 50002 127.0.0.1:50002` instead of ` HiddenServicePort 50001 127.0.0.1:50001`
> 
>     * this section is correct?? should be it 50002 port instead of 50001 to use SSL??
>       `* Add following lines
> 
> 
> ```shell
> ### TOR
> tor_hostname=xyz... .onion
> tor_tcp_port=50001
> ```
> 
> ```shell
> $ sudo systemctl restart fulcrum.service
> $ sudo ufw allow 50001 <--- why this line is necessary? previously we openned 50002 port to SSL connection... ---> * Allow port SSL: ` sudo ufw allow 50002`
> ```
> 
>     * You should now be able to connect to your fulcrum server remotely via Tor using your hostname and port ### 50001` why???

1) Fixed permission command
2) To my understanding, rpcport and whitelist (used in electrum as well) should be specified but I might be wrong, deleted unnecessary lines and kept those so far. Deleted also rpcbind, apparently these are not neccesary, my fulcrum works without them. 
3) It was not neccessary, thanks
4) Deleted
rest) Should be ok now?

twofaktor:
I wrote directly in code this

For me, this lines are not necessary, `rpcuser / rpcpassword` never were mentioned in BTC core section and `rpcport=8332` is by default integrated in Bitcoin code config

DanielPX1:
> I wrote directly in code this
> 
> For me, this lines are not necessary, `rpcuser / rpcpassword` never were mentioned in BTC core section and `rpcport=8332` is by default integrated in Bitcoin code config

Thanks for feedback, I will therefore remove them:) When comes to rpcuser/password, I was using it for programs outside raspibolt (Dojo etc.) and run into problems with cookie auth, so thought it might be nice to mention it. But no problem removing it as well

twofaktor:
> > I wrote directly in code this
> > 
> > For me, this lines are not necessary, `rpcuser / rpcpassword` never were mentioned in BTC core section and `rpcport=8332` is by default integrated in Bitcoin code config
> 
> Thanks for feedback, I will therefore remove them:) When comes to rpcuser/password, I was using it for programs outside raspibolt (Dojo etc.) and run into problems with cookie auth, so thought it might be nice to mention it. But no problem removing it as well

I know, but currently for Raspibolt none of the services needs `rpcuser/rpcpassword` yet

DanielPX1:
> > > I wrote directly in code this
> > > For me, this lines are not necessary, `rpcuser / rpcpassword` never were mentioned in BTC core section and `rpcport=8332` is by default integrated in Bitcoin code config
> > 
> > 
> > Thanks for feedback, I will therefore remove them:) When comes to rpcuser/password, I was using it for programs outside raspibolt (Dojo etc.) and run into problems with cookie auth, so thought it might be nice to mention it. But no problem removing it as well
> 
> I know, but currently for Raspibolt none of the services needs `rpcuser/rpcpassword` yet

No prob, I understand, it is gone. Btw I figured out why i put $ sudo ufw allow 8332, without it being in place, samourai dojo cant receive blocks because of docker. I accidetanlly put it there thinking its for fulcrum

twofaktor:
new suggestions in code comments! @DanielPX1 , revise please! thanks!

DanielPX1:
> new suggestions in code comments! @DanielPX1 , revise please! thanks!

Hmmm I checked everything I could and do not see any of your comments, am I doing something wrong? Sorry I am still a noob when comes to this...

twofaktor:
> 
On top, in tab "Files changed" left to "Checks (0)" you can see the code and all my comments

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,336 @@
...
+### Corrupting a database after the finishning initial sync 
+
+* If you corrupted the fulcrum database after initial sync, copy your fulcrum database backup into db file 
+
+  ```sh
+  $ sudo systemctl stop fulcrum.service
+  $ sudo rm -r /data/fulcrum/fulcrum_db/*
+  $ sudo cp /data/fulcrum/fulcrum_db_backup/* /data/fulcrum/fulcrum_db/
+  $ sudo systemctl restart fulcrum.service
+  ```
```
DanielPX1, line 336: test

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,336 @@
...
+  $ sudo rm -r /data/fulcrum/fulcrum_db; sudo mkdir /data/fulcrum/fulcrum_db
+  $ sudo chown -R fulcrum:fulcrum /data/fulcrum/*
+  $ sudo systemctl restart fulcrum.service
+  ```
+  
+### Corrupting a database after the finishning initial sync 
+
+* If you corrupted the fulcrum database after initial sync, copy your fulcrum database backup into db file 
+
+  ```sh
```
DanielPX1, line 331: test

twofaktor (commented):
Check this!

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,336 @@
...
+   ```
+  
+* Download fulcrum for raspberry pi, open and unpackage it, move all files to our fulcrum directory
+ 
+  ```sh
+  $ cd /tmp
+  $ wget https://github.com/cculianu/Fulcrum/releases/download/v1.7.0/Fulcrum-1.7.0-arm64-linux.tar.gz
+  $ tar xvf Fulcrum-1.7.0-arm64-linux.tar.gz
+  $ sudo mv Fulcrum-1.7.0-arm64-linux/* /data/fulcrum
+  ```
```
twofaktor, line 93: Add: sudo chown -R fulcrum:fulcrum /data/fulcrum/

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,336 @@
...
+* If you corrupted the fulcrum database during initial sync (power outage, hard kill) you need to start from the scratch
+
+  ```sh
+  $ sudo systemctl stop fulcrum.service
+  $ sudo rm -r /data/fulcrum/fulcrum_db; sudo mkdir /data/fulcrum/fulcrum_db
+  $ sudo chown -R fulcrum:fulcrum /data/fulcrum/*
+  $ sudo systemctl restart fulcrum.service
+  ```
+  
+### Corrupting a database after the finishning initial sync
```
twofaktor, line 327: Same question relationed with line 295

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,249 @@
...
+  worker_threads = 1
+  deb_mem=1024.0
+  db_max_open_files=60
+  fast-sync = 1024
+  
+  # for 8GB RAM
+  #db_max_open_files=100
+  #fast-sync = 2048
+  ```
+
```
twofaktor, line 118: Include: 
Exit the “fulcrum” user session.

`$ exit`

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,249 @@
...
+  deb_mem=1024.0
+  db_max_open_files=60
+  fast-sync = 1024
+  
+  # for 8GB RAM
+  #db_max_open_files=100
+  #fast-sync = 2048
+  ```
+  
+* Allow port SSL:
```
twofaktor, line 119: Change to:

`As user "admin", configure the firewall to allow incoming requests:`

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,249 @@
...
+
+  ```sh
+  $ sudo ufw allow 50002
```
twofaktor, line 122: Change to:

`$ sudo ufw allow 50002/tcp comment 'allow Fulcrum SSL'`

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,249 @@
...
+  ```
+  
+* Now we have configured our conf file for fulcrum, however we need to set up fulcrum to start automatically by creating fulcrum service and set up configuration file
```
twofaktor, line 125: Change to: 

```
## Autostart on boot
Electrs needs to start automatically on system boot.
* As user "admin", create the Fulcrum systemd unit and copy/paste the following configuration. Save and exit.
```

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,249 @@
...
+  [Install]
+  WantedBy=multi-user.target
+  ```
+  
+* If you are booting from SD card, you will not be able to execute from SSD as it is not permitted. You will achieve that deleting "noexec" line in fstab file, add permissions to the folders if you will encounter errors later during start of fulcrum
+  
+  ```sh
+  $ sudo nano /etc/fstab
+  ```
+
```
twofaktor, line 157: Add final result: 

`UUID=123456 /data ext4 rw,nosuid,dev,noatime,nodiratime,auto,nouser,async,nofail 0 2`

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,249 @@
...
+  $ sudo systemctl start fulcrum.service
+  ```
+
+* We can check if everything goes right using these commands
+
+  ```sh
+  $ sudo systemctl status fulcrum.service
+  $ sudo journalctl -fu fulcrum.service
+  ```
+
```
twofaktor, line 189: Do you have a output example like this of Fulcrum?:

````sh
  Starting electrs 0.9.4 on aarch64 linux with Config { network: Bitcoin, db_path: "/data/electrs/db/bitcoin", daemon_dir: "/home/bitcoin/.bitcoin", daemon_auth: CookieFile("/home/bitcoin/.bitcoin/.cookie"), daemon_rpc_addr: 127.0.0.1:8332, daemon_p2p_addr: 127.0.0.1:8333, electrum_rpc_addr: 127.0.0.1:50001, monitoring_addr: 127.0.0.1:4224, wait_duration: 10s, jsonrpc_timeout: 15s, index_batch_size: 10, index_lookup_limit: Some(1000), reindex_last_blocks: 0, auto_reindex: true, ignore_mempool: false, sync_once: false, disable_electrum_rpc: false, server_banner: "Welcome to electrs 0.9.4 (Electrum Rust Server)!", args: [] }
  [2021-11-09T07:09:42.744Z INFO  electrs::metrics::metrics_impl] serving Prometheus metrics on 127.0.0.1:4224
  [2021-11-09T07:09:42.744Z INFO  electrs::server] serving Electrum RPC on 127.0.0.1:50001
  [2021-11-09T07:09:42.812Z INFO  electrs::db] "/data/electrs/db/bitcoin": 0 SST files, 0 GB, 0 Grows
  [2021-11-09T07:09:43.174Z INFO  electrs::index] indexing 2000 blocks: [1..2000]
  [2021-11-09T07:09:44.665Z INFO  electrs::chain] chain updated: tip=00000000dfd5d65c9d8561b4b8f60a63018fe3933ecb131fb37f905f87da951a, height=2000
  [2021-11-09T07:09:44.986Z INFO  electrs::index] indexing 2000 blocks: [2001..4000]
  [2021-11-09T07:09:46.191Z INFO  electrs::chain] chain updated: tip=00000000922e2aa9e84a474350a3555f49f06061fd49df50a9352f156692a842, height=4000
  [2021-11-09T07:09:46.481Z INFO  electrs::index] indexing 2000 blocks: [4001..6000]
  [2021-11-09T07:09:47.581Z INFO  electrs::chain] chain updated: tip=00000000dbbb79792303bdd1c6c4d7ab9c21bba0667213c2eca955e11230c5a5, height=6000
  ...
  ```
 Would be fine show it here.

File: guide/bonus/bitcoin/Fulcrum.md [hunk 11]
```diff
@@ -0,0 +1,249 @@
...
+ 
+* Set swapfile to defaults after finishing db sync - it will then be created dynamically, comment following lines
+
+  ```sh
+  $ sudo nano /etc/dphys-swapfile
+  ```
+  
+  ```sh
+  #CONF_SWAPSIZE=10000
+  #CONF_MAXSWAP=2000
```
twofaktor, line 207: Change to:

`#CONF_MAXSWAP=10000`

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,249 @@
...
+ $ sudo systemctl reload tor
+ $ sudo cat /var/lib/tor/hidden_service_fulcrum/hostname
+ > abcdefg..............xyz.onion
+ ```
+ 
+ * You should now be able to connect to your fulcrum server remotely via Tor using your hostname and port 50002
+
+### Backup the database
+
+Because the sync can take up to 5 days and more, it is recommended to have at least any backup of the database. It doesnt need to be the latest one and you can backup only once, it is still better to sync for a few hours instead of week (from scratch). Should be done on external drive.
```
twofaktor, line 249: Typo:
*doesn't

twofaktor (commented):
Pending review

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,300 @@
...
+ > abcdefg..............xyz.onion
+ ```
+ 
+ * You should now be able to connect to your fulcrum server remotely via Tor using your hostname and port 50002
+
+### Backup the database
+
+Because the sync can take up to 5 days and more, it is recommended to have at least any backup of the database. It doesn't need to be the latest one and you can backup only once, it is still better to sync for a few hours instead of week (from scratch). Should be done on external drive.
+
+### Uninstall Fulcrum
```
twofaktor, line 266: Change to: ## Uninstall Fulcrum

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,310 @@
...
+  
+  * Set following values in zram configuration file
+  
+  ```sh
+  $ sudo nano /etc/default/zram-swap
+  ```
+  
+  ```sh
+  # override fractional calculations and specify a fixed swap size 
+  _zram_fixedsize="6G"¿?
```
twofaktor, line 175: ¿?...  is my doubt, in the example of the zram repository puts 6GB, how many size of zram do you think that is the correct? 6? 10?

DanielPX1: hmmm not sure about the zram size, I will look

> ¿?... is my doubt, in the example of the zram repository puts 6GB, how many size of zram do you think that is the correct? 6? 10?

Sorry, will fix it.. I did not notice it. Hmmm, not sure about the size tho, will look what is recommended

DanielPX1: Does it need to be fixed tho? Can it be created dynamically?

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,310 @@
...
+  _zram_fixedsize="6G"
+ 
+  # compression algorithm to employ (lzo, lz4, zstd, lzo-rle) 
+  _zram_algorithm="lz4"
+  ```
+  
+  ```sh
+  $ systemctl restart zram-swap.service
+  ```
+
```
twofaktor, line 184: Add command for ensure that's working fine:

`sudo cat /proc/swaps `

And output example:
```
Filename                                Type                Size           Used    Priority
/var/swap                                  file                 102396         0       -2
/dev/zram0                             partition           4899744        0       5
```

guide/bonus/bitcoin/Fulcrum.md
DanielPX1: it is in there in line 70?

guide/bonus/bitcoin/Fulcrum.md
DanielPX1: done

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,322 @@
...
+  $ cd zram-swap && sudo ./install.sh
+  ```
+  
+  * Set following values in zram configuration file
+  
+  ```sh
+  $ sudo nano /etc/default/zram-swap
+  ```
+  
+  ```sh
```
twofaktor, line 173: Chage to only:

```
# portion of system ram to use as zram swap (expression: "1/2", "2/3", "0.5", etc)
# default: "1/2"
_zram_fraction="1/2"

```
_zram_algorithm by default is already "lz4":

```
# compression algorithm to employ (lzo, lz4, zstd, lzo-rle)
# default: lz4
_zram_algorithm="lz4"
```

guide/bonus/bitcoin/Fulcrum.md
DanielPX1: this part was my mistake, deleted it all

guide/bonus/bitcoin/Fulcrum.md
twofaktor: Yeah, is correct in line 70

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,295 @@
...
+zram-swap is neccesary for proper functioning of fulcrum during sync process
+  
+  * Ensure that you are logged with user "admin", clone and install zram-swap
+  
+  ```sh
+  $ git clone https://github.com/foundObjects/zram-swap.git 
+  $ cd zram-swap && sudo ./install.sh
+  ```
+  
+  * Set following values in zram configuration file
```
twofaktor, line 161: Sorry, the default config is OK, delete this section, including command until to next line 174

DanielPX1: Did not notice that, I have removed it now

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,279 @@
...
+  $ sudo mkdir /data/fulcrum
+  $ sudo mkdir /data/fulcrum/fulcrum_db
+   ```
+  
+* Download fulcrum for raspberry pi, open and unpackage it, move all files to our fulcrum directory
+ 
+  ```sh
+  $ cd /tmp
+  $ wget https://github.com/cculianu/Fulcrum/releases/download/v1.7.0/Fulcrum-1.7.0-arm64-linux.tar.gz
+  $ tar xvf Fulcrum-1.7.0-arm64-linux.tar.gz
```
Stadicus, line 68: The `/data` directory should only hold data, not executable applications. These are by default stored in `/usr/local/bin`.

We can just move these two files first, then everything else stays the same:

```sh
$ sudo mv Fulcrum-1.7.0-arm64-linux/Fulcrum Fulcrum-1.7.0-arm64-linux/FulcrumAdmin /usr/local/bin
```

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,279 @@
...
+* We will create fulcrum user and add him to bitcoin group with user "admin"
+
+  ```sh
+  $ sudo adduser --disabled-password --gecos "" fulcrum
+  $ sudo adduser fulcrum bitcoin
+  ```
+
+* Create a following folders
+
+  ```sh
```
Stadicus, line 58: These two commands could be combined:

```sh
$ sudo mkdir -p /data/fulcrum/fulcrum_db
```

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,279 @@
...
+  $ sudo su - fulcrum
+  $ cd /data/fulcrum
+  $ openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout key.pem -out cert.pem
+  ```
+
+* Next, we have to set up our fulcrum configurations. Troubles could be found without optimizations for raspberry pi. Choose either one for raspberry 4GB or 8GB depending on your hardware. Create the config file with the following content:
+
+  ```sh
+  $ sudo nano /data/fulcrum/fulcrum.conf
```
Stadicus, line 84: User `fulcrum` cannot use `sudo`

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,279 @@
...
+  ```
+  
+* As user "admin", configure the firewall to allow incoming requests:
+
+  ```sh
+  $ sudo ufw allow 50002/tcp comment 'allow Fulcrum SSL'
+  ```
+  
+## Autostart on boot
+Electrs needs to start automatically on system boot.
```
Stadicus, line 123: s/Electrs/Fulcrum/

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,279 @@
...
+* As user "admin", create the Fulcrum systemd unit and copy/paste the following configuration. Save and exit.
+
+  ```sh
+  $ sudo nano /etc/systemd/system/fulcrum.service
+  ```
+  
+  ```sh
```
Stadicus, line 130: Add the standard header:

```
# RaspiBolt: systemd unit for Fulcrum
# /etc/systemd/system/fulcrum.service
```

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,279 @@
...
+  [Unit]
+  Description=Fulcrum
+  Wants=bitcoind.service
+  After=bitcoind.service
+  StartLimitBurst=2
+  StartLimitIntervalSec=20
+
+  [Service]
+  ExecStart=/data/fulcrum/Fulcrum /data/fulcrum/fulcrum.conf
```
Stadicus, line 139: Should now be 

```
ExecStart=/usr/local/bin/Fulcrum /data/fulcrum/fulcrum.conf
```

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,279 @@
...
+  [Install]
+  WantedBy=multi-user.target
+  ```
+  
+### Install zram-swap
+zram-swap is neccesary for proper functioning of fulcrum during sync process
+  
+  * Ensure that you are logged with user "admin", clone and install zram-swap
+  
+  ```sh
```
Stadicus, line 156: In what user should the user be at this point? I think we should make that explicit, e.g. `/tmp` or admin home

```
$ cd /tmp
```

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,279 @@
...
+{:toc}
+
+---
+
+
+## Requirements
+
+* Bitcoin
+* Little over 100GB of free storage for database (external backup recommended)
+
```
Stadicus, line 40: I think it would be good to mention that this is a replacement for Electrs, and these two services cannot be run at the same time (due to the same standard ports used).

Stadicus (commented):
Great guide! I made it up until sync now and left a few comments for this first part. :)

guide/bonus/bitcoin/Fulcrum.md
twofaktor: Change to:` nano /data/fulcrum/fulcrum.conf`

DanielPX1:
Thanks for feedback! Will fix that

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,279 @@
...
+  $ sudo mkdir /data/fulcrum/fulcrum_db
+   ```
+  
+* Download fulcrum for raspberry pi, open and unpackage it, move all files to our fulcrum directory
+ 
+  ```sh
+  $ cd /tmp
+  $ wget https://github.com/cculianu/Fulcrum/releases/download/v1.7.0/Fulcrum-1.7.0-arm64-linux.tar.gz
+  $ tar xvf Fulcrum-1.7.0-arm64-linux.tar.gz
+  $ sudo mv Fulcrum-1.7.0-arm64-linux/* /data/fulcrum
```
twofaktor, line 69: Change to: sudo mv Fulcrum-1.7.0-arm64-linux/Fulcrum Fulcrum-1.7.0-arm64-linux/FulcrumAdmin /usr/local/bin

guide/bonus/bitcoin/Fulcrum.md
twofaktor: Add: `$ cd /tmp`

DanielPX1:
Alright @twofaktor @Stadicus, should be fixed

twofaktor:
Hello guys, I'm not sure what's the correct configuration to API connect to BTC RPC Explorer of Fulcrum, I think that is necessary to change this lines of .env file conf of BTC RPC Explorer, because Fulcrum use directly SSL connection:

#BTCEXP_ADDRESS_API=(electrum|electrumx|blockchain.com|blockchair.com|blockcypher.com)
#BTCEXP_ELECTRUM_SERVERS=tls://electrumx.server.com:50002,tcp://127.0.0.1:50001,...

Do you know what's the correct configuration? @Stadicus ??

I think it is like that, but i'm not sure:

BTCEXP_ADDRESS_API=electrum
BTCEXP_ELECTRUM_SERVERS=tls://127.0.0.1:50002

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,282 @@
...
+zram-swap is neccesary for proper functioning of fulcrum during sync process
+  
+  * Ensure that you are logged with user "admin", clone and install zram-swap
+  
+  ```sh
+  $ cd /tmp
+  $ git clone https://github.com/foundObjects/zram-swap.git 
+  $ cd zram-swap && sudo ./install.sh
+  ```
+
```
twofaktor, line 163: Add here: 

* Add kernel parameters to make better use of ZRAM

```sh
$ sudo nano /etc/sysctl.conf
```
Here are the lines you’ll want to add at the end of your /etc/sysctl.conf file:
```
vm.vfs_cache_pressure=500
vm.swappiness=100
vm.dirty_background_ratio=1
vm.dirty_ratio=50
```
Then reboot, or enable with:
```sh
$ sudo sysctl --system
```

twofaktor (commented):
Check the suggestions in code

DanielPX1:
Done (y)

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,314 @@
...
+  StartLimitBurst=2
+  StartLimitIntervalSec=20
+
+  [Service]
+  ExecStart=/usr/local/bin/Fulcrum /data/fulcrum/fulcrum.conf
+  KillSignal=SIGINT
+  User=fulcrum
+  Type=exec
+  TimeoutStopSec=300
+  RestartSec=5
```
Stadicus, line 146: I'd recommend to increase this number to 30 seconds or more. Otherwise, `systemd` refuse to start it more than a few times with the error message "Start request repeated too quickly"

```log
admin@raspibolt3:~ $ sudo journalctl -fu fulcrum.service
-- Journal begins at Sat 2022-07-23 05:55:52 BST. --
Aug 01 13:20:27 raspibolt3 Fulcrum[1057]: [2022-08-01 13:20:27.841] Shutdown complete
Aug 01 13:20:27 raspibolt3 systemd[1]: fulcrum.service: Main process exited, code=exited, status=1/FAILURE
Aug 01 13:20:27 raspibolt3 systemd[1]: fulcrum.service: Failed with result 'exit-code'.
Aug 01 13:20:27 raspibolt3 systemd[1]: fulcrum.service: Consumed 1.318s CPU time.
Aug 01 13:20:54 raspibolt3 systemd[1]: fulcrum.service: Scheduled restart job, restart counter is at 2.
Aug 01 13:20:54 raspibolt3 systemd[1]: Stopped Fulcrum.
Aug 01 13:20:54 raspibolt3 systemd[1]: fulcrum.service: Consumed 1.318s CPU time.
Aug 01 13:20:54 raspibolt3 systemd[1]: fulcrum.service: Start request repeated too quickly.
Aug 01 13:20:54 raspibolt3 systemd[1]: fulcrum.service: Failed with result 'exit-code'.
Aug 01 13:20:54 raspibolt3 systemd[1]: Failed to start Fulcrum.
```

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,282 @@
...
+  Jul 28 12:20:19 rasp Fulcrum[181811]: [2022-07-28 12:20:19.780] Initializing header merkle cache ...
+  Jul 28 12:20:21 rasp Fulcrum[181811]: [2022-07-28 12:20:21.643] Checking tx counts ...
+  ```
+  
+  DO NOT REBOOT OR STOP THE SERVICE DURING DB CREATION PROCESS. YOU MAY CORRUPT THE FILES -
+  in case of that happening, start sync from scratch by deleting and recreating fulcrum_db file.
+  
+## After Installation
+Continue after fulcrum db sync is finished 
+
```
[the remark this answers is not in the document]
Stadicus: Two notes:

* I'd make the `cd` command explicit, e.g., `cd /tmp/zram-swap`
* The `/tmp` folder is purged on each reboot, so this directory might no longer be available.

Stadicus:
> Hello guys, I'm not sure what's the correct configuration to API connect to BTC RPC Explorer of Fulcrum, I think that is necessary to change this lines of .env file conf of BTC RPC Explorer, because Fulcrum uses directly SSL connection:` #ssl = 0.0.0.0:50002` 
> 
> #BTCEXP_ADDRESS_API=(electrum|electrumx|blockchain.com|blockchair.com|blockcypher.com) #BTCEXP_ELECTRUM_SERVERS=tls://electrumx.server.com:50002,tcp://127.0.0.1:50001,...
> 
> Do you know what's the correct configuration? @Stadicus ??
> 
> I think it is like that, but i'm not sure:
> 
> BTCEXP_ADDRESS_API=electrum BTCEXP_ELECTRUM_SERVERS=tls://127.0.0.1:50002

Good point. I'll check once Fulcrum has fully synced.

I had to stop the sync after 3.5 days, as I configured it for an 8GB Pi instead of the required 4GB Pi... During that time, Fulcrum synced up to 64%. Unfortunately, even with the 6 minutes allowed for shutdown, Fulcrum did not shut down properly on a system unit restart and corrupted the database. So I have to start synching from start.

guide/bonus/bitcoin/Fulcrum.md
DanielPX1: > > Hello guys, I'm not sure what's the correct configuration to API connect to BTC RPC Explorer of Fulcrum, I think that is necessary to change this lines of .env file conf of BTC RPC Explorer, because Fulcrum uses directly SSL connection:` #ssl = 0.0.0.0:50002`
> > #BTCEXP_ADDRESS_API=(electrum|electrumx|blockchain.com|blockchair.com|blockcypher.com) #BTCEXP_ELECTRUM_SERVERS=tls://electrumx.server.com:50002,tcp://127.0.0.1:50001,...
> > Do you know what's the correct configuration? @Stadicus ??
> > I think it is like that, but i'm not sure:
> > BTCEXP_ADDRESS_API=electrum BTCEXP_ELECTRUM_SERVERS=tls://127.0.0.1:50002
> 
> Good point. I'll check once Fulcrum has fully synced.
> 
> I had to stop the sync after 3.5 days, as I configured it for an 8GB Pi instead of the required 4GB Pi... During that time, Fulcrum synced up to 64%. Unfortunately, even with the 6 minutes allowed for shutdown, Fulcrum did not shut down properly on a system unit restart and corrupted the database. So I have to start synching from start.

Yeah it can happen during the sync, I recommend using signals like SIGINT as stated in the guide - after synchronisation is finished, to ensure it will not happen again. 

Inspired from my question: https://github.com/cculianu/Fulcrum/issues/115

guide/bonus/bitcoin/Fulcrum.md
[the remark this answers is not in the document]
twofaktor: What is the permanent folder that you are suggesting? /download folder for example?

guide/bonus/bitcoin/Fulcrum.md
Stadicus: Yes, although a bit arbitrary, `RestartSec=30` works fine for me (even on repeated restarts due to the corrupted database).

Stadicus:
Wow, the two settings for 4GB and 8GB really make a huge difference. On the 4GB optimizations, I'm already at 62% after 17 hours (compared to 64% with the wrong 8GB settings after 3.5 days)

DanielPX1:
> Wow, the two settings for 4GB and 8GB really make a huge difference. On the 4GB optimizations, I'm already at 62% after 17 hours (compared to 64% with the wrong 8GB settings after 3.5 days)

Yeah its really a difference, I am syncing BTC core on my 4GB node, so tommorow should test out the guide myself as well

guide/bonus/bitcoin/Fulcrum.md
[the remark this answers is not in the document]
twofaktor: If you choose subfolder inside `/tmp` we will be the same problem because `/zram-swap` folder will be purged too, right? should be better to choose other permanet folder e.g `/download` folder for install & uninstall?

Stadicus:
Unfortunately, my 4GB Pi crashed during the Fulcrum sync, and the database got corrupted again. I restarted the sync with all other services disabled, I hope it will go through this time.

The reason also might be that the Pi is only cooled passively. Not an issue, usually, but it gets pretty hot during the intense sync...

DanielPX1:
> Unfortunately, my 4GB Pi crashed during the Fulcrum sync, and the database got corrupted again. I restarted the sync with all other services disabled, I hope it will go through this time.
> 
> The reason also might be that the Pi is only cooled passively. Not an issue, usually, but it gets pretty hot during the intense sync...

hmmm that is strange. I am going to try it out now, so we will see. My raspi is cooled actively

DanielPX1:
> Unfortunately, my 4GB Pi crashed during the Fulcrum sync, and the database got corrupted again. I restarted the sync with all other services disabled, I hope it will go through this time.
> 
> The reason also might be that the Pi is only cooled passively. Not an issue, usually, but it gets pretty hot during the intense sync...

I still suspect low amount of swap being a problem. When I was using swapfile - 10GB allocation was great to sync with no issues and then switch it back to default once done. However at 2GB,4GB swap I had corruption issues same as you. 

I am now testing out the guide and syncing fulcrum. I allocated 10GB for zramswap. So we will see (I use rpi 4GB).
It is possible that different amount (lower) may be enough for 8GB RAM version. Not sure yet. Will give feedback soon, right now I am at 40%

Stadicus:
The Pi was running very hot, around 80 degrees celsius. I added active cooling now and started syncing again, with the same settings.

Stadicus:
Ok, active cooling did not help, the sync crashed at 57%. As `zram` simply creates a compressed volume within the regular RAM, its size can't be increased. As I'm wary to just up the regular swap file size, I'll test with the following conservative Fulcrum config first:

```config
db_mem = 256.0
db_max_open_files = 20
fast-sync = 0
db_use_fsync = true
```

DanielPX1:
> Ok, active cooling did not help, the sync crashed at 57%. As `zram` simply creates a compressed volume within the regular RAM, its size can't be increased. As I'm wary to just up the regular swap file size, I'll test with the following conservative Fulcrum config first:
> 
> ```
> db_mem = 256.0
> db_max_open_files = 20
> fast-sync = 0
> db_use_fsync = true
> ```
> 
> If this works, I hope to be able to use `fast-sync` with a small value of `256` as mentioned in [cculianu/Fulcrum#96 (comment)](https://github.com/cculianu/Fulcrum/issues/96#issuecomment-1099420035).
> 
> Note, @openoms mentioned a few weeks ago that Fulcrum simply might not be ready for 4GB systems: [rootzoll/raspiblitz#2924 (comment)](https://github.com/rootzoll/raspiblitz/issues/2924#issuecomment-1179598840). I wonder if that's still the case, or if we actually can get it to work.

I have set zram to 10GB and right now I am 84% into sync with guide settings, wanna try it?

Stadicus:
> I have set zram-swap to 10GB

Interesting. I'm not sure what this actually does, though. Zram creates a swap partition within the RAM, compressing it to make it (virtually) bigger. If you only have 4GB RAM, it cannot possibly create a 10GB partition within that. Does it then automatically use the SSD swap partition as well? Or do I have a basic misunderstanding here?

Stadicus:
My very conservative settings seem to work so far as well, but it's pretty slow. I'm at 64% (the previous breaking point) after two days, but with very few resources used:

image

DanielPX1:
@Stadicus me neither, It is just what I know from experimenting with swap. I have experienced crashes on 4GB system with lower amounts of it. However I noticed that with lower amounts of swap, I had 300mb of RAM available and kept decreasing until crash. Now with 10GB swap I had stable ~1GB free RAM everytime I checked my system. Does not zram-swap behave as a normal swap? 

_Swap is a space on a disk that is used when the amount of physical RAM memory is full. When a Linux system runs out of RAM, inactive pages are moved from the RAM to the swap space. Swap space can take the form of either a dedicated swap partition or a swap file_ - wikipedia

I have today successfully sinced fulcrum with this setting

obrazek

twofaktor:
We have a new tested and possibly a final version of the guide. Could someone proceed to review it? @VajraOfIndra @Stadicus @kristapsk??

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+vm.dirty_background_ratio=1
+vm.dirty_ratio=50
+```
+
+* Then apply the changes with
+
+```sh
+$ sudo sysctl --system
+```
+
```
Stadicus, line 98: In my case, `zram-swap` was not active here. I think it needs to be started (or at least restarted with the changed config) first. Using `restart` works for both cases (not running yet, or already running)

```sh
$ sudo systemctl restart zram-swap
```

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+```sh
+$ sudo ufw allow 50002 comment 'allow Fulcrum SSL'
+```
+
+### Configure Bitcoin Core
+
+We need to set up settings in Bitcoin Core configuration file - add new lines if they are not present
+
+* In `bitcoin.conf`, add the following line under `whitelist=download@127.0.0.1          # for Electrs`. Save and exit.
```
Stadicus, line 144: Is this a copy/paste artefact, or is whitelisting also required for Fulcrum? Then I'd change the comment

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+has_toc: false
+---
+
+## Bonus guide: Fulcrum server
+
+{: .no_toc }
+
+---
+
+[Fulcrum](https://github.com/cculianu/Fulcrum){:target="_blank"} Fulcrum is a fast & nimble SPV server for Bitcoin Cash & Bitcoin BTC created by Calin Culianu. It can be used as an alternative to Electrs because of its performance, as we can see in Craig Raw's [comparison](https://www.sparrowwallet.com/docs/server-performance.html) of servers
```
Stadicus, line 16: The second "Fulcrum" word could be deleted

Stadicus: Add `{:target="_blank"}` to the Craig Raw link

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+
+* With user "admin", make sure that all necessary software packages are installed
+
+```sh
+$ sudo apt install libssl-dev
+```
+
+### Install zram-swap
+
+zram-swap is necessary for the proper functioning of Fulcrum during sync process using compressed swap in memory (increase performance when memory usage is high)
```
Stadicus, line 57: Maybe extend to "compressed swap in memory and on disk", otherwise the 10GB size could be confusing

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+
+* Access to admin home folder, clone the repository of Github and install zram-swap
```
Stadicus, line 59: nit :)

s/Github/GitHub

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+
+```sh
+$ cd /home/admin/
+$ git clone https://github.com/foundObjects/zram-swap.git 
+$ cd zram-swap && sudo ./install.sh
+```
+
+* Set following size value in zram configuration file. Save a exit
```
Stadicus, line 67: s/Save a exit/Save and exit/

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+_zram_fixedsize="10G" #Uncomment and edit
+```
+
+* Add kernel parameters to make better use of zram
+
+```sh
+$ sudo nano /etc/sysctl.conf
+```
+
+* Here are the lines you’ll want to add at the end of your /etc/sysctl.conf file to make better use of zram. Save a exit
```
Stadicus, line 84: s/Save a exit/Save and exit/

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+
+```sh
+$ sudo systemctl restart bitcoind
+```
+
+## Installation
+
+### Download and set up Fulcrum
+
+We have our bitcoin core configuration file set up and now we can move to next part - installation of Fulcrum
```
Stadicus, line 164: s/bitcoin core/Bitcoin Core/

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+$ sha256sum --check Fulcrum-1.7.0-arm64-linux.tar.gz.sha256sum
+> Fulcrum-1.7.0-arm64-linux.tar.gz: OK
+```
+
+* Install Fulcrum and check the correct installation requesting the version
+
+```sh
+$ tar -xvf Fulcrum-1.7.0-arm64-linux.tar.gz
+$ sudo install -m 0755 -o root -g root -t /usr/local/bin Fulcrum-1.7.0-arm64-linux/Fulcrum Fulcrum-1.7.0-arm64-linux/FulcrumAdmin 
+$ ./Fulcrum --version
```
Stadicus, line 203: The executable `Fulcrum` should be called from `/usr/local/bin`. Therefore, it must be called as follows, not referencing the `./` current directory:

```sh
$ Fulcrum --version
```

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+* Change to fulcrum data folder and generate cert and key files for SSL
+
+```sh
+$ cd /data/fulcrum
+$ openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout key.pem -out cert.pem
+```
+
+### Configuration
+
+* Next, we have to set up our fulcrum configurations. Troubles could be found without optimizations for Raspberry Pi. Choose either one for Raspberry 4GB or 8GB depending on your hardware. Create the config file with the following content. Save a exit
```
Stadicus, line 242: s/fulcrum/Fulcrum/
s/Save a exit/Save and exit/

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+...
+```
+
+Fulcrum will now index the whole Bitcoin blockchain so that it can provide all necessary information to wallets. With this, the wallets you use no longer need to connect to any third-party server to communicate with the Bitcoin peer-to-peer network.
+
+DO NOT REBOOT OR STOP THE SERVICE DURING DB CREATION PROCESS. YOU MAY CORRUPT THE FILES - in case of that happening, start sync from scratch by deleting and recreating `fulcrum_db` folder.
+
+💡 Fulcrum must first fully index the blockchain and compact its database before you can connect to it with your wallets. This can take a few hours. Only proceed with the [Desktop Wallet Section](desktop-wallet.md) once Fulcrum is ready.
+
+💡 After the initial sync of fulcrum, if you want to still use zram, you can return to the default zram config.
```
Stadicus, line 359: s/fulcrum/Fulcrum

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+
+* As user "admin", access to zram config again and return to default config. Save a exit
```
Stadicus, line 361: s/Save a exit/Save and exit./

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+```sh
+_zram_fraction="1/2"   #Uncomment this line 
+#_zram_fixedsize="10G" #Comment this line
+```
+
+* Then apply the changes with
+
+```sh
+$ sudo sysctl --system
+```
```
Stadicus, line 376: For me, this command does not apply the updated zram settings

```sh
admin@raspibolt3:~/zram-swap $ sudo sysctl --system
* Applying /usr/lib/sysctl.d/50-pid-max.conf ...
kernel.pid_max = 4194304
* Applying /etc/sysctl.d/98-rpi.conf ...
kernel.printk = 3 4 1 3
vm.min_free_kbytes = 16384
net.ipv4.ping_group_range = 0 2147483647
* Applying /etc/sysctl.d/99-sysctl.conf ...
vm.vfs_cache_pressure = 500
vm.swappiness = 100
vm.dirty_background_ratio = 1
vm.dirty_ratio = 50
* Applying /usr/lib/sysctl.d/protect-links.conf ...
fs.protected_fifos = 1
fs.protected_hardlinks = 1
fs.protected_regular = 2
fs.protected_symlinks = 1
* Applying /etc/sysctl.conf ...
vm.vfs_cache_pressure = 500
vm.swappiness = 100
vm.dirty_background_ratio = 1
vm.dirty_ratio = 50

admin@raspibolt3:~/zram-swap $ sudo cat /proc/swaps
Filename				Type		Size		Used		Priority
/var/swap                               file		2097148		0		-2
/dev/zram0                              partition	10485756	256		15
```

The following command works:

```sh
$ sudo systemctl restart zram-swap
```

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+
+## Extras
+
+### Remote access over Tor (optional)
+
+To use your Fulcrum server when you're on the go, you can easily create a Tor hidden service.
+This way, you can connect the BitBoxApp or Electrum wallet also remotely, or even share the connection details with friends and family. Note that the remote device needs to have Tor installed as well.
+
+* Ensure that you are logged with user "admin" and add the following three lines in the section for "location-hidden services" in the torrc file. Save a exit
```
Stadicus, line 399: s/Save a exit/Save and exit./

guide/bonus/bitcoin/Fulcrum.md
```diff
@@ -0,0 +1,547 @@
...
+
+* As “admin” user, stop the Fulcrum service
+
+```sh
+$ sudo systemctl stop fulcrum
+```
+
+* Download, verify and install the latest Fulcrum binaries as described in the [Fulcrum section](fulcrum.md#download-and-set-up-fulcrum) of this guide.
+
+## After installation
```
Stadicus, line 463: This title is confusing, and implies that the following steps must be done after installation. 

I'd remove it, and call this section simply "Uninstall"

guide/bonus/bitcoin/Fulcrum.md
twofaktor: Do you suggest changing `$ sudo sysctl --system` to `$ sudo systemctl restart zram-swap` in line 96?

twofaktor: I found this in "install.sh" script of zram

guide/bonus/bitcoin/Fulcrum.md
twofaktor: Then, is you suggest changing `$ sudo sysctl --system` to `$ sudo systemctl restart zram-swap`??

twofaktor: Then, is you suggest changing `$ sudo sysctl --system` to `$ sudo systemctl restart zram-swap`??

guide/bonus/bitcoin/Fulcrum.md
Stadicus: No, not replacing, I just used the command in addition to the guide.

The install script `enables` the service. That means on restart it would be running, but not immediately. This confirms that the `restart` command should be added to the guide.

guide/bonus/bitcoin/Fulcrum.md
Stadicus: Ah, I misread the text, sorry. But your change makes is even more clear, without relying on Electrs being installed first.

guide/bonus/bitcoin/Fulcrum.md
Stadicus: No, it's an additional command.

Stadicus:
The Fulcrum config specified is pretty volatile... :) that makes it hard to test. I'm currently syncing with the previous config, which uses

* `db_use_fsync = true` --> now no longer explicitly enabled
* `db_max_open_files= 60` --> now `200`

Did you test this specific configuration on a 4GB RPi, @DanielPX1?

twofaktor:
> The Fulcrum config specified is pretty volatile... :) that makes it hard to test. I'm currently syncing with the previous config, which uses
> 
> * `db_use_fsync = true` --> now no longer enabled (default: false)
> * `db_max_open_files= 60` --> now `200`
> 
> Any special reason to disable `fsync`, @DanielPX1? It seems like a good option for low-resource systems using ext4 fs:
> 
> ```
> # RocksDB use "fsync" - 'db_use_fsync' - DEFAULT: false
> #
> # This boolean value, if true, tells the rocksdb library to use fsync() calls as
> # opposed to fdatasync(). fsync() is slower but works around some potential
> # issues in older Linux kernels and the ext4 filesystem.
> #
> # From the rocksdb documentation:
> #
> #  " By default, writes to stable storage use fdatasync (on platforms where this
> #    function is available). If this option is true, fsync is used instead.
> #
> #    fsync and fdatasync are equally safe for our purposes and fdatasync is
> #    faster, so it is rarely necessary to set this option. It is provided as a
> #    workaround for kernel/filesystem bugs, such as one that affected datasync
> #    with ext4 in kernel versions prior to 3.7. " - rocksdb/options.h
> #
> # db_use_fsync = false
> ```
> 
> https://github.com/cculianu/Fulcrum/blob/master/doc/fulcrum-example-config.conf#L666
> 
> Did you test this specific configuration on a 4GB RPi?

I caught [Raspiblitz's references](https://github.com/rootzoll/raspiblitz/blob/b20293a0c90f531477774c9a810b5491c81d0166/home.admin/config.scripts/bonus.fulcrum.sh) script to changes of values and other things, and the before comment "db_use_fsync = false" --> "so it is rarely necessary to set this option" made me believes that it wasn't necessary.

but anyway if you think it's necessary to keep both, let me know and I put it, no problem.

DanielPX1:
> > The Fulcrum config specified is pretty volatile... :) that makes it hard to test. I'm currently syncing with the previous config, which uses
> > 
> > * `db_use_fsync = true` --> now no longer enabled (default: false)
> > * `db_max_open_files= 60` --> now `200`
> > 
> > Any special reason to disable `fsync`, @DanielPX1? It seems like a good option for low-resource systems using ext4 fs:
> > ```
> > # RocksDB use "fsync" - 'db_use_fsync' - DEFAULT: false
> > #
> > # This boolean value, if true, tells the rocksdb library to use fsync() calls as
> > # opposed to fdatasync(). fsync() is slower but works around some potential
> > # issues in older Linux kernels and the ext4 filesystem.
> > #
> > # From the rocksdb documentation:
> > #
> > #  " By default, writes to stable storage use fdatasync (on platforms where this
> > #    function is available). If this option is true, fsync is used instead.
> > #
> > #    fsync and fdatasync are equally safe for our purposes and fdatasync is
> > #    faster, so it is rarely necessary to set this option. It is provided as a
> > #    workaround for kernel/filesystem bugs, such as one that affected datasync
> > #    with ext4 in kernel versions prior to 3.7. " - rocksdb/options.h
> > #
> > # db_use_fsync = false
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://github.com/cculianu/Fulcrum/blob/master/doc/fulcrum-example-config.conf#L666
> > Did you test this specific configuration on a 4GB RPi?
> 
> I caught a new [Raspiblitz's references](https://github.com/rootzoll/raspiblitz/blob/b20293a0c90f531477774c9a810b5491c81d0166/home.admin/config.scripts/bonus.fulcrum.sh) script to changes of values and other things, and the before comment "db_use_fsync = false" --> "so it is rarely necessary to set this option" made me believes that it wasn't necessary.
> 
> But anyway if you think it's necessary to keep both at the above values, let me know and we put it, no problem.

tbh I am not sure how big of difference it makes... Worked for me being off every time so far and as @twofaktor stated, it appeared unneccesary to have it turned on

Stadicus:
Unfortunately, my 4GB Pi crashed again at 65%, this seems to be the exact point-in-time when the sync becomes too much for it to handle... 

I don't want to be the blocker here, and maybe it's just my Pi. But I'm personally not convinced that the current 4GB config will work well in the wild.

DanielPX1:
> Unfortunately, my 4GB Pi crashed again at 65%, this seems to be the exact point-in-time when the sync becomes too much for it to handle...
> 
> I don't want to be the blocker here, and maybe it's just my Pi. But I'm personally not convinced that the current 4GB config will work well in the wild.

@Stadicus can you please try the setting with 10GB zram and let me know? I think it should not make these issues for you raspi

Stadicus:
> try the setting with 10GB zram and let me know?

I tried that before, to no avail. Until today, I was not able to sync Fulcrum on my 4GB Pi, it crashed at 85% the latest. But That does not need to be a blocker if it works for your and there ain't any issues with 8GB Pis.

Stadicus:
If you see issues popping up with 4GB Pis going forward, maybe we can add a disclaimer/warning at the beginning.

Stadicus (approved):
tACK, good to merge! :)

Caveat: I was not able to sync on my 4GB Pi, but others didn't have this issue.

twofaktor:
Did someone try this section?

https://raspibolt.org/guide/bonus/bitcoin/fulcrum.html#enable-fulcrum-api-connection-to-btc-rpc-explorer

PR #1036, opened 2022-06-08
Ignore RC pre-releases for RTL

kristapsk:
#### What

RTL has tags also for RC (release candidate) pre-releases, you want to ignore them when checking for latest version.

Currently, without this change, will think v0.12.4-rc1 is the latest, after - v0.12.3.

#### How

Ignore git tags that contain "rc".

#### Scope

- [ ] independent bonus guide
- [X] simple bug fix

#### Test & maintenance

Just check before / after bash oneliner result.

### DIFF
--- a/guide/lightning/web-app.md
+++ b/guide/lightning/web-app.md
@@ -287,6 +287,7 @@ Make sure to read the release notes first.
   ```sh
   $ cd /home/rtl/RTL
   $ git fetch
+  $ git reset --hard
   $ git tag | grep -E "v[0-9]+.[0-9]+.[0-9]+$" | sort --version-sort | tail -n 1
   $ git checkout v0.12.3
   $ git verify-tag v0.12.3
--- a/guide/lightning/web-app.md
+++ b/guide/lightning/web-app.md
@@ -102,12 +102,12 @@ We do not want to run Ride the Lightning alongside bitcoind and lnd because of s
   $ git clone https://github.com/Ride-The-Lightning/RTL.git
   $ cd RTL
 
-  $ git tag | sort --version-sort | tail -n 1
-  > v0.12.2
+  $ git tag | grep -E "v[0-9]+.[0-9]+.[0-9]+$" | sort --version-sort | tail -n 1
+  > v0.12.3
 
-  $ git checkout v0.12.2
+  $ git checkout v0.12.3
 
-  $ git verify-tag v0.12.2
+  $ git verify-tag v0.12.3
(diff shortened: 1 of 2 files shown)


mktech-gh (commented):
I replaced my old script with the new script on my Raspibolt: The RTL version v0.12.3 was still displayed in red. I deleted the .raspibolt.version file. Now it works: The RTL version v0.12.3. is displayed in green.

kristapsk:
Anyone has more comments or plan to test this?

VajraOfIndra:
> Anyone has more comments or plan to test this?

I tried to test it a few days ago but I encountered an issue while installing RTL so can't test it until my issue is resolved...

However, as @mktech-gh tested it, I could approve and merge it now if that's ok with you?

Also, what do you think of using your command in the main guide as well? as there is the same issue there.

pleb4868:
updated the welcome script according to @kristapsk change and can confirm it works, i.e RTL is shown as updated

guide/lightning/web-app.md
```diff
@@ -282,14 +282,14 @@ Make sure to read the release notes first.
...
 
   ```sh
   $ cd /home/rtl/RTL
   $ git fetch
-  $ git tag | sort --version-sort | tail -n 1
-  $ git checkout v0.12.2
-  $ git verify-tag v0.12.2
+  $ git tag | grep -E "v[0-9]+.[0-9]+.[0-9]+$" | sort --version-sort | tail -n 1
+  $ git checkout v0.12.3
+  $ git verify-tag v0.12.3
```
Stadicus, line 292: Following the upgrade steps, I get an error message because some versioned files have changed. This has nothing to do with the proposed changes, but the fix is simple and the following line could be added after the `git fetch` command`:

```sh
$ git reset --hard
```

Full context:
```sh
$ git checkout v0.12.3
error: Your local changes to the following files would be overwritten by checkout:
	package-lock.json
Please commit your changes or stash them before you switch branches.
Aborting

$ git reset --hard
HEAD is now at cf66708d Release 0.11.2 (#786)

$ git checkout v0.12.3
Updating files: 100% (1291/1291), done.
Previous HEAD position was cf66708d Release 0.11.2 (#786)
HEAD is now at f84719e2 Removed test script from build
```

Stadicus (approved):
utACK, looks good! :)

guide/lightning/web-app.md
kristapsk: Now remember that I also have noticed this in the past. Will add extra commit. My solution was to do `git stash`, but `git reset --hard` seems better.

Stadicus:
Awesome, thanks a log! :tada:

PR #1039, opened 2022-06-18
Section for macOS users on remote SSH through Tor

sutterseba:
#### What

Added a seperate section for macOS Users on how to setup remote SSH through Tor.

#### Why

Using `torify` or `torsocks` as suggested will cause issues with _System Integrity Protection_.

```text
ERROR: /usr/local/bin/ssh is located in a directory protected by Apple's System Integrity Protection.
```

Simply turning SIP off is dangerous and not recommended, even though it's probably the fastest solution to this.

#### How

The solution below simply adds a SSH config file with a proxy command. Once set up, one can conveniently SSH to the Pi with `ssh HOSTNICKNAME`.

#### Scope

- [x] significant change to core configuration (?)
- [ ] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

Tested on macOS 12.4.

I am unsure from what version of macOS onwards this issue occurs, but some users on StackExchange have suggested 10.11 i.e. "El Capitan" which should affect almost every macOS user.

### DIFF
--- a/guide/raspberry-pi/privacy.md
+++ b/guide/raspberry-pi/privacy.md
@@ -132,11 +132,12 @@ A few examples:
   ```sh
   $ torify ssh admin@abcdefg..............xyz.onion
   ```
+
   ```sh
   $ torsocks ssh admin@abcdefg..............xyz.onion
   ```
 
-  * **macOS**: Using `torify` or `torsocks` may not work due to Apple's *System Integrity Protection (SIP)* which will deny access to `/usr/bin/ssh`.
+* **macOS**: Using `torify` or `torsocks` may not work due to Apple's *System Integrity Protection (SIP)* which will deny access to `/usr/bin/ssh`.
 
   To work around this, first make sure Tor is installed and running on your Mac:
 
@@ -150,7 +151,7 @@ A few examples:
   $ ssh -o "ProxyCommand nc -X 5 -x 127.0.0.1:9050 %h %p" admin@abcdefg..............xyz.onion
   ```
 
-  For a more permanent solution, add these six lines below to your local SSH config file. Choose any Hostnickname you want, save and exit.
+  For a more permanent solution, add these six lines below to your local SSH config file. Choose any HOSTNICKNAME you want, save and exit.
 
   ```sh
   $ sudo nano .ssh/config
--- a/guide/raspberry-pi/privacy.md
+++ b/guide/raspberry-pi/privacy.md
@@ -144,15 +144,20 @@ A few examples:
   $ brew install tor && brew services start tor
   ```
 
-  Add the six lines below to your local SSH config file. Choose any Hostnickname you want, save and exit.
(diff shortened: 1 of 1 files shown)


DanielPX1:
Great idea, had issues myself in the past

DanielPX1:
What came into my mind as well - if you dont want to set up this config, you can use this command. It will create the config automatically afterwards. At least did for me and connected me to my node as well - using Tor installed via Brew. I guess you can use it with Tor browser running on background, but with a different port 

$ brew services start tor 
$ ssh -o "ProxyCommand nc -X 5 -x 127.0.0.1:9050 %h %p" admin@xyz.onion

sutterseba:
Works great! It doesn't create a config file on it's own though, at least for me. Am I missing something? Would be great if we can get it to work.

Adding this as an alias would probably require similar/more steps than my current solution.

DanielPX1:
> Works great! It doesn't create a config file on it's own though, at least for me. Am I missing something? Would be great if we can get it to work.
> 
> Adding this as an alias would require the same amount of steps as my current solution but reducing it down to that single line would be neat.

Ok i am retarded, you are right. I thought it did for some reason and it did not. Will have to add them manually like you, however this is a nice shortcut, maybe it can be mentioned in your version? Just like a alternative. Also setting some shortcut for it would be nice

sutterseba:
> maybe it can be mentioned in your version

Done! 👍

guide/raspberry-pi/privacy.md
```diff
@@ -136,6 +136,47 @@ A few examples:
   $ torsocks ssh admin@abcdefg..............xyz.onion
   ```
 
+  * **macOS**: Using `torify` or `torsocks` may not work due to Apple's *System Integrity Protection (SIP)* which will deny access to `/usr/bin/ssh`.
```
Stadicus, line 139: The space before the `*` bullet indents the whole section and should be removed

guide/raspberry-pi/privacy.md
```diff
@@ -136,6 +136,47 @@ A few examples:
...
+  $ brew install tor && brew services start tor
+  ```
+
+  You can SSH to your Pi "out of the box" with the following proxy command:
+
+  ```sh
+  $ ssh -o "ProxyCommand nc -X 5 -x 127.0.0.1:9050 %h %p" admin@abcdefg..............xyz.onion
+  ```
+
+  For a more permanent solution, add these six lines below to your local SSH config file. Choose any Hostnickname you want, save and exit.
```
Stadicus, line 153: The capitalized "Hostnickname" looks a bit weird. How about using "HOSTNICKNAME" instead, which also improves consistency with the code block below?

Stadicus (approved):
Commented on a few nits, otherwise this LGTM! :)

sutterseba:
Fixed it :)

VajraOfIndra:
utACK, LGTM

Thanks for this nice update @sutterseba :+1: :)

Let's go to master!

Issue #1038, opened 2022-06-13
[BUG] permission denied when running command mkdir /data/bitcoin


artizzle:
After following the Bonus Guide: Boot from microSD instead of external drive, and reaching the Bitcoin Client guide and when trying to run the command "mkdir /data/bitcoin" I got the permission denied error. This happens because the owner of the /data directory is root , not admin. This only happens when you follow the Bonus Guide: Boot from microSD instead of external drive. When you follow the regular guide this problem doesn't happen because in the regular guide (Raspberry Pi --> System Configuration --> Data Directory) there is the command sudo chown admin:admin /data which is run right after making the /data directory. sudo chown admin:admin /data is not in the Bonus Guide: Boot from microSD instead of external drive.

My suggestion is to add this command (sudo chown admin:admin /data) some where in the Bonus Guide: Boot from microSD instead of external drive.

PR #1045, opened 2022-06-30
Fixed permission issue in microSD bonus guide

sutterseba:
#### What

Added `chown admin:admin /data` which is included in the default configuration guide, but was missing in the _Boot from microSD_ bonus guide resulting in permission denied when trying to create the bitcoin directory.

`mkdir: cannot create directory '/data/bitcoin': Permission denied`

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

Fixes https://github.com/raspibolt/raspibolt/issues/1038


VajraOfIndra (approved):
utACK, LGTM...!

Issue #957, opened 2022-03-07
[FEATURE REQ] Use new RPi Imager feature to create RaspiOS image with user "admin" directly


twofaktor:
### Describe
Create directly "admin" user instead "pi" user and set the password [A], include "Set locale settings" in “Advanced options” of Raspberry Pi Imager v1.7.1 and modify steps according to the correct instructions.

### Location in guide
https://raspibolt.org/guide/raspberry-pi/operating-system.html#configure-boot-options

Change the screenshot to this:
https://drive.google.com/file/d/1ZAwPeOmUaUnK3WtKEYxei34bMaPbfG4E/view?usp=sharing

### Additional context

- Modify: https://raspibolt.org/guide/raspberry-pi/remote-access.html#access-with-secure-shell

--> Change "pi" user to "admin" user login

- Modify https://raspibolt.org/guide/raspberry-pi/system-configuration.html#add-the-admin-user-and-log-in-with-it. 
--> Put first, steps:
- Create a new connection with the admin user
- Log in again using SSH (see [Access with Secure Shell](https://raspibolt.org/guide/raspberry-pi/remote-access.html#access-with-secure-shell) section), but now with the user “admin” and your password [A]
--> Modify title according to steps
--> Delete step: "Create a new user called “admin” with your password [A]"
--> Delete step: "Exit your current “pi” user session and exit SSH"

Stadicus:
Thanks, I think that would be a nice optimization of the core guide, simplifying the setup process. I'll reopen this issue to keep that in mind if someone wants to implement this.

visskiss:
Only issue with this is that for future upgrades or system restores, you will have different numbered users for all users, so that if you attempt to use rsync and retain owner and group, they will refer to the incorrect user. This is because users are numbers, not names. User admin will restore as bitcoin, user electrs will restore as bitcoin, etc...

VajraOfIndra:
This is still to be done and an important update to simplify the guide. Re-opening.

VajraOfIndra:
TBD

twofaktor:
Abandoned but there is an implementation thread on #1050. Proceed to close this problem

PR #1050, opened 2022-07-15
Setup user admin directly in Raspberry Pi Imager

sutterseba:
### What

Added instructions to create user `admin` directly in the boot configuration of the Raspberry Pi Imager v1.7.2 and removed/edited steps during the setup guides accordingly.

### Why

Setting up the admin user directly from boot option configuration in the Imager simplifies the setup process.

### How

* Changed Screenshot in [Configure boot options](https://raspibolt.org/guide/raspberry-pi/operating-system.html#configure-boot-options) to match the new instructions (Set username and password, Set locale settings).

   To match version number (in title bar) and window UI, I changed the other screenshot as well. 

* Changed username from `pi` to `admin` in [Access with Secure Shell](https://raspibolt.org/guide/raspberry-pi/remote-access.html#access-with-secure-shell).

* Removed the (now) unneeded steps to create the admin user manually and changed title + description accordingly in [Add the admin user (and log in with it)](https://raspibolt.org/guide/raspberry-pi/system-configuration.html#add-the-admin-user-and-log-in-with-it).

### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

Implements Feature Request https://github.com/raspibolt/raspibolt/issues/957

### Testing

Note: I have **not** taken this into account:

>Only issue with this is that for future upgrades or system restores, you will have different numbered users for all users, so that if you attempt to use rsync and retain owner and group, they will refer to the incorrect user. This is because users are numbers, not names. 

_@visskiss in https://github.com/raspibolt/raspibolt/issues/957#issuecomment-1062457020_


kristapsk:
Not everyone will use Raspberry Pi Imager, for example, when using this guide with other hardware, but we could assume they are advanced enough users to know how to add / modify users themselves.

twofaktor:
Hello guys, what happened with this PR? 

This is Raspibolt and this guide for Raspberry Pi using its Raspbian official OS. We can take advantage of the Raspberry Pi imager and its special features for preconfiguring Raspbian OS.
In another dedicated guide for other architectures we can consider other types of more universal steps, but for this use case, I don't think it's necessary to rethink anything else.

WDYT?

VajraOfIndra (approved):
Raspberry Pi Imager section: tACK
Rest: utACK (I didn't do a full test on a fresh new node)

LGTM

Really happy to see this upgrade, it makes the guide even more straightforward.

VajraOfIndra:
> Note: I have **not** taken this into account:
> 
> > Only issue with this is that for future upgrades or system restores, you will have different numbered users for all users, so that if you attempt to use rsync and retain owner and group, they will refer to the incorrect user. This is because users are numbers, not names.

I don't understand the initial comment and concern. Is it really an issue..? @Stadicus @kristapsk?

kristapsk:
> > Note: I have **not** taken this into account:
> > > Only issue with this is that for future upgrades or system restores, you will have different numbered users for all users, so that if you attempt to use rsync and retain owner and group, they will refer to the incorrect user. This is because users are numbers, not names.
> 
> I don't understand the initial comment and concern. Is it really an issue..? @Stadicus @kristapsk?

Idea is that in filesystem level uid / gid numbers are stored not usernames / passwords. If person installed RaspiBolt with old guide and then reinstall with the new one and just reuses old data files, uid / gid numbers will be wrong. Not sure how much this is an issue, you already will have different uid / gid numbers when installing various additional tools in different order.

P.S. There is no issue with `rsync`, it maps owners and groups by name not id by default, unless you specify `--numeric-ids` parameter.

VajraOfIndra:
> > > Note: I have **not** taken this into account:
> > > > Only issue with this is that for future upgrades or system restores, you will have different numbered users for all users, so that if you attempt to use rsync and retain owner and group, they will refer to the incorrect user. This is because users are numbers, not names.
> > 
> > 
> > I don't understand the initial comment and concern. Is it really an issue..? @Stadicus @kristapsk?
> 
> Idea is that in filesystem level uid / gid numbers are stored not usernames / passwords. If person installed RaspiBolt with old guide and then reinstall with the new one and just reuses old data files, uid / gid numbers will be wrong. Not sure how much this is an issue, you already will have different uid / gid numbers when installing various additional tools in different order.
> 
> P.S. There is no issue with `rsync`, it maps owners and groups by name not id by default, unless you specify `--numeric-ids` parameter.

Thanks for the explanation. So it's not really an issue, only for a very small subsets of users (if any..).

Let's merge it then :+1:

PR #1055, opened 2022-08-02
[UPDATE BONUS GUIDE] Aliases bonus guide

twofaktor:
#### What

Improve of the aliases bonus guide simplifying installation, adding new aliases to the list and fixing typos / wrong information

#### Why

This can lead to significant improvements to the Raspibolt node runners.

#### How

- Fixed typos and other wrong information
- Simplied the installation process
- Pulled aliases list in a separate file inside "resources" folder, useful for future updates and cleary the guide
- Added new useful aliases commands tools to the list
- Changed head image according to the new improvements

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

The changes were tested by myself. I will be updating the list of commands as new functions are added in the main guide / bonus guide

#### Animated GIF

giphy


resources/.bash_aliases
```diff
@@ -0,0 +1,254 @@
...
+# /resources/.bash_aliases
+
+###########################
+# GENERAL STATUS & OTHERS #
+###########################
+
+# SYSTEM
+alias update='sudo apt update'
+alias listupgradable='sudo apt list --upgradable'
+alias upgrade='sudo apt upgrade'
```
VajraOfIndra, line 11:
```suggestion
alias upgrade='sudo apt dist-upgrade'
```
I was told this was the recommended command to upgrade. Wdyt?

resources/.bash_aliases
```diff
@@ -0,0 +1,254 @@
...
+  lightning-cli --version ; \
+  Fulcrum --version | grep Fulcrum'
+
+alias fail2banreport='sudo fail2ban-client status sshd'
+alias overview='raspibolt'
+alias testscb-backup='sudo touch /data/lnd/data/chain/bitcoin/mainnet/channel.backup'
+
+# EXTRA LOGS
+alias authlogs='sudo tail -f /var/log/auth.log'
+alias ufwlogs='sudo tail -f /var/log/ufw.log'
```
VajraOfIndra, line 43: I got an error message with that one. Turns out UFW logging does not seem to be enabled by default (mine were OFF after checking with `sudo ufw status verbose`.

Do you think we should activate the logs after installing UFW in the main guide?

VajraOfIndra (approved):
LGTM! Nice job

I tested all the aliases that wouldn't mess up with my services. 

I just have 2 small comments, see below.

The new download/check/use the file is better than the old copy/pasting way that was cumbersome :+1:

resources/.bash_aliases
twofaktor: Yeah, could be a good idea, only low log level (default level) 

https://www.linode.com/docs/guides/configure-firewall-with-ufw/#:~:text=Logging

Change `sudo ufw logging off` to `sudo ufw logging on` 

But anyway really i think the most important command to detect possible force brute attack is `authlog`

resources/.bash_aliases
twofaktor: apt upgrade will upgrade all packages that can be upgraded without the need to install additional packages or remove any conflicting installed packages. Basically it will apply all package upgrades that do not include changed dependencies.

apt full-upgrade / dist-upgrade (the correct equivalent for apt-get dist-upgrade) applies package upgrades as well if they require either the install of new packages or the removal of conflicting installed packages. Basically it will apply all package upgrades including those with changed dependencies.

On stable distro suites, when not adding new APT sources, one will face changed dependencies rarely, hence usually apt full-upgrade is not required or does not apply any additional upgrades compared to apt upgrade, except for Linux image package upgrades.

I think only `upgrade` has less risky for a system in production that full/dist upgrade. I would leave it like this

VajraOfIndra: Ok, sounds logical, let's keep as it is

resources/.bash_aliases
VajraOfIndra: ok, so a PR will be needed in the main guide if we want to add the UFW logging.

In the meantime, I think we can merge this PR as is.

VajraOfIndra:
🎉

PR #1060, opened 2022-08-10
[UPDATE BONUS GUIDE] Simpler mempool configuration

RooSoft:
#### What

Completely removed NGINX config, should allow for easier node restarts and mempool app upgrades

### Why

The app needed a refresh and NGINX configure had some potential DNS issues on node restarts

#### How

Run the frontend according to the [official documentation](https://github.com/mempool/mempool/tree/master/frontend)

#### Scope

- [X] guide simplification

#### Test & maintenance

Restart the raspibolt and make sure the mempool site works at https://raspibolt:4200

#### Animated GIF (optional)

blabla


RooSoft:
The frontend should be using the backend provided in the same guilde... at the moment, I can't guarantee it's the case. Can someone inspect the network to make sure it's not calling the public mempool API?

sig-noi (changes requested):
latest version appears to be 2.4.0 - should line 90 be updated? 
`$ git checkout v2.4.0`

RooSoft:
@sig-noi the code literally gets the latest version

```bash
git checkout $latestrelease
```

guide/bonus/bitcoin/mempool.md
```diff
@@ -490,7 +375,7 @@ To expose Mempool app via a Tor hidden service (if only Tor address is used, no
   # Mempool Hidden Service
   HiddenServiceDir /var/lib/tor/hidden_service_mempool
   HiddenServiceVersion 3
-  HiddenServicePort 443 127.0.0.1:4081
```
twofaktor, line 493: The correct line following your modifications is: `HiddenServicePort 80 127.0.0.1:4081`

Tor network has encrypted communications by design, this would be redundant

guide/bonus/bitcoin/mempool.md
```diff
@@ -490,7 +375,7 @@ To expose Mempool app via a Tor hidden service (if only Tor address is used, no
   # Mempool Hidden Service
```
twofaktor, line 375: Add in the line before:
`############### This section is just for location-hidden services ###`

ref e.g: https://raspibolt.org/guide/bitcoin/blockchain-explorer.html#remote-access-over-tor-optional

to standardize with the rest of the guide

twofaktor (approved):
Few suggestions

twofaktor (changes requested):
Add: `"USE_ONION": true,`
Between: 195-196

VajraOfIndra:
> Add: `"USE_ONION": true,` Between: 195-196

This has already been merged with #1121

VajraOfIndra:
I don't think removing the nginx reverse proxy is desired as this is a core security feature of our guides. 

I didn't seem to have an issue with Mempool lately so maybe things have changed with the latest verisons of Mempool. Otherwise, I think working with the various service files (bitcoind, mempool, nginx) and playing with various options (Wants, PartOf etc) might solve any remaining issues.

I propose we close this PR for now.

twofaktor:
> I don't think removing the nginx reverse proxy is desired as this is a core security feature of our guides.
> 
> I didn't seem to have an issue with Mempool lately so maybe things have changed with the latest verisons of Mempool. Otherwise, I think working with the various service files (bitcoind, mempool, nginx) and playing with various options (Wants, PartOf etc) might solve any remaining issues.
> 
> I propose we close this PR for now.

- What about creating frontend autoboot? 
- We don't use `rpcuser/rpcpassword` auth on `bitcoin.conf``, instead we use rpcauth (cookie file), was tested this guide on RaspiBolt v3? 
```
  "CORE_RPC": {
    "HOST": "127.0.0.1",
    "PORT": 8332,
    "USERNAME": "raspibolt",
    "PASSWORD": "Password[B]"
```
- What about using an SSL connection through Tor? We know Tor is encrypted by design. I'm wrong if I think that with this configuration mempool is listening directly on the SSL port.

PR #1066, opened 2022-08-20
Upgrading Bitcoin CoreRaspibolt3

Ralph-Naders-ghost:
#### What

Possible failure mode as written; LND may not re-connect to bitcoind after restart

## For the future: upgrade Bitcoin Core
...
...
...
...

* Download, verify, extract and install the Bitcoin Core binaries as described in the [Bitcoin section](bitcoin-client.md#installation) of this guide.

* Restart the Bitcoin Core systemd unit

  ```sh
  $ sudo systemctl restart bitcoind
  ```

<br /><br />

ADD:

LND may not reconnect to bitcoind.  Check this with 
 ```sh
$ sudo journalctl -f -u lnd
  ```

If it shows any errors, LND may need to be restarted with
```sh
$ sudo systemctl restart lnd
```


twofaktor:
Thanks for your contribution! 

1º For this PR shouldn't be necessary to change 49 files - you must create a new branch and modify only bitcoin-client.md file: https://github.com/raspibolt/raspibolt/blob/master/guide/bitcoin/bitcoin-client.md

BTW: You should fork the original master branch, not another forked project.
https://github.com/raspibolt/raspibolt
https://github.com/VajraOfIndra/RaspiBolt

2º We are all supposed to know that when the BTC Core is restarted the rest of the services that depend on it must be configured in the code to wait for the BTC Core to be available again. Otherwise, the rest of the services must be restarted manually. Maybe we should put this notice in another more general section, not in every service

I think that this PR is not correctly opened

VajraOfIndra:
As requested by OP, closing this PR for the time being. OP will create a new one from a forked original RaspiBolt branch later on.

PR #1072, opened 2022-08-26
New bonus guide: Safu Ninja, a DIY steel mnemonics backup solution

VajraOfIndra:
#### What

- A guide on how to make a cool DIY steel backup for mnemonics with washers and a stamping set.

### Why

- This off-the-shelf DIY backup solution so a perfect match with the whole RaspiBolt ethos. It's not _directly_ related to the node but very close to it so i think it has its place as a bonus guide. Wdyt?

#### How

- Wrote a brand new guide but gave it the name Safu Ninja and acknowlegde them as the inventor of the technique (their Reddit post of June 2020 is the oldest reference to the tehcnique I could find and also Jameson Lopp in his stress test page refers to this technique under the Safu Ninja name).

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

Fixes non-resilient paper storage!

#### Test & maintenance

Review instructions and links.

#### Ideas for improvements (before merging or later on)

- Improve technical recommendations for washers, bolt and stamping set selection. The sizes of them all is critical and the choice is so large and a bit confusing. There might be some differences between US vs Europe vs other parts of the world etc.. So maybe adding a picture of a table with washer size could help.
- Add images to illustrate the process

#### Animated GIF (optional)

This is fine


guide/bonus/bitcoin/safu-ninja.md
```diff
@@ -0,0 +1,152 @@
...
+Instead, it is more private and ultimately secure to build this steel mnemonics backup from off-the-shelf components that have not be designed for Bitcoin usage. Like you did with the RaspiBolt node vs buying a plug-and-play node!
+
+For example, a popular DIY method relies on standard steel washers stacked and locked on a bolt. The mnemonic words and ordering numbers are punched on the washers using a metal stamp set. All the components can be found easily in a hardware shop or online and none are associated with Bitcoin usage. Simple and private.
+
+As far as I know, SafuNinja was the first one to come up with [this idea](https://www.reddit.com/r/Bitcoin/comments/h0j136/cheap_and_easy_diy_metal_bitcoin_seed_storage/){:target="_blank"} back in June 2020. You can check their [website](https://safu.ninja/){:target="_blank"}.
+
+
+This metal backup has been stress tested by Jameson Lopp. Check his full review [here](https://blog.lopp.net/metal-bitcoin-seed-storage-stress-test-iv/){:target="_blank"} and below is a summary:
+
+![Stress test](../../../images/safu-ninja-stress-test-result.png)
```
Stadicus, line 61: The actual image files name uses an UPPER CASE file extension, so this image is not displayed here.

guide/bonus/bitcoin/safu-ninja.md
```diff
@@ -0,0 +1,152 @@
...
+
+* Step 1: Punch the first word on a washer. It is better to punch the entire word, rather than just the first 4 letters. While the 4 first letters are sufficient to identify the unique word from the BIP 39 list of words... it removes some of the redundancy provided by the remaining letters. If the washer is damaged and the first two letters are not readable, you'll have more chance to identify the word if all the remaining letters of the word are present than if only the following 2 letters are present.
+
+* Step 2: Punch the number corresponding to the position of the word on the list. This will help in case the washers get loose and mixed.
+
+* Repeat step 1 and 2 above for all the remaining words
+
+* (Optional) Punch the name of the wallet on a washer (e.g. "MY COLDCARD" or "LND"). It might come handy in a few years to remember what wallet was associated with this backup and also because some wallets might use uncommon derivation paths. Check this [list of derivation paths](https://walletsrecovery.org/){:target="_blank"} supported by each wallet is maintained by @NVK and @J9Roem.
```
Stadicus, line 100: I think we should avoid mentioning specific brands if not really relevant. I also try to keep the BitBox out of it so that it's a neutral guide. 

How about this?
`s/MY COLDCARD/WALLET XY/`

guide/bonus/bitcoin/safu-ninja.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Recovery
+
+Below are a few pointers in case you need to use your backup to recover funds from one of your wallet.
+
+* Open Sparrow
+
+* Create a new wallet (`Ctrl`+`n`)
+
+* Enter a wallet name (e.g., "Recovered Seedsigner")
```
Stadicus, line 120: Same here, the wallet name doesn't add any real value

`s/Recovered Seedsigner/Recovered Wallet XY/`

guide/bonus/bitcoin/safu-ninja.md
```diff
@@ -0,0 +1,152 @@
...
+
+* Screw and tighten the wingnut
+
+You're set! You now have a DIY steel metal backup that will be resistant to fire, flood, schocks and other kind of environmental damages.
+
+---
+
+## Recovery
+
+Below are a few pointers in case you need to use your backup to recover funds from one of your wallet.
```
Stadicus, line 114: In general, I would refrain from advising people to restore their recovery words into a hot software wallet, unless it's an emergency. All cold storage security is immediately lost and funds are at risk of remote attacks.

Best practice is to NEVER enter your 24 recovery words on an online device unless it's just a few hundred bucks (which is unlikely if you go to great lengths to create the most private and secure steel backup)

The recovery process should STRONGLY advise using a hardware wallet. The recovery words can then be entered into the dedicated and secure offline device without putting them at risk.

Stadicus (changes requested):
Great guide, with nice pictures. I added a few nits about naming specific brands.

One important change we should consider: NEVER enter your recovery words on an online device unless it's an emergency. I think the guide should strongly advise recovery using a hardware wallet so that the cold storage security is not lost and the recovery words are not put at risk.

guide/bonus/bitcoin/safu-ninja.md
VajraOfIndra: Fixed

guide/bonus/bitcoin/safu-ninja.md
VajraOfIndra: Good point, fixed.

guide/bonus/bitcoin/safu-ninja.md
VajraOfIndra: Addressed in comment below

guide/bonus/bitcoin/safu-ninja.md
VajraOfIndra: Very true, didn't thunk of that. I reworded the Recovery section, PTAL.

guide/bonus/bitcoin/safu-ninja.md
```diff
@@ -0,0 +1,124 @@
...
+
+* Step 2: Punch the number corresponding to the position of the word on the list. This will help in case the washers get loose and mixed.
+
+* Repeat step 1 and 2 above for all the remaining words
+
+* (Optional) Punch the name of the wallet on a washer (e.g. "WALLET XY"). It might come handy in a few years to remember what wallet was associated with this backup and also because some wallets might use uncommon derivation paths. Check this [list of derivation paths](https://walletsrecovery.org/){:target="_blank"} supported by each wallet is maintained by @NVK and @J9Roem.
+
+* (Optional) Punch the date of creation of your wallet, known as the "birth date". This could help you quicken the search for your UTXOs when reconstructing the wallet in a few years. The wallet software used for recovery will not need to scan the entire blockhain but will start at the birth date.
```
sutterseba, line 102: Maybe you could add something to distinct between classic BIP-39 mnemonic phrases and the aezeed cipher scheme. Normally this wouldn't be that important, but most users reading this bonus guide will both have "normal" cold/hot storage and the on-chain wallet of their LND client, using aezeed.  

Users punching an aezeed cipher should know about limited recovery options and, for example, that the birth date is already encoded in the cipher (and punching it is not necessary).

VajraOfIndra: Good idea, I added a reference to the azeed cypher scheme and the contained birth date.

Stadicus (approved):
LGTM! :)

PR #1074, opened 2022-08-27
New Bonus guide: Samourai Dojo

DanielPX1:
#### What

[Samourai Dojo](https://docs.samourai.io/en/dojo) is the backing server for Samourai Wallet. It will allow samourai wallet users to connect to their own backend, removing trust to a third party servers.

[Samourai Whirlpool](https://docs.samourai.io/en/whirlpool) is a free and open source (FOSS), non custodial, chaumian CoinJoin platform. Currently used by samourai wallet and sparrow wallet.

### Why

Samourai offers plug and play hardware and own [operational system](https://ronindojo.io/) similiar to [Umbrel](https://umbrel.com/) - However, I have encountered errors for raspberry pi and while it can be easy solution for a samourai-only user, many others of us use different wallets, clients (lightning etc.) and want to have complete control over our nodes and software versions.
Dojo can be installed on raspberry pi and run separately in a DIY way, without restricting ourselves to special purpose OS or buying new hardware

Whirlpool is a great addition to raspibolt privacy section as an alternative to joinmarket for people using samourai wallet and perhaps sparrow, if connecting to own client will be available in the future.

Samourai tools are one of the most advanced , being more and more relevant as blockchain surveilance improves. I believe having these additions as a bonus part of raspibolt is a very good step forward. 

#### How

By testing on my own raspberry and using [First Time Install](https://docs.samourai.io/dojo/first-install) section as an inspiration, but it does not contain all neccessary information. Some changes, as for example firewall configuration have to be in place for a proper function.

#### Scope

Reviews:
- [ ] Review formatting and standardization.
- [ ] Review guide (proper function)

#### Test & maintenance

Following guide and reporting issues (permissions, errors etc.)


VajraOfIndra:
First, thanks so much for writing this thorough and complex guide :clap: 

I love the idea of adding the Samourai software stack to the RaspiBolt offering, as Samourai is an important player in the Bitcoin ecosystem. 

One thing I'm wondering about is the use of Docker (as I understand it it's kinda compulsory, right?). Usually, the RaspiBolt approach and ethos is DIY and it's the opposite of the Docker approach.. I don't know if the guide _never_ accepts Docker solutions or might in some cases? We'd need the vision and decision of @Stadicus on that one.

Looking forward to hear from the other maintainers and other Bolters on this question and on this guide! :)

DanielPX1:
will do a new merge

Issue #1043, opened 2022-06-28
[FEATURE REQ] Add OpenTimestamps client guide


kristapsk:
### What & why

Since [v0.15.0](https://github.com/raspibolt/raspibolt/pull/1042) LND timestamps the manifest file along with its signature with [OpenTimestamps](https://opentimestamps.org/). Would be good that RaspiBolt users could also check these before installation.

### How

See https://github.com/opentimestamps/opentimestamps-client for OTS client installation info.

VajraOfIndra:
> ### What & why
> 
> Since [v0.15.0](https://github.com/raspibolt/raspibolt/pull/1042) LND timestamps the manifest file along with its signature with [OpenTimestamps](https://opentimestamps.org/). Would be good that RaspiBolt users could also check these before installation.
> ### How
> 
> See https://github.com/opentimestamps/opentimestamps-client for OTS client installation info.

I had a look and it should be easy to add to the instructions.

Just out of curiosity, I tried to find an example of when this time stamp check would actually prevent something bad to run on my node but couldn't really find any use case.. could you ELI5 the main use case of such a timestamp check for node operators? (assuming the binary checksum has been checked and the signature of the checksum file has also been checked)

kristapsk:
> could you ELI5 the main use case of such a timestamp check for node operators? what bad scenario does it protect against?

You can check that signature was made at release date (or before that), not forged later with different binaries (for example, somebody could gain access to private keys used for signing after original release was made).

>  (assuming the binary checksum has been checked and the signature of the checksum file has also been checked)

Yes, this kinda is enough, but extra security checks never hurts. What if all Github accounts with commit access to RaspiBolt master have been compromised?

VajraOfIndra:
> > could you ELI5 the main use case of such a timestamp check for node operators? what bad scenario does it protect against?
> 
> You can check that signature was made at release date (or before that), not forged later with different binaries (for example, somebody could gain access to private keys used for signing after original release was made).
> 
> > (assuming the binary checksum has been checked and the signature of the checksum file has also been checked)
> 
> Yes, this kinda is enough, but extra security checks never hurts. What if all Github accounts with commit access to RaspiBolt master have been compromised?

Ok so an example would be a checksum and signature that works but the timestamps would not meaning that the release has been modified since its release, right?

Anyways, let's add it for sure! And also for Bitcoin Core. Do you want to have a go at the PR?

VajraOfIndra:
Reopening this issue?

[...]

PR #1084, opened 2022-09-02
Adds checksum timestamp check to Bitcoin Core installation steps

VajraOfIndra:
#### What

Starting with Bitcoin Core v0.21.1 (May 2021), the file that contains the binary checksums (CHECKSUMS) is timestamped using the Bitcoin blockchain and the [OpenTimestamps protocol](https://opentimestamps.org/).

A timestamp proof (CHECKSUMS.ots) can be downloaded from the repository and can be tested against the timestamped file (CHECKSUM) to see if it matches.

There are 2 options to do the check:

1. Using the OpenTimestamps website. By uploading the file and letting the website backend do the check.
2. Runing the OpenTimestamp client locally on a computer or node and doing the check locally. This requires a bitcoin node.

As Bitcoin Core is not running yet at this stage of the guide, only option 1 is available.

If this PR is accepted. A follow up PR could be made to add a similar timestamp check for LND. We would need to decide if we would then go for option 1 (easy but trusting the website) or option 2 (makes the guide longer to explain how to install the client on the node, but trust-minimized).

### Why

Don't trust, verify :)
See discussion thread #1043 for some rationales as to when a timestampe can be relevant to protect the user.

#### How

- Added a section in the Bitcoin client guide
- Added subsections in the Installation section, to make this section clearer (as it was starting to get long with all the various checks. The section is now divided in the following steps:
  * Preparations
  * Checksum check
  * Signature check
  * Timestamp check
  * Installation

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

Test the new instructions.

stamp


kristapsk:
Concept ACK. Could do more careful review after Baltic Honeybadger. :)

Checking OTS proofs using bitcoind cannot be done during initial install, but can be done when upgrading to newer version.

sig-noi (approved):
This reads well and works! 

Agree with adding instructions to install the OTS client to verify updates and LND. I think it may make the most sense to write it as a mini-guide between this one and the Electrum guide, sort of an interlude that can be done while waiting for the blockchain to download for first-timers, but independently accessible for veterans who need to install it (as referenced by other guides that refer to OTS verification). Not a bonus guide as presumably better security is worth it rather than a bonus. Just my two sats.

kristapsk (approved):
ACK. I think this is good enough for now, can be improved later with follow-up PRs.

Issue #747, opened 2021-07-30
Electrs guide is quite outdated + other suggestions


Kixunil:
There are several things out-of-date in the `electrs` guide and I have some other suggestion

- [x] While the guide suggests dynamically checking version using `curl | jq` (more on it later), the screen recording does **not* use that technique, which [confused at least one user](https://github.com/romanz/electrs/issues/448#issuecomment-889635368)
- [x] The guide uses `cookie` option which is deprecated and **will be removed in the upcoming release**, please consider using `cookie_file`, which is safer.
- [ ] Consider recommending [dynamic linking](https://github.com/romanz/electrs/blob/master/doc/usage.md#dynamic-linking), which is faster - I don't have exact numbers but feels like the compile time is about half with dynamic linking. This may be especially helpful for RPi users since RPi is less powerful
- [x] Bitcoind needs `whitelist=download@127.0.0.1`

Additional suggestions:

- [x] Recommend installing Rust from repository instead of manually. Thanks @xanoni for reminding!
- [x] Recommend getting @romanz GPG key (`gpg --recv-keys 15C8C3574AE4F1E25F3F35C587CAE5FA46917CBB` - don't trust, verify the fingerprint from multiple sources) and running `git verify-tag "electrsgit"` before compiling
- [x] Put double quotes around `$(curl ...)` and `electrsgit` to prevent github from sending garbage value (with spaces) that could be dangerous
- [x] Add `echo "$electrsgit"` and recommend checking from an independent source that this is the latest version of `electrs` to avoid downgrade attacks (@romanz how about adding latest electrs version to your Twitter bio?)
- [x] Consider using `db_dir` option instead of symlinking `electrs` directory
- [x] Suggest upgrading by only backing up electrs binary, running `git clean -x -f d` followed by `git fetch` and `git checkout` - this is faster because only changes need to be downloaded
- [ ] Don't double-encrypt traffic via onion+ssl, use onion only; suggest skipping nginx+TLS if not needed.

Things for the future (unsupported by `electrs` but wanted):

- [ ] when `electrs` can skip default configs and `db_dir` is used add `ProtectHome`
- [x] when electrs has `--version` flag run it instead of parsing `RELEASE-NOTES.md`

xanoni:
Please also note that it's not required to build/install your own custom Rust version ... the one from Raspbian is fine ...

See here: https://github.com/romanz/electrs/issues/207#issuecomment-881924630

Kixunil:
[...]

PR #1102, opened 2022-09-24
Several Electrs upgrades based on #747 + a few bug fixes

VajraOfIndra:
#### What

This PR aims at addressing several improvement suggestions made by @Kixunil in #747:

- Use **dynamic linking** when compiling Electrs. The advantages of dynamic linking seem to be better than the ones for static linking, [see details here](https://github.com/romanz/electrs/blob/master/doc/install.md)
- Use `install` rather than `cp` -> More explicit
- Changed Tor hidden service port to 50001: Avoid un-necessary double encryption.
- Adding `git clean -x -f d` to the update instructions -> Removes untracked files from the working tree [(source)](https://git-scm.com/docs/git-clean)
- Harden systemd service by adding `ProtectSystem=full`
- Harden systemd service by adding `ProtectHome` option. To do this, I had to add the `--skip-default-conf-files `argument to the `ExecStart` command of the service file. I also add to add `cookie_file= "/data/bitcoin/.cookie"` to the Electrs conf file as otherwise I would get the following permissions error:
> Sep 24 10:00:23 raspibolt electrs[1173697]:     0: failed to open bitcoind cookie file: /home/bitcoin/.bitcoin/.cookie
> Sep 24 10:00:23 raspibolt electrs[1173697]:     1: Permission denied (os error 13)

@Kixunil Is that (the point above) the correct way to do it?
- Prepending `cargo build` command with `CARGO_NET_GIT_FETCH_WITH_CLI=true` which will workaround the bug on Debian systems with broken cargo
- Adding `git verify-tag` upgrade instructions to verify the release
- Changing "# Start electrs" in upgrade instructions to more accurate "# Restart electrs"

Several suggestions in #747 are not addressed in this PR and should be discussed in #747 (I'll list them there).

The PR also:
- Fixes a typo in the intro: removing extra text #prepare-nginx-reverse-proxy
- Fixes the language of the Electrs conf file from ini to toml
- As /home is protected with `ProtectHome`, replaced path to Bitcoin core daemon from "/home/bitcoin/.bitcoin" to "/data/bitcoin"

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

Install Electrs from scratch using the new dynamic linking, and test that the systemd service file runs fine.

todo


twofaktor:
Take the opportunity to put in line 284/291 the name of the hidden service:

`# Hidden Service Electrs SSL`

guide/bitcoin/electrum-server.md
```diff
@@ -36,14 +36,14 @@ Desktop wallets like [Sparrow](https://sparrowwallet.com/){:target="_blank"}, th
...
-Make sure that you have [reduced the database cache of Bitcoin Core](bitcoin-client.md#reduce-dbcache-after-full-sync)#prepare-nginx-reverse-proxy after full sync.
+Make sure that you have [reduced the database cache of Bitcoin Core](bitcoin-client.md#reduce-dbcache-after-full-sync) after full sync.
 
 ### Install dependencies
 
 * Install build tools needed to compile Electrs from the source code
 
   ```sh
-  $ sudo apt install cargo clang cmake
+  $ sudo apt install cargo clang cmake build-essential librocksdb-dev
```
kristapsk, line 46: https://github.com/romanz/electrs/blob/master/doc/install.md says 6.11.4 release of rocksdb is required. Not specifying version on my system (bullseye) installs 5.17.2-3. When it's not available, that guide suggests building manually.

VajraOfIndra:
> Take the opportunity to put in line 284/291 the name of the hidden service:
> 
> `# Hidden Service Electrs SSL`

Done :+1: 

_(sorry I edited your comment by mistake while trying to reply)_

guide/bitcoin/electrum-server.md
VajraOfIndra: Using `librocksdb-dev=6.11.4-3` should work on RasPiOS, right? If so we could simply use that command.

kristapsk: Not sure, I'm running Odroid HC4, for which default is Ubuntu Server.

VajraOfIndra: On my Raspberry Pi 4 with Raspberry Pi OS Lite 64-bit I get:
> admin@raspibolt:~ $ sudo apt list --all-versions librocksdb-dev 
> Listing... Done
> librocksdb-dev/stable 6.11.4-3+deb11u1 arm64

So this should work for the users that follow the default guide.

I am out of my depths here about what would be best for the guide.. We have 3 options I believe:

1. Static linking (existing method)
2. Dynamic linking, using `apt` to install `librocksdb`
3. Dynamic linking, manually building `librocksdb`

Which one would be the most robust and cross-OS compatible option..?

twofaktor:
hello guys, is possible to change lines 101-102 to using the `/temp` folder?

I watched this dev video installing and it made me think:

[asciicast](https://asciinema.org/a/XKznxilP4O7lCZiVZ9vZNd5vx)

Then in line 123 we make it persistent and It should work. Wdyt?

VajraOfIndra:
> hello guys, is possible to change lines 101-102 to using the `/temp` folder?

You mean _/tmp_? The issue I think is that the GitHub repo would be deleted at the next reboot and Electrs couldn't be simply updated, right?

However, I was wondering why do we use `/home/admin/rust/` to host the Electrs GitHub repo and not the home directory of the Electrs user `/home/electrs` ..?

twofaktor:
> > hello guys, is possible to change lines 101-102 to using the `/temp` folder?
> 
> You mean `/tmp`? The issue I think is that the GitHub repo would be deleted at the next reboot and Electrs couldn't be simply updated, right?

https://github.com/romanz/electrs/blob/master/doc/upgrading.md#upgrading-manual-installation

`1. Enter your electrs source directory, usually in ~/ but some people like to put it in something like ~/sources. **If you've deleted it, you need to git clone again**.` 

I think we just need to re-clone the git repository in the upgrade process

> 
> However, I was wondering why do we use `/home/admin/rust/` to host the Electrs GitHub repo and not the home directory of the Electrs user `/home/electrs` ..?

I don't know either, maybe due to permissions, and why "rust" as the subfolder name?

guide/bitcoin/electrum-server.md
```diff
@@ -119,7 +119,7 @@ We get the latest release of the Electrs source code, verify it, compile it to a
   The compilation process can take up to one hour.
 
   ```sh
-  $ cargo build --locked --release
+  $ ROCKSDB_INCLUDE_DIR=/usr/include ROCKSDB_LIB_DIR=/usr/lib CARGO_NET_GIT_FETCH_WITH_CLI=true cargo build --locked --release
   $ sudo cp ./target/release/electrs /usr/local/bin/
```
twofaktor, line 123: do you think is better to use this command instead? `sudo install -m 0755 -o root -g root -t /usr/local/bin ./target/release/electrs`

twofaktor (changes requested):
what do you think about this?

guide/bitcoin/electrum-server.md
Kixunil: @kristapsk that's really surprising, bullseye is supposed to have 6.11.4. Are you really sure you have bullseye and not a messed up OS? (presumably something wrong with `/etc/apt/sources.list*`)

@VajraOfIndra Unless Raspibolt has something else using the same version of rocksdb 3 is definitely worse than 1. If you can't do 2, do 1.

Kixunil:
> just need to re-clone the git repository in the upgrade process

Unless there is a lack of storage space to motivate deletion you will save some resources by not re-cloning. If you don't have enough space to store electrs source code you have much bigger problems.

guide/bitcoin/electrum-server.md
Kixunil: `install` is indeed more explicit, which is nice. In practice they do the same thing here.

guide/bitcoin/electrum-server.md
```diff
@@ -328,16 +335,20 @@ Make sure to check the [release notes](https://github.com/romanz/electrs/blob/ma
   $ cd /home/admin/rust/electrs
 
   # update the local source code and show the latest release tag (example: v0.9.9)
+  $ git clean -x -f d
   $ git fetch
   $ git tag | sort --version-sort | tail -n 1
   > v0.9.9
 
+  # Verify the release
+  $ git verify-tag v0.9.9
```
Kixunil, line 344: Ideally people Would verify through independent channels (twitter?) that this is indeed the latest version. Otherwise if there is a vulnerable version an attacker could supply that one instead. (OTOH TBF practically exploitable bug is unlikely in `electrs` since it normally doesn't have published RPC. It could go through chain but that'd be very visible.)

Kixunil (commented):
Looks mostly good. I definitely recommend double checking the paths though.

guide/bitcoin/electrum-server.md
```diff
@@ -280,7 +287,7 @@ Note that the remote device needs to have Tor installed as well.
   ```
 
   ```sh
-  ############### This section is just for location-hidden services ###
```
twofaktor, line 283: Recover this line to standardize with the rest of the guide
Ref: e.g https://raspibolt.org/guide/raspberry-pi/privacy.html#ssh-server

guide/bitcoin/electrum-server.md
twofaktor, line 123: Change to: `$ sudo install -m 0755 -o root -g root -t /usr/local/bin ./target/release/electrs`

twofaktor (approved):
Some suggestions

guide/bitcoin/electrum-server.md
```diff
@@ -280,7 +287,7 @@ Note that the remote device needs to have Tor installed as well.
...
+  # Hidden Service Electrs SSL
   HiddenServiceDir /var/lib/tor/hidden_service_electrs/
   HiddenServiceVersion 3
   HiddenServicePort 50002 127.0.0.1:50002
```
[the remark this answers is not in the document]
twofaktor: Some wallets like Electrum Wallet Mobile only support TLS connection type with Electrum Servers, It's only to maintain compatibility with these wallets

guide/bitcoin/electrum-server.md
VajraOfIndra: Fixed!
I updated the OP to add it to the list of changes.

guide/bitcoin/electrum-server.md
VajraOfIndra: Fixed. I find it weird we always include that line though, as the user does not need to add that to the config file.

VajraOfIndra:
> Some suggestions
Applied your suggestions, PTAL and resolve the conversations if you're happy with teh changes.

guide/bitcoin/electrum-server.md
VajraOfIndra: I'd go for #2 as it seems to work fine with RasPiOS Lite 64-bit..

@Stadicus wdyt? #2 is recommended but could there be too many version issues and therefore #1 is safer?

guide/bitcoin/electrum-server.md
VajraOfIndra: Thanks both, I replaced `cp` by this `install` command, see below.

guide/bitcoin/electrum-server.md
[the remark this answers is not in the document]
VajraOfIndra: > Some wallets like Electrum Wallet Mobile only support TLS connection type with Electrum Servers, It's only to maintain compatibility with these wallets

Are you sure?

In their [documentation](https://electrum.readthedocs.io/en/latest/tor.html), they give this example launch command:
`./electrum -1 -s electrums3lojbuj.onion:50001:t -p socks5:localhost:9050`

This is using Tor without TLS connection, right?

guide/bitcoin/electrum-server.md
VajraOfIndra: > Ideally people Would verify through independent channels (twitter?) that this is indeed the latest version.

What's the best independent channel that the Electrs dev use to announce their latest release?

guide/bitcoin/electrum-server.md
```diff
@@ -146,13 +146,14 @@ We get the latest release of the Electrs source code, verify it, compile it to a
...
-  ```sh
+  ```toml
   # RaspiBolt: electrs configuration
   # /data/electrs/electrs.conf
 
   # Bitcoin Core settings
   network = "bitcoin"
-  daemon_dir= "/home/bitcoin/.bitcoin"
+  daemon_dir= "/data/bitcoin"
+  cookie_file= "/data/bitcoin/.cookie"
```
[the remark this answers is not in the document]
VajraOfIndra: > Unless the other guides changed the directories this would break `electrs`. To fix permissions problems you can set `ReadOnlyPaths=/home/bitcoin/.bitcoin` in the service file.

How would it break electrs? aren't these options meant to be used to give the absolute paths of the daemon and cookie file?

I added that together with the `--skip-default-conf-files` flag in the service file in order for us to enable the `ProtectHome` option... I'm a bit confused now by what's the best way to reconcile all this. Would love some suggestions.

VajraOfIndra:
@Stadicus @twofaktor @Kixunil @kristapsk 

Following the latest [feedback ](https://github.com/raspibolt/raspibolt/issues/747#issuecomment-1291646191) from @Stadicus, I updated this PR accordingly. 
The complete and updated list of changes is in the OP.

The remaining unresolved topics are:
- Runtime directory permissions (and why it remains empty even after Electrs is launched)
- How to enable `ProtectHome` in service file while not breaking up Electrs
- (minor) What channel to use to independently verify version of latest release 

I'll try to continue testing these changes on my node buy having an aready installed Electrs i'm not sure I can identify all issues. Would be great to have someone test it from a virgin node.

If you think this PR contains too many changes at once, I could break it down in several PRs to make it easier to test and merge (but I hope it won;t be necesarry! :P)

guide/bitcoin/electrum-server.md
```diff
@@ -281,9 +288,10 @@ Note that the remote device needs to have Tor installed as well.
 
   ```sh
   ############### This section is just for location-hidden services ###
+  # Hidden Service Electrs
   HiddenServiceDir /var/lib/tor/hidden_service_electrs/
   HiddenServiceVersion 3
-  HiddenServicePort 50002 127.0.0.1:50002
+  HiddenServicePort 50001 127.0.0.1:50001
```
twofaktor, line 294: Why this change? Some wallets are incompatible with no-SSL connections pairing

twofaktor (commented):
Question about no-SSL connection incompatibility throught Tor

guide/bitcoin/electrum-server.md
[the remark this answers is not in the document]
Kixunil: Is `bitcoind` storing its data in `/data/bitcoin` when configured according to Raspibolt? If not then the path is invalid.

guide/bitcoin/electrum-server.md
Kixunil: @romanz what do you think is best here?

Kixunil:
> Runtime directory permissions (and why it remains empty even after Electrs is launched)

I think you have to specify `ReadWritePaths=/data/electrs`. Or put it into the standardized location `/var/lib/electrs` and set `RuntimeDir=electrs`.

> How to enable `ProtectHome` in service file while not breaking up Electrs

Add `--skip-default-conf-files` to `electrs` invocation and IFF there are any paths in `/home` that electrs **should** access add them to `ReadOnlyPaths`

guide/bitcoin/electrum-server.md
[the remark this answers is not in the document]
twofaktor: "Electrum Wallet Mobile"

twofaktor:
> > just need to re-clone the git repository in the upgrade process
> 
> Unless there is a lack of storage space to motivate deletion you will save some resources by not re-cloning. If you don't have enough space to store electrs source code you have much bigger problems.

I highly doubt we don't have enough space to store 1.87MB. In either case, the contents of the /temp folder are removed on every reboot or may be removed after installation.

VajraOfIndra:
Closing this large PR as lack of feedback on some points/questions.
Will reopen a couple of smaller PR for the agreed-upon changes for Electrs.

PR #1085, opened 2022-09-05
Update Bonus Guide: Joinmarket

sig-noi:
#### What

What is the reason of this change?
Update and integrate with RaspiBolt layout. Port guide from original author kristapsk's github repo. Added dedicated joinmarket user approach to mirror other RaspiBolt applications. Added activate.sh script to make starting up Joinmarket slightly easier. Added some description and links that I found helpful as I was learning how to use JoinMarket.

### Why

Why is this change important?
Will better integrate JoinMarket.md with the RaspiBolt bonus guides

#### How

How was this change accomplished?
I made most of the changes as I worked through the guide on my own RaspiBolt v3, then modified it to mirror some of the more canonical RaspiBolt guides.

#### Scope

- [ ] significant change to core configuration
- [x] independent bonus guide *update*
- [ ] simple bug fix

Fixes # (link issue)
n/a

#### Test & maintenance

How can the changes be tested?
I've been using this configuration on RaspiBolt v3 without issue. Getting another few folks to run it would be good. 

Is there ongoing maintenance effort? If this is a bonus guide: are you willing to update it from time to time?
 Yes. Sounds like original author kristapsk will also help keep updated.

#### Animated GIF (optional)
learning github was hard enough! wa wa

L9JMnv

PR #1105, opened 2022-09-27
Fix JoinMarket guide location in menu

kristapsk:
See https://github.com/raspibolt/raspibolt/pull/1085#issuecomment-1260002075.


VajraOfIndra (approved):
LGTM, ACK

PS: you didn't include `has_children: false` but I don't think it matters.. I don't remember if this option is there for a specific reason or not.

PR #1116, opened 2022-10-05
Update security.md

xmready:
Change `/etc/security/limits.conf` to `/etc/security/limits.d/90-limits.conf` to avoid the configuration file being overwritten during a system upgrade. Using `.d` directories whenever possible is a good strategy for longevity.

### DIFF
--- a/guide/raspberry-pi/security.md
+++ b/guide/raspberry-pi/security.md
@@ -182,16 +182,18 @@ The initial configuration is fine, as it protects SSH by default.
 If your RaspiBolt is swamped with internet requests (honest or malicious due to a DoS attack), you will quickly encounter the "can't accept connection: too many open files" error.
 This is due to the limit of open files (representing individual TCP connections) set too low.
 
-Edit each of the following three files, add the additional line(s) right before the end comment, save and exit.
+Create the file `/etc/security/limits.d/90-limits.conf`, copy these lines into it, save and exit.
 
 ```sh
-$ sudo nano /etc/security/limits.conf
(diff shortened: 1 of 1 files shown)


kristapsk (approved):
ACK cb032a5d68ad30938eb495878a44d9a80c07806b

VajraOfIndra (approved):
tACK

With `/etc/security/limits.conf` all commented out and an empty `/etc/security/limits.d/90-limits.conf`:

> admin ~ $ ulimit -Sn
> 1024
> admin ~ $ ulimit -Hn
> 1048576

With `/etc/security/limits.conf` all commented out and the limits set in `/etc/security/limits.d/90-limits.conf`:

> admin ~ $ ulimit -Sn
> 128000
> admin ~ $ ulimit -Hn
> 128000

VajraOfIndra:
What process are we trying to protect by increasing these values..? Are they controlled by systemd? 

If so, how does the `DefaultLimitNOFILE` option from `/etc/systemd/user.conf` (soft limit) and `/etc/systemd/system.conf` (hard limit) interact with our limits above set in `/etc/security/limits.d/90-limits.conf`?

The default value for `DefaultLimitNOFILE` on my node is:

> admin ~ $ sudo systemctl show -p DefaultLimitNOFILE
> DefaultLimitNOFILE=524288

Based on this [StackExchange comment](https://superuser.com/a/1322187) it would seem that the values in /etc/security/limits.conf` (and associated `limits.d` conf files) values are ignored if using `systemd`.

Wdyt?

PR #1117, opened 2022-10-06
[NEW BONUS GUIDE] Samourai Whirlpool

DanielPX1:
#### What

- [Samourai Whirlpool](https://docs.samourai.io/whirlpool/start) is a free and open source (FOSS), non custodial, chaumian CoinJoin platform. Its goal is to mathematically disassociate the ownership of inputs to outputs in a given bitcoin transaction.

- Whirlpool is installed in a way to require own Dojo backend connection to assure no trust in third party servers

- Whirlpool WST tool added in "Extra section" for users to verify without trust anonsets of their own UTXO's. WST tool helps users have idea about anonymity sets "they hide in", thus giving option to consider, if number of mixes was "safe enough" to spend UTXO (ex. leaving Whirlpool)

### Why

- Additional guide for Samourai Dojo users, who want to mix on their node in a safe and trustless DIY way. Mixing on the node happens 24/7. Mixing client is not required to be open on the android device, as node manages mixing. 
- More additional privacy and coinjoin varieties for Raspibolt.

#### How

By installing and testing on Raspberry pi 4GB 

#### Scope

- [ ] Review of formatting, mistakes, bugs etc.

#### Test & maintenance

Following guide and reporting issues.

-  Installation can be tested without Samourai Wallet up to "Authenticate Section"
- WST Tool in "Extra" can be verified both without SW and Whirlpool

Willing to maintain w/ Dojo 

pic

### DIFF
--- a/guide/bonus/lightning/lntop.md
+++ b/guide/bonus/lightning/lntop.md
@@ -149,7 +149,7 @@ To use all the functionalities of lntop, use the following keys:
 
   ```sh
   $ cd /usr/local/bin
-  $ rm -r lntop
+  $ sudo rm -r lntop
   ```
 
 <br /><br />


DanielPX1:
Tested on Raspberry pi 4GB. Working on my system

greenart7c3:
Tested on a x86 machine and it worked

DanielPX1:
Ready 4 reviews, tested on 2 x86 machines and 1 Raspi 4GB

PR #1118, opened 2022-10-08
Replace "Wants" by "PartOf" in systemd service files

### LINKED ISSUE #848: [BUG] BTC RPC Explorer breaks on Bitcoin Core restart
### Describe the bug
If Bitcoin Core is restarted, the .cookie file is recreated with new access credentials. Currently, BTC RPC Explorer does not detect this change and just keeps trying to query bitcoind with the old credentials it read from the cookie on startup. This renders the application unusable, until it is restarted (reading the new credentials and accessing bitcoind successfully again).

image

### Location in guide
<https://raspibolt.org/btcrpcexplorer.html>

### Expected behavior
BTC RPC Explorer should detect a change in the `.cookie` file content and always use the latest RPC access credentials.

This is not something this guide can fix, so I opened an issue in the project's repository:  
https://github.com/janoside/btc-rpc-explorer/issues/393

VajraOfIndra:
#### What

This is a proposal to remove `Wants=` from all the service files and when appropriate replace by `PartOf=`.

### Why

> 
> Wants= 
>      Units listed in this option will be started if the configuring unit is. However, if the listed units fail to start or cannot be added to the transaction, this has no impact on the validity of the transaction as a whole, and this unit will still be started. This is the recommended way to hook the start-up of one unit to the start-up of another unit. [(source)](https://www.freedesktop.org/software/systemd/man/systemd.unit.html)

Using `Wants` can lead to unwanted restarts of Bitcoin Core when trying to shut it down but a dependent program like the BTC RPC Explorer is still running (e.g. #848).

> PartOf=
>     Configures dependencies similar to Requires=, but limited to stopping and restarting of units. When systemd stops or restarts the units listed here, the action is propagated to this unit. Note that this is a one-way dependency — changes to this unit do not affect the listed units.
       When PartOf=b.service is used on a.service, this dependency will show as ConsistsOf=a.service in property listing of b.service. ConsistsOf= dependency cannot be specified directly. [(source)](https://www.freedesktop.org/software/systemd/man/systemd.unit.html)

Instead, using `PartOf` allows for the following behaviour (using Bitcoin Core and the Exlorer as an example):
- Start the Explorer when Bitcoin Core starts
- When starting Bitcoin Core, if the Explorer can't start (e.g. not installed), Bitcoin Core should still starts.
- The Explorer should start only AFTER Bitcoin Core has started
- If the Explorer is started (e.g. manually) but Bitcoin Core is not, do NOT start Bitcoin Core
- Stop the Explorer if Bitcoin Core stops

Note: this fixes #848 (even though the core of the issue is due to the mishandling by the Explorer of the cookie file)

#### How

This PR proposes the following dependencies for the programs from the core guide (if approved and merged, another pR could be used to update the bonus guides to use `PartOf`)

|Service|After|PartOf|
|-------|------|-------|
|bitcoind|network.target|  |
|electrs|bitcoind.service|bitcoind.service|
|btcrpcexplorer|bitcoind.service electrs.service|bitcoind.service|
|lnd|bitcoind.service|bitcoind.service|
|scb-backup|lnd.service|lnd.service|
|rtl|lnd.service|lnd.service|

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

Fixes #848

#### Test & maintenance

- Apply changes. Test by restartin Bitcoin Core and verifying that the Explorer is accessible (+ other tests with other programs/dependencies).

domino effect


guide/bitcoin/blockchain-explorer.md
```diff
@@ -252,8 +252,8 @@ In order to do that, we create a systemd unit that starts the service on boot di
 
   [Unit]
   Description=BTC RPC Explorer
-  After=network.target bitcoind.service
-  After=electrs.service
```
twofaktor, line 256: Why delete this line?

twofaktor (commented):
Some comment

guide/bitcoin/blockchain-explorer.md
VajraOfIndra: My thinking was that the Explorer is not dependent on Electrs, the Electrum server just provide extra features. But since the Explorer expects an Electrs instance to be running, maybe it's better if we start it after Electrs. And if Electrs can't start, then the Explorer starts anyway with After so it doesn't hurt... expect that it might take a bit more time to be available in the browser.

I'll add `electrs.service` to the line 255.

VajraOfIndra:
I just added `PartOf=lnd.service` to the SCB backup systemd service as well. I forgot about that one. There is no reason for the backup service to run when LND is shut down. And if LND restarts, the backup service should retart as well.

sig-noi (approved):
tACK

VajraOfIndra:
> tACK

Does starting a service make the dependent programs start as well? I need to do more testing but it works for me when I stop a service (e.g. LND, it stops RTL) but it does not seem to work as intended when starting a service...

VajraOfIndra:
Tested on my node, PartOf allows for services to shut down when the parent service shut down. When the parent service starts, the children services do NOT start. 

A way to make them start is to add Wants in the parent service.

However the scope of this PR is too large. + lack of interest. Closing it.
I'll open a PR focused on BTC RPC Explorer only, to solve the cookie bug it has.

PR #1120, opened 2022-10-10
Update Lightning Terminal to v0.8.1 (contains the LND v0.15.2 hotfix)

VajraOfIndra:
#### What

A PR to update Lighting Terminal to v0.8.1

### Why

- The guide is outdated (many versions behind)
- v0.8.1 contains the hot fix of LND v0.15.2. It does not really matter for this guide as the guidelines assume we are using an external LND node and NOT the one contained in Lightning Terminal... but just in case it's safer to have the guide with the version that contains the bug fix just in case a user would wnat to try out to use the LND as part of LiT..

#### Scope

- [ ] independent bonus guide
- [x] update of bonus guide
- [ ] simple bug fix

#### Test & maintenance

- Verify syntax
- If possible, update on a node with LiT

### DIFF
--- a/guide/bonus/lightning/lightning-terminal.md
+++ b/guide/bonus/lightning/lightning-terminal.md
@@ -50,7 +50,7 @@ Because Pool is alpha software, Lightning Terminal is also alpha software.
 
 We will configure Lightning Terminal to use our own LND node. However, when installing Lightning Terminal, it replaces the existing `lncli` binary (the CLI tool that allows us to interact with our LND daemon) with its own version of the `lncli` binary. The Lightining Terminal `lncli` binary is generally an older version and therefore might not have all the functionalities of the `lncli` associated with the version of LND that you're running. 
 
-Just don't be surpised if you want to use a brand new `lncli` command and it does not seem to be recognized! :)
+Just don't be surprised if you want to use a brand new `lncli` command and it does not seem to be recognized! :)
 
 ### Firewall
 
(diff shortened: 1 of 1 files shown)


blckbx:
Since quite a few problems occurred around `lncli` and Lightning Terminal installations. What do you think about a warning saying installing LiT via `sudo instal....` overwrites the current `lncli` binary in `/usr/local/bin`. Most of the time LiT contents are behind current LND versions, so users end up using an older version of `lncli`.

VajraOfIndra:
@blckbx could you PTAL? 🙌

guide/bonus/lightning/lightning-terminal.md
```diff
@@ -46,6 +46,12 @@ Because Pool is alpha software, Lightning Terminal is also alpha software.
 
 ## Preparations
 
+### Warning
+
+We will configure Lightning Terminal to use our own LND node. However, when installing Lightning Terminal, it replaces the existing `lncli` binary (the CLI tool that allows us to interact with our LND daemon) with its own version of the `lncli` binary. The Lightining Terminal `lncli` binary is generally an older version and therefore might not have all the functionalities of the `lncli` associated with the version of LND that you're running. 
+
+Just don't be surpised if you want to use a brand new `lncli` command and it does not seem to be recognized! :)
```
blckbx, line 53:
```suggestion
Just don't be surprised if you want to use a brand new `lncli` command and it does not seem to be recognized! :)
```

blckbx (approved):
LGTM 🎉

PR #1121, opened 2022-10-10
Update mempool.md

xmready:
1) Fix typo

2) Add `USE_ONION` property to mempool-config.json to fix pool name warnings

### DIFF
--- a/guide/bonus/bitcoin/mempool.md
+++ b/guide/bonus/bitcoin/mempool.md
@@ -119,7 +119,7 @@ For improved security, we create the new user "mempool" that will run the Mempoo
   > MariaDB [(none)]>
   ```
 
-* Enter the following commands in the shell and exit. The instructions to enter in the MariaDB shell with start with "MDB$". Change "Password[M]" to the random password generated above.
+* Enter the following commands in the shell and exit. The instructions to enter in the MariaDB shell start with "MDB$". Change "Password[M]" to the random password generated above.
 
   ```sql
   MDB$ create database mempool;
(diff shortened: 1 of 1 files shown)


VajraOfIndra:
Thanks for your PR! 👍 

>     2. Add `USE_ONION` property to mempool-config.json to fix pool name warnings

Could you explain what does this option do? I tried to find information in their repo but couldn't find anthing.. What happens if this optioj is NOT in teh config file? and what is the default?

xmready:
> Could you explain what does this option do? I tried to find information in their repo but couldn't find anthing.. What happens if this option is NOT in the config file? and what is the default value, true or false?

Without this option, the mempool backend was throwing errors about not being able to retrieve pool names. There isn't enough information about these properties in the mempool repo to say exactly what the default is, but looking at the sample config file that comes with the latest build, the section in question has the following keys and values:
```
"SOCKS5PROXY": {
    "ENABLED": false,
    "USE_ONION": true,
    "HOST": "127.0.0.1",
    "PORT": 9050,
    "USERNAME": "",
    "PASSWORD": ""
```

This may require some testing to be sure

VajraOfIndra (approved):
Found this in the Mempool repo (in this [file]( https://github.com/mempool/mempool/blob/da6c72e9b7a153c8da28eab830e10abee7f8bed1/backend/src/api/fiat-conversion.ts)):

`   const fiatConversionUrl = (config.SOCKS5PROXY.ENABLED === true) && (config.SOCKS5PROXY.USE_ONION === true) ? config.PRICE_DATA_SERVER.TOR_URL : config.PRICE_DATA_SERVER.CLEARNET_URL;`

So looks like is USE_ONION is false, the website will use the clearnet address to get the fiat conversion rate. So has to be true to use the onion address and be more private (if I understand correcttly..)

utACK (tested on my node but I don't know how to check if the onion address is indeed being used and not the clearnet address)

PR #1126, opened 2022-10-15
JoinMarket guide: add syntax highlighting

sig-noi:
#### What

Make consistent with main guides. Looks pretty.

### Why

Pretty = good?

#### How

I learned a little more about markdown and added the sh and ini before the code blocks

#### Scope

- [ ] independent bonus guide
- [x] simple formatting fix

Fixes # n/a

#### Test & maintenance

Can't test but believe it will work since it's the same syntax as the primary guides.

### DIFF
--- a/guide/bonus/bitcoin/joinmarket.md
+++ b/guide/bonus/bitcoin/joinmarket.md
@@ -40,7 +40,7 @@ We set up [Joinmarket clientserver](https://github.com/JoinMarket-Org/joinmarket
 
 * With user "admin", install necessary dependencies
 
-```
+```sh
 $ sudo apt install python-virtualenv curl python3-dev python3-pip build-essential automake pkg-config libtool libgmp-dev libltdl-dev libssl-dev libatlas3-base libopenjp2-7
 ```
 
@@ -51,7 +51,7 @@ If you get `E: Package 'python-virtualenv' has no installation candidate` error 
 
 * This wallet will be used by JoinMarket to store addresses as watch-only. It will use this wallet when it communicates with bitcoin core via rpc calls.
 
-```
+```sh
 $ bitcoin-cli -named createwallet wallet_name=jm_wallet descriptors=false
 ```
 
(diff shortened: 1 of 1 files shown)


sig-noi:
Thank you for the suggestion @VajraOfIndra. Updated pull with the following:

- indented code blocks by two spaces per @VajraOfIndra's suggestion
- added line breaks between major sections per main guide convention
- removed double returns per main guide convention
- updated "upgrade JoinMarket" section to spoonfeed the process
-> I did have issues with the unlink /data/joinmarket - That command wanted sudo to work and seemed unnecessary as we are still using the `.joinmarket -> /data/joinmarket` symlink. I changed the command to `unlink joinmarket` to remove the `joinmarket -> joinmarket-clientserver-currentversion` that would be replaced by the `joinmarket -> joinmarket-clientserver-newversion` link. This worked for me but let me know if it's wrong for some reason.

kristapsk:
Not sure about these changes to `sh` in many places where there is command output. It IMHO doesn't make it better any way, just adds unnecessary "syntax highlightning" to output, which isn't related to shell hightlighting at all.

image

VajraOfIndra:
> Not sure about these changes to `sh` in many places where there is command output. It IMHO doesn't make it better any way, just adds unnecessary "syntax highlightning" to output, which isn't related to shell hightlighting at all.

Makes sense.. so for blocks of command ouputs we should not define any language, just have the three backticks on their own, right?

sig-noi:
> Not sure about these changes to sh in many places where there is command output. It IMHO doesn't make it better any way, just adds unnecessary "syntax highlighting" to output, which isn't related to shell highlighting at all.

Fair point about indiscriminate syntax highlighting. While this seems to be the convention in most of the main guides, that doesn't make it right. I could see fixing this here and then possibly standardizing the main guides to this one.

sig-noi:
In removing syntax highlighting from the command output I ended up creating more code blocks, which may be confusing. Let me know if you think these should be re-consolidated back into single blocks of code (with shell syntax highlighting).

VajraOfIndra:
> Thoughts on bringing this convention back to the main guide? @VajraOfIndra

Have no idea. It seems to make more sense in terms of markdown syntax, but does it still look good in all cases? I'll let others opine on that.

VajraOfIndra:
I think we can merge this PR..
@kristapsk are you ok with new formatting?

guide/bonus/bitcoin/joinmarket.md
```diff
@@ -34,135 +34,144 @@ We set up [Joinmarket clientserver](https://github.com/JoinMarket-Org/joinmarket
...
+  ```
+  ```
+  > gpg: assuming signed data in 'joinmarket-clientserver-0.9.8.tar.gz'
+  > gpg: Signature made Thu Sep 15 15:46:38 2022 EEST
+  > gpg:                using RSA key 2B6FC204D9BF332D062B461A141001A1AF77F20B
+  > gpg: Good signature from "Adam Gibson (CODE SIGNING KEY) <ekaggata@gmail.com>" [unknown]
+  > gpg: WARNING: This key is not certified with a trusted signature!
+  > gpg:          There is no indication that the signature belongs to the owner.
+  > Primary key fingerprint: 2B6F C204 D9BF 332D 062B  461A 1410 01A1 AF77 F20B
+  ```
```
kristapsk, line 123: I would just use single ``` block here and below, without `sh`, which doesn't do any colouring here anyway. But might be just my preference.

sig-noi: Fixed

VajraOfIndra:
LGTM, let's go to master 🎉

PR #1130, opened 2022-10-20
`ufw allow ssh` should be in lowercase

kristapsk:
Uppercase `SSH` gives me an error, lowercase works (Ubuntu Server).

```
admin ~ ₿ sudo ufw allow SSH 
ERROR: Could not find a profile matching 'SSH'
admin ~ ₿ sudo ufw allow ssh
Skipping adding existing rule
Skipping adding existing rule (v6)
```

### DIFF
--- a/guide/raspberry-pi/security.md
+++ b/guide/raspberry-pi/security.md
@@ -127,7 +127,7 @@ We'll open the port for Electrs and web applications later if needed.
   $ sudo apt install ufw
   $ sudo ufw default deny incoming
   $ sudo ufw default allow outgoing
-  $ sudo ufw allow SSH
+  $ sudo ufw allow ssh
   $ sudo ufw logging off
   $ sudo ufw enable
   ```


VajraOfIndra:
FYI for others (just learnt about it as I didn't know how it worked): When using a service by name (rather than a port number), UFW looks into `/etc/service` for a match (as per [man page](https://manpages.ubuntu.com/manpages/jammy/en/man8/ufw.8.html)):

> [...] ufw will also check /etc/services
>        for the port and protocol if specifying a service by name.  Eg:
> 
>          ufw allow smtp

On our node, we can check with:

```sh
$ cat /etc/services | grep ssh
> ssh             22/tcp                          # SSH Remote Login Protocol
```
So we get:

```sh
$ sudo ufw allow ssh
> 22/tcp                     ALLOW       Anywhere
```

VajraOfIndra:
Btw, what about using a [non-standard SSH port to harden our security](https://www.techrepublic.com/article/how-to-configure-ssh-to-use-a-non-standard-port-with-selinux-set-to-enforcing/)? yea/nay..?

VajraOfIndra (approved):
utACK

twofaktor:
> Btw, what about using a [non-standard SSH port to harden our security](https://www.techrepublic.com/article/how-to-configure-ssh-to-use-a-non-standard-port-with-selinux-set-to-enforcing/)? yea/nay..?

Yes, I have a non-standard port configured to improve security, 

otherwise: what about using from `192.168.0.0/16` instead of `anywhere`?

VajraOfIndra:
> otherwise: what about using from `192.168.0.0/16` instead of `anywhere`?

Maybe we could write a bonus guide in the +Raspberry Pi section with a few ideas on how to harden network security, including:
1. Use non-standard SSH port
2. Restrict to `192.168.0.0/16` (instead of `anywhere`)

Would make interested users learn a bit more about firewall and networking while not making the main guide overly complicated.

Wdyt?

PR #1135, opened 2022-10-27
[NEW BONUS GUIDE]  Amboss.space heartbeat pings without Thunderhub

satsHIVE:
#### What

Easy to use script to send heartbeat pings to the amboss.space monitoring service

### Why

If Thunderhub is not available on a node due to variety of reasons, the only remaining option is a "manual" ping to be sent to the amboss monitoring service.

#### How

Simple bash script executed on predefined intervals via cronjob.

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

Following the instructions, check if the status in the amboss health dashboard is being updated. One should disable the Thunderhub pings if they are activated from there.

#### Animated GIF (optional)

n/a


twofaktor:
If not enable channel balance sending info in ThunderHub, ThunderHub does not send this info to Amboss by default. We shouldn't worry about this, this works the same as health checks and backups options, so I think is not a privacy risk to the RaspiBolt nodes

Don't trust, verify: https://github.com/apotdevin/thunderhub/blob/master/src/client/src/views/amboss/Balances.tsx#L26-L30

satsHIVE:
Can confirm what you say. Checked with my installation of 0.13.16. 
On the other hand, though, I found reporting via the script way more reliable than TH. Used this script for the recent weeks and the false-positives in the reporting fell by 90% at least. From ~10x10min reported outages per day in the past (while my nod is actually routing) I fell to 1-2 a week.

twofaktor:
This is not necessary. If you do not enable the sending of channel balance information in ThunderHub, ThunderHub does not send this information to Amboss by default. We shouldn't worry about this, it works the same as the health check and backup options, so I think it's not a privacy risk for the RaspiBolt nodes if we don't enable it.

https://github.com/apotdevin/thunderhub/blob/master/src/client/src/views/boss/Balances.tsx#L26-L30

Maybe we could add a notice in the Thunderhub extras section: https://raspibolt.org/guide/bonus/lightning/thunderhub.html#extras

satsHIVE:
Ok, the "why" become a bit misleading, but the point of this at the end is still to provide the ability to use the hatcheck status without TH.

twofaktor:
> Can confirm what you say. Checked with my installation of 0.13.16. On the other hand, though, I found reporting via the script way more reliable than TH. Used this script for the recent weeks and the false-positives in the reporting fell by 90% at least. From ~10x10min reported outages per day in the past (while my nod is actually routing) I fell to 1-2 a week.

I agree with you, I usually get a lot of false-positives with Thunderhub -> Amboss health check method.

I'll try to leave you a review. First I think I can see, is it necessary to create a unique user for this use? In other similar guides, this step was not included

refs: https://raspibolt.org/guide/lightning/channel-backup.html
https://raspibolt.org/guide/bonus/lightning/lntop.html

satsHIVE:
I consider it more cleaner approach. If we set up a user, it is way easier to uninstall by only deleting the user. We basically set the files and the cron in the user, so therefore no leftovers after the “uninstallation”.

guide/bonus/lightning/ambossping.md
```diff
@@ -0,0 +1,154 @@
...
+  Please remember that the error messages of curl are not explicit and are not always 100% correct. From the error messages, you could make some assumptions about the issues you may experience are caused. From the errors mentioned above, only **Error: 35** leads to remote side issues, though.
+
+---
+
+## Uninstall
+
+* If you want to remove the ambossping and stop reporting health status with this script, delete the “ambossping” user with the “root” user.
+
+  ```sh
+  $ userdel -r ambossping
```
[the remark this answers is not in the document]
satsHIVE: It is executed with the “root” user.

twofaktor:
I think the Thunderhub -> Amboss method issue may be due to issues with Tor, as we have this line configured for the Thunderhub config:

`TOR_PROXY_SERVER=socks://127.0.0.1:9050`

It usually fails me at the time I've had problems with Tor. 

BTW: Is Tor used in this guide for the communication of our node with Amboss?

guide/bonus/lightning/ambossping.md
```diff
@@ -0,0 +1,154 @@
...
+
+You should see something like this in terminal.
+
+  ```ini
+  Timestamp: 2022-10-27T15:53:33+0000
+  Signature: d71y...uduf
+  Sending ping...
+  2022-10-27T15:53:33+0000 - Ping sent
+  ```
+
```
twofaktor, line 105: add: 

`exit`

to return to the "admin" user

twofaktor (changes requested):
a suggestion in code

guide/bonus/lightning/ambossping.md
satsHIVE: Fixed. Doesn't need to exit the current session, so adapted the next paragraph.

guide/bonus/lightning/ambossping.md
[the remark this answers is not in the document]
satsHIVE: fixed

guide/bonus/lightning/ambossping.md
```diff
@@ -0,0 +1,165 @@
...
+  ```ini
+  #!/bin/bash
+  URL="https://api.amboss.space/graphql"
+  NOW=$(date -u +%Y-%m-%dT%H:%M:%S%z)
+  echo "Timestamp: ${NOW}"
+  SIGNATURE=$(/usr/local/bin/lncli signmessage "$NOW" | jq -r .signature)
+  echo "Signature: ${SIGNATURE}"
+  JSON="{\"query\": \"mutation HealthCheck(\$signature: String!, \$timestamp: String!) { healthCheck(signature: \$signature, timestamp: \$timestamp) }\", \"variables\": {\"signature\": \"$SIGNATURE\", \"ti>
```
blckbx, line 74: JSON string is cut off after `\"ti`

blckbx (commented):
incomplete JSON string

guide/bonus/lightning/ambossping.md
satsHIVE: Thanks, found it. Should be OK now.

VajraOfIndra:
> Ok, the "why" become a bit misleading, but the point of this at the end is still to provide the ability to use the hatcheck status without TH.

Good point.. as I don't use ThunderHub, I was not even aware that there was such a thing as a Amboss monitoring service.. this is definitely a useful bonus guide 💯

guide/bonus/lightning/ambossping.md
```diff
@@ -0,0 +1,166 @@
...
+  ```ini
+  #!/bin/bash
+  URL="https://api.amboss.space/graphql"
+  NOW=$(date -u +%Y-%m-%dT%H:%M:%S%z)
+  echo "Timestamp: ${NOW}"
+  SIGNATURE=$(/usr/local/bin/lncli signmessage "$NOW" | jq -r .signature)
+  echo "Signature: ${SIGNATURE}"
+  JSON="{\"query\": \"mutation HealthCheck(\$signature: String!, \$timestamp: String!) { healthCheck(signature: \$signature, timestamp: \$timestamp) }\", \"variables\": {\"signature\": \"$SIGNATURE\", \"timestamp\": \"$NOW\"}}"
+  echo "Sending ping..."
+  echo "$JSON" | curl -s -f --data-binary @- -H "Content-Type: application/json" -X POST --output /dev/null $URL
```
kristapsk, line 76: Can and should use Tor here.
```suggestion
  echo "$JSON" | torsocks curl -s -f --data-binary @- -H "Content-Type: application/json" -X POST --output /dev/null $URL
```

guide/bonus/lightning/ambossping.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Set the script
+
+* Edit the config file and make sure you have the following changes in the beginning.
+
+  ```sh
+  $ nano /home/ambossping/ping.sh
+  ```
+
+  ```ini
```
kristapsk, line 67: This is a shell script.
```suggestion
  ```sh
```

guide/bonus/lightning/ambossping.md
```diff
@@ -0,0 +1,166 @@
...
+  exitstatus=$?
+  if [ $exitstatus -ne 0 ]
+  then
+   echo "${NOW} - Error: ${exitstatus}"
+  else
+   echo "${NOW} - Ping sent"
+  fi
+  echo " "
+  ```
+
```
kristapsk, line 86: Should add `chmod +x /home/ambossping/ping.sh` somewhere here, otherwise `./ping.sh` below will not work.

guide/bonus/lightning/ambossping.md
satsHIVE: I think I could add it as an option. Some may find it unnecessary - either not really concerned by revealing to amboss or when behind a VPN for example. wdyt?

kristapsk: RaspiBolt guide defaults to LND being Tor only. No other guides should reveal it's true IP address to other guides by default.

satsHIVE: Fair enough, but I will have to test it before I change it here. Thanks for the review.

kristapsk: Of course, you should verify, not trust! :)

guide/bonus/lightning/ambossping.md
satsHIVE: fixed

guide/bonus/lightning/ambossping.md
```diff
@@ -0,0 +1,172 @@
...
+* Make the script executable (check by displaying the file name, it should have become green)
+
+  ```sh
+  chmod chmod +x /home/ambossping/ping.sh
```
VajraOfIndra, line 90:
```suggestion
  $ chmod chmod +x /home/ambossping/ping.sh
```

guide/bonus/lightning/ambossping.md
satsHIVE: Ok, I have tested it as suggested but I have to say that results are expected. The service turns into TOR service monitoring rather than node monitoring - quite a lot of false-positives where an OFFLINE status is reported, but the node is running just fine doing RBs and FWs. I will maybe append a notification about that to the guide.

kristapsk: Ok, then keep it as clearnet, with note about that. If user still wants to use Tor here, alternative approach is to add `torsocks /home/ambossping/ping.sh` to cron instead of `/home/ambossping/ping.sh`.

satsHIVE: Updated.

guide/bonus/lightning/ambossping.md
```diff
@@ -126,7 +126,7 @@ You should see something like this in terminal.
   ```
   This will execute the bash script every minute.
 
-  🚨 Setting up the cronjob like this will send the pings via clearnet - meaning amboss.space will know where the ping comes from. You can prevent revealing this information by either tunneling all traffic of your node via VPN or setup your cronjob to use TOR.
+  🚨 Setting up the cronjob like this will send the pings via clearnet - meaning amboss.space will know where the ping comes from. You can prevent revealing this information by either tunneling all traffic of your node via VPN or setup your cronjob to use TOR. Be aware that running the ping via TOR will cause some false-positives depending on the health of the TOR network status and availability, which will not represent the availability of your node.
```
VajraOfIndra, line 129:
```suggestion
  🚨 Setting up the cronjob like this will send the pings via clearnet - meaning amboss.space will know where the ping comes from. You can prevent revealing this information by either tunneling all traffic of your node via VPN or setup your cronjob to use Tor. Be aware that running the ping via Tor will cause some false-positives depending on the health of the Tor network status and availability, which will not represent the availability of your node.
```

[juts a nit](https://support.torproject.org/about/why-is-it-called-tor/) ;D

kristapsk:
ACK

guide/bonus/lightning/ambossping.md
```diff
@@ -0,0 +1,180 @@
...
+  ```
+  
+---
+
+## Test run
+
+* Run the script (while in "ambossping" user session)
+
+  ```sh
+  $ ./ping.sh
```
VajraOfIndra, line 100: The warning about revealing one's IP address to Amboss is placed later on for the cron job. But this step here would have already leaked the IP, right?

Is it possible to torify this step?

guide/bonus/lightning/ambossping.md
```diff
@@ -0,0 +1,180 @@
...
+ 
+  ```ini
+  * * * * * /home/ambossping/ping.sh >> /home/ambossping/ping.log
+  ```
+  This will execute the bash script every minute.
+
+  🚨 Setting up the cronjob like this will send the pings via clearnet - meaning amboss.space will know where the ping comes from. You can prevent revealing this information by either tunneling all traffic of your node via VPN or setup your cronjob to use Tor. Be aware that running the ping via Tor will cause some false-positives depending on the health of the Tor network status and availability, which will not represent the availability of your node.
+
+  ```ini
+  * * * * * torsocks /home/ambossping/ping.sh >> /home/ambossping/ping.log
```
blckbx, line 132:
```suggestion
  * * * * * /usr/bin/torsocks /home/ambossping/ping.sh >> /home/ambossping/ping.log
```
I think this needs to be specified.

satsHIVE: Worked for me even without it, but let's keep it that way. Thanks.

guide/bonus/lightning/ambossping.md
satsHIVE: fixed

guide/bonus/lightning/ambossping.md
```diff
@@ -97,7 +97,7 @@ Table of contents
 * Run the script (while in "ambossping" user session)
 
   ```sh
-  $ ./ping.sh
+  $ /usr/bin/torsocks /home/ambossping/ping.sh
```
blckbx, line 100: This is not necessary, only cron has no user-context and requires exact paths. 
You could leave this at `./ping.sh`.

satsHIVE: fixed

blckbx (approved):
LGTM

PR #1155, opened 2022-11-06
[UPDATE BONUS GUIDE] Add custom Banners for Fulcrum

DanielPX1:
#### What

Add custom banner option for Fulcrum users:)

### Why

To have fun and be creative with own server

#### How

By  following steps stated in the guide

#### Scope

- [x] test & review

#### Test & maintenance

By following the guide and testing on own device


guide/bonus/bitcoin/fulcrum.md
```diff
@@ -445,6 +445,56 @@ This way, you can connect the BitBoxApp or Electrum wallet also remotely, or eve
...
+
+```sh
+$ sudo nano /data/fulcrum/fulcrum.conf
+```
+
+* Specify path to banner at the end of your configuration file. Save and exit
+
+```sh
+# Banner path
+banner = /data/fulcrum/banner/banner.txt
```
twofaktor, line 485: change to:

`banner = /data/fulcrum/banner.txt`

twofaktor (changes requested):
Only one bug

guide/bonus/bitcoin/fulcrum.md
```diff
@@ -445,6 +445,56 @@ This way, you can connect the BitBoxApp or Electrum wallet also remotely, or eve
...
+    ____      __
+   / __/_  __/ /___________  ______ ___
+  / /_/ / / / / ___/ ___/ / / / __ `__ \
+ / __/ /_/ / / /__/ /  / /_/ / / / / / /
+/_/  \__,_/_/\___/_/   \__,_/_/ /_/ /_/
+
+
+server version: $SERVER_VERSION
+bitcoind version: $DAEMON_VERSION
+
```
twofaktor, line 472: is that extra final space necessary?

guide/bonus/bitcoin/fulcrum.md
```diff
@@ -445,6 +445,56 @@ This way, you can connect the BitBoxApp or Electrum wallet also remotely, or eve
...
+* Restart Fulcrum
+
+```sh
+$ sudo systemctl restart fulcrum.service
+```
+
+Now you can connect to your own server for ex. with Sparrow to see the banner
+
+<img src="../../../images/Fulcrum_Banner.png" width=50% height=50%>
```
twofaktor, line 496: Change to:
`Fulcrum banner`

twofaktor (commented):
Comments and suggestion of changes

guide/bonus/bitcoin/fulcrum.md
DanielPX1: Updated (y)

guide/bonus/bitcoin/fulcrum.md
DanielPX1: Yeah, you are right, its deleted now. But its just demonstration:)

guide/bonus/bitcoin/fulcrum.md
DanielPX1: Format does not support resizing, will have to resize outside GitHub

DanielPX1:
It is a very small improvement with only several commands, ready 4 test/review:)

guide/bonus/bitcoin/fulcrum.md
```diff
@@ -445,6 +445,52 @@ This way, you can connect the BitBoxApp or Electrum wallet also remotely, or eve
 
 * You should now be able to connect to your Fulcrum server remotely via Tor using your hostname and port 50002
 
+### Add banner to Fulcrum server (For fun!)
+
+You can get creative when making your server banner, for example creating your own [ASCII art](https://patorjk.com/software/taag/#p=display&f=Slant&t=Fulcrum). In [Fulcrum docs](https://github.com/cculianu/Fulcrum/blob/master/doc/fulcrum-example-config.conf) you can find additional info about making a banner in the "# Server banner text file - 'banner'" section.
+
+* Create and open "banner.txt" file inside Fulcrum directory
```
VajraOfIndra, line 452:
```suggestion
* Create and open `banner.txt` file inside Fulcrum directory
```

Afaik, I think we usually use code formattting for file names, same as what you did for fulcrum.conf at line 471

guide/bonus/bitcoin/fulcrum.md
```diff
@@ -445,6 +445,52 @@ This way, you can connect the BitBoxApp or Electrum wallet also remotely, or eve
...
+
+```sh
+$ sudo nano /data/fulcrum/banner.txt
+```
+
+* Paste your own creation into "banner.txt". Save and exit
```
VajraOfIndra, line 458:
```suggestion
* Paste your own creation into `banner.txt`. Save and exit
```

VajraOfIndra (changes requested):
Great fun addition! love it ❤️ 

LGTM, just a couple of nits.

Also, could you maybe add the 2 spaces in front of the code blocks that are below the bullet points..? (couldn't do it in the review as the three back ticks messes up with the GitHub comment formatting)

DanielPX1:
> Great fun addition! love it ❤️
> 
> LGTM, just a couple of nits.
> 
> Also, could you maybe add the 2 spaces in front of the code blocks that are below the bullet points..? (couldn't do it in the review as the three back ticks messes up with the GitHub comment formatting)

Glad to hear! changed the " to `, sorry however I am not sure I understand with the spaces?

VajraOfIndra:
> Sorry, I am not sure I understand the one with the spaces?

Like here, in front of all the lines that form a code block, we usually put two blank spaces to align with the start of the text in the bullet point above (like in the rest of the Fulcrum guide for example)
image

> PS: Just found out my Fulcrum.service did not start only with Electrs service being disabled, because it was also necessary to remove Electrs Nginx config. Should I add this info to the guide? Or will we count that user runs only one Electrs implementation?

Disable alone is not enough afail, it simply stops Electrs to be started when booting/rebooting... If a user has Electrs running, they would also need to stop the service. So both stop and disable are needed. 

Regarding the nginx reverse proxy, are you using it for Fulcrum? if not, why not? (just curious) I'm not sure to understand why removing the nginx reverse proxy config file is ncessary for Fulcrum to start.

DanielPX1:
> > Sorry, I am not sure I understand the one with the spaces?
> 
> Like here, in front of all the lines that form a code block, we usually put two blank spaces to align with the start of the text in the bullet point above (like in the rest of the Fulcrum guide for example) image
> 
> > PS: Just found out my Fulcrum.service did not start only with Electrs service being disabled, because it was also necessary to remove Electrs Nginx config. Should I add this info to the guide? Or will we count that user runs only one Electrs implementation?
> 
> Disable alone is not enough afail, it simply stops Electrs to be started when booting/rebooting... If a user has Electrs running, they would also need to stop the service. So both stop and disable are needed.
> 
> Regarding the nginx reverse proxy, are you using it for Fulcrum? if not, why not? (just curious) I'm not sure to understand why removing the nginx reverse proxy config file is ncessary for Fulcrum to start.

ok should be fixed!

Fulcrum has it integrated If I am aware, so it is not necessary to use it with Fulcrum. But my service could not start until I removed nginx config for electrs... Not sure why it was issue, maybe would need to inspect it more

VajraOfIndra:
> ok should be fixed!

👍 
 
> Fulcrum has it integrated If I am aware, so it is not necessary to use it with Fulcrum. But my service could not start until I removed nginx config for electrs... Not sure why it was issue, maybe would need to inspect it more

Hum.. so if Fulcrum wanted to use port 50002 and it was still being used by the nginx reverse proxy for Electrs.. maybe there was a conflict that prevented Fulcrum to start.

Adding a section for Electrs user at the start might be helpful yes.. like stop and disable the Electrs service and delete (or simply backup, changing the .conf to .conf.bak would be enough to disable the reverse proxy) the nginx Electrs conf file and reload nginx.

VajraOfIndra (approved):

VajraOfIndra:
Let's go to master 🎉

PR #1157, opened 2022-11-07
Update version in the Run LND section

Bitcoinite:
The LND section was still showing v0.14.2.


VajraOfIndra (approved):
Well spotted!

Now v0.14.4 in November 2021.. we have a proof of a parallel universe ;)

PR #1160, opened 2022-11-07
Update web-app.md

xmready:
Change `localhost` to `127.0.0.1` in the RTL config file to avoid the following errors:
```
$ node rtl
[11/7/2022, 3:34:35 PM] INFO: RTL => Server is up and running, please open the UI at http://localhost:3000 or your proxy configured url.

[11/7/2022, 3:35:05 PM] ERROR: Common => Error in Channel Backup for xxxxx: connect ECONNREFUSED ::1:8080

[11/7/2022, 3:35:05 PM] ERROR: GetInfo => Get Info Error: {"name":"RequestError","message":"Error: connect ECONNREFUSED ::1:8080","cause":{"errno":-111,"code":"ECONNREFUSED","syscall":"connect","address":"::1","port":8080},"error":{"errno":-111,"code":"ECONNREFUSED","syscall":"connect","address":"::1","port":8080},"options":{"url":"https://localhost:8080/v1/getinfo","rejectUnauthorized":false,"json":true,"headers":{},"method":"GET","qs":{},"simple":true,"resolveWithFullResponse":false,"transform2xxOnly":false}}
```
Here is a link to an issue on the RTL GitHub describing the same problem with this change posted as the solution
https://github.com/Ride-The-Lightning/RTL/issues/1120

Here is a link to an issue I created in the LND GitHub trying to troubleshoot it before I figured it out. 
https://github.com/lightningnetwork/lnd/issues/7125

### DIFF
--- a/guide/lightning/web-app.md
+++ b/guide/lightning/web-app.md
@@ -164,7 +164,15 @@ Now we take the sample configuration file and add change it to our needs.
     "macaroonPath": "/home/rtl"
     "configPath": "/data/lnd/lnd.conf"
   ```
+  
+* Change `localhost` to `127.0.0.1` on the following lines to avoid errors
 
+  ```ini
+    "lnServerUrl": "https://127.0.0.1:8080"
+    "swapServerUrl": "https://127.0.0.1:8081"
+    "boltzServerUrl": "https://127.0.0.1:9003"
+  ```
+  
 * Save and exit
 
 


blckbx (commented):
LGTM
always had it set to 127.0.0.1 out of habit

VajraOfIndra:
LGTM

why did `localhost` cause an issue? and what's the link with ipv6..? (just curious to understand the issue and why changing to 127.0.0.1 solves it :)

VajraOfIndra (approved):
LGTM, utACK

why did localhost cause an issue? and what's the link with ipv6..? (just curious to understand the issue and why changing to 127.0.0.1 solves it :)

xmready:
> why did localhost cause an issue?

Appears to be a problem with RTL that needs fixing on their end. If you want to know more than that, maybe you can dig deeper using the error codes posted in the OP.

> and what's the link with ipv6..?

Not sure what you are asking here

VajraOfIndra:
> Not sure what you are asking here

In the Issue you linked, I don't understand the link between changing localhost to 127.0.0.1 and the rest of the comment about ipv6 loopback:

> Was able to solve the issue by changing the RTL config lnServerUrl from "localhost" to 127.0.0.1. I logged the errRes in the catch block and it showed address: '::1' which is the ipv6 loopback. I guess my LND was only listening on ipv4.

xmready:
> In the Issue you linked, I don't understand the link between changing localhost to 127.0.0.1 and the rest of the comment about ipv6 loopback:

You can see in both my error messages and the one from the issue your referencing, it fails on address `::1` which is the ipv6 loopback. Apparently using `localhost` in the RTL conf tries the ipv6 loopback address instead of the ipv4 address. It seems that LND only listens on the ipv4 address, which is why this fix works.

VajraOfIndra:
> > In the Issue you linked, I don't understand the link between changing localhost to 127.0.0.1 and the rest of the comment about ipv6 loopback:
> 
> You can see in both my error messages and the one from the issue your referencing, it fails on address `::1` which is the ipv6 loopback. Apparently using `localhost` in the RTL conf tries the ipv6 loopback address instead of the ipv4 address. It seems that LND only listens on the ipv4 address, which is why this fix works.

I did read the issue but didn't understand what it meant. I think I get it now, thanks for taking the time to explain.

PR #1162, opened 2022-11-09
Improve the Upgrade section of the regolancer bonus guide

Bitcoinite:
Changed the upgrade command from a specific version to the latest one. This makes it easier for users to upgrade, since they don't need to check the exact version number of the latest release.

### DIFF
--- a/guide/bonus/lightning/regolancer.md
+++ b/guide/bonus/lightning/regolancer.md
@@ -126,16 +126,14 @@ To run it continuously, you will need to run it in a loop or a cron.
 
   ```sh
   $ sudo su - regolancer
-  $ go install github.com/rkfg/regolancer@v1.6.x 
+  $ go install github.com/rkfg/regolancer@latest
   ```
-
-Please replace the v.1.6.x part with the version you would like to install.
 
 * Confirm if the upgrade was successfull
 
   ```sh
   $ go/bin/regolancer -v
-  Regolancer v1.6.3, built with go1.19.2
+  Regolancer v1.8.0, built with go1.19.2
   Source: https://github.com/rkfg/regolancer
   ```
 
--- a/guide/bonus/lightning/regolancer.md
+++ b/guide/bonus/lightning/regolancer.md
@@ -76,8 +76,9 @@ Table of contents
 * Create a working copy of the sample config file. You can use either .json or .toml configs, up to your preference.  
 
   ```sh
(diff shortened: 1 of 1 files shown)


blckbx (commented):
LGTM

for first time setup: do we want to use `latest` here as well? 
But then I guess we need to update the guide for the version check anyway.

blckbx (commented):
utACK

twofaktor (changes requested):
Line 65: `GOARCH=arm64 go install github.com/rkfg/regolancer@latest`

` For RPi run it as GOARCH=arm64 go install github.com/rkfg/regolancer@latest if you run a 64 bit system (and you should!)`

Ref: https://github.com/rkfg/regolancer#installation

what about an 'Autostart on boot'  section through regolancer.service file on systemd?

satsHIVE:
This guide is in the making right now. ;)

twofaktor:
Line 67: https://github.com/raspibolt/raspibolt/pull/1162/files#diff-2be0a3351d69f05b8f7c1a90a3592b0679ca82bff853936485457b67baa09758L67

Change to --> Note: Adjust the "regolancer@v1.9.0" part from the commands below to the actual version you have installed.

satsHIVE:
> Line 67: https://github.com/raspibolt/raspibolt/pull/1162/files#diff-2be0a3351d69f05b8f7c1a90a3592b0679ca82bff853936485457b67baa09758L67
> 
> Change to --> Note: Adjust the "regolancer@v1.9.0" part from the commands below to the actual version you have installed.

I guess this should not be adjusted but rather removed. It no longer makes sense since started using "@latest".

Bitcoinite:
> > Line 67: https://github.com/raspibolt/raspibolt/pull/1162/files#diff-2be0a3351d69f05b8f7c1a90a3592b0679ca82bff853936485457b67baa09758L67
> > Change to --> Note: Adjust the "regolancer@v1.9.0" part from the commands below to the actual version you have installed.
> 
> I guess this should not be adjusted but rather removed. It no longer makes sense since started using "@latest".

I removed the note altogether since we don't need it anymore.

twofaktor (approved):
LGTM! utACK

Bitcoinite:
> One more nit: Line 80:
> 
> ```shell
> $ cp /home/regolancer/go/pkg/mod/github.com/rkfg/regolancer@v1.6.3/config.json.sample /home/regolancer/config.json
> ```
> 
> still references v1.6.3 instead of 1.9.0. After that it's gtg!

Done. Also added the note I just deleted from line 67, since we need a version number here.

blckbx (approved):
~~utACK~~
tACK amd64

PR #1166, opened 2022-11-11
Add missing LNDg entry in bonus guides index

VajraOfIndra:
#### What

Adds missing LNDg entry in bonus guide index

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix


blckbx (approved):
🚀

PR #1172, opened 2022-11-13
[UPDATE BONUS GUIDE] Fulcrum: Add TCP support

DanielPX1:
#### What

Add TCP support for Fulcrum users, as some wallets do not support TCP over Tor and for other reasons.

### Why

More ways to connect with your wallet

#### How

Adding following steps stated in the guide

#### Scope

- [ ] Review

#### Test & maintenance

Following the guide and reporting issues, improvements

PR #1184, opened 2022-11-20
Update fulcrum.md

phil71x:
Correct a typo in the fulcrum.conf file

#### What

### Why

Why is this change important?

#### How

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

#### Animated GIF (optional)

### DIFF
--- a/guide/bonus/bitcoin/fulcrum.md
+++ b/guide/bonus/bitcoin/fulcrum.md
@@ -281,7 +281,7 @@ Now that Fulcrum is installed, we need to configure it to run automatically on s
   bitcoind_timeout = 600
   bitcoind_clients = 1
   worker_threads = 1
-  deb_mem = 1024.0
+  db_mem = 1024.0
   
   # 4GB RAM (default)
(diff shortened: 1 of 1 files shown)


DanielPX1 (approved):
Looks good, not sure how that slipped in there...

From fulcrum config, line 650:
Max RocksDB Memory in MiB - 'db_mem' - DEFAULT: 512.0
[...]
db_mem = 512.0

source: https://github.com/cculianu/Fulcrum/blob/master/doc/fulcrum-example-config.conf

twofaktor (approved):
Nice catch! Thanks! utACK!

We have another PR #1172 opened

Would we approve before this PR and then the other? wdyt?

DanielPX1:
> Nice catch! Thanks! utACK!
> 
> We have another PR #1172 opened
> 
> Would we approve before this PR and then the other? wdyt?

Yeah, agree would be the best

PR #1201, opened 2022-12-02
[NEW BONUS GUIDE] BTCPay Server

nicheosala:
#### What

This guide allows to install BTCPay Server on our Raspibolt.

### Why

BTCPay Server allows to manage Bitcoin and Lightning payments.
It can be useful, for example, to receive donations on a blog.

#### How

The installation of BTCPay Server is possible thanks to the detailed installation guide of this pull request.

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

Any testing of the following guide, along with bug reporting, would be really appreciated.


ericsupreme:
i 'd like to test this out on my raspibolt, you can send me the guide!!

satsHIVE:
Awesome! This has been asked a few times in the Telegram group as well. You can probably find testers of the guide there as well. :)
Can you add a header image/screenshot as the other guides have? Just above TOC.

kristapsk:
Looks it's time for me to finally look into writing [SatSale](https://github.com/SatSale/SatSale) guide for RaspiBolt. :) BTCPay Server has a lot more features (it's great!), but SatSale is more lightweight alternative, which might be even more important on less powerful hardware, like Raspberry Pi.

kristapsk:
P.S. Will try to test this in a few days.

guide/bonus/lightning/btcpayserver.md
```diff
@@ -0,0 +1,308 @@
...
+* Exit PostgreSQL
+
+  ```sh
+  postgres=# \q
+  ```
+
+### NBXplorer configuration
+
+  ```sh
+  $ nano /home/btcpay/.nbxplorer/Main/settings.config
```
ericsupreme, line 124: hmmmm not sure if this is the place but following the guide up to this point and running "nano /home/btcpay/.nbxplorer/Main/settings.config" says directory does not exist. i confirmed running ls -la i don't have a .nbxplorer directory. any ideas? again sorry if this is the wrong place.

nicheosala: Thank you for your feedback! I've just updated the guide: at the beginning, I forgot to make a directory called "src" where to place all the source code. This may be the problem. Please, try following the guide from scratch and let me know.

guide/bonus/lightning/btcpayserver.md
```diff
@@ -0,0 +1,316 @@
...
+
+### Install .NET 6.0 SDK
+
+With user "btcpay", we install .NET 6.0 SDK. This software is used to build and execute BTCPay Server.
+
+* Install .NET SDK 6.0
+
+  ```sh
+  $ curl https://dot.net/v1/dotnet-install.sh
+  $ chmod +x ./dotnet-install.sh
```
kristapsk, line 61: This will not work, `curl` will not download to a file by default. Simplest way would be to just use `wget` here instead.

guide/bonus/lightning/btcpayserver.md
BoyvdR: I've followed the steps and had the same issue. As an experiment I run the run.sh script, it made the config file and exited with an error. After adding the configuration as in the guide it NBXplorer workt... 
Had the same issue with btcpayserver later on, running the run.sh script fixed the issue there aswell

nicheosala: A thousand thanks! I forgot to point out that both NBXplorer and BTCPay Server need to be run once, in order to generate the default configuration files. I've just pushed a commit that should fix this issue.

hdlslm: @BoyvdR thank you. i will give it another try later on. @nicheosala again great work!!

hdlslm: running the run.sh worked for me as well!!! thanks

hdlslm (commented):
ok made it to the btcpay server login page and succesfully logged in.  i had to run run.sh for nbxplorer as well as btcpay per @BoyvdR above. i just  had to add reverse proxy via nginx but after i login i get an warning message that says "BTCPay is expecting you to access this website from http://192. blah.blah.blah/. If you use a reverse proxy, please set the X-Forwarded-Proto header to http" with this link : https://docs.btcpayserver.org/FAQ/Deployment/#cause-3-btcpay-is-expecting-you-to-access-this-website-from
but it didn't seem to work work. i added it to the nginx.conf so still working on it. doesn't seem to effect anything. i can just x out the warning.

jtymoszczuk:
This is great! not that its advisable, but would this setup work with a neutrino supported node? I'm just looking to test it out first on a neutrino node.

guide/bonus/lightning/btcpayserver.md
```diff
@@ -0,0 +1,327 @@
...
+  $ nano /home/btcpay/src/NBxplorer/run.sh
+  ```
+
+  Change `dotnet` to `/home/btcpay/.dotnet/dotnet`, save and exit.
+
+
+* Create PostgreSQL database for NBXplorer
+
+  ```sh
+  $ sudo -u postgres psql
```
jtymoszczuk, line 115: i can't sudo. I don't believe user btcpay was ever given sudo privileges?

hdlslm: > i can't sudo. I don't believe user btcpay was ever given sudo privileges?

install btcpay as admin. that was the step "prior" to creating btcpay user per guide

guide/bonus/lightning/btcpayserver.md
```diff
@@ -0,0 +1,327 @@
...
+### Install .NET 6.0 SDK
+
+With user "btcpay", we install .NET 6.0 SDK. This software is used to build and execute BTCPay Server.
+
+* Install .NET SDK 6.0
+
+  ```sh
+  $ wget https://dot.net/v1/dotnet-install.sh
+  $ chmod +x ./dotnet-install.sh
+  $ ./dotnet-install.sh --channel 6.0
```
kristapsk, line 62: Is this the right way to do .Net install? Gives warning:
```
dotnet-install: Note that the intended use of this script is for Continuous Integration (CI) scenarios, where:
dotnet-install: - The SDK needs to be installed without user interaction and without admin rights.
dotnet-install: - The SDK installation doesn't need to persist across multiple CI runs.
dotnet-install: To set up a development environment or to run apps, use installers rather than this script. Visit https://dotnet.microsoft.com/download to get the installer.
```
There are alternative [manual install](https://learn.microsoft.com/en-us/dotnet/core/install/linux-scripted-manual#manual-install) instructions from Microsoft.

Also, probably we want system wide install of .Net runtime, so that potentially other .Net apps can be ran in the future.

nicheosala: BTCPay requires dotnet sdk 6.x. Installing dotnet via apt, using the Debian repository, I wasn't able to install it.
I totally agree that a system-wide installation via apt would be much better, but I couldn't figure out how to achieve this on the Raspberry Pi, so I decided to publish the "manual" installation guide for dotnet.
If you can install dotnet sdk 6.0 system-wide using apt, please, create a pull request.

kristapsk: Haven't tested with Raspbian (don't have working RPi with me now), but with Microsoft package repository works like a charm on Ubuntu Server. And you can even have dotnet 6.0 and 7.0 in parallel. Somebody should try Debian link I posted above.

guide/bonus/lightning/btcpayserver.md
```diff
@@ -0,0 +1,327 @@
...
+  HiddenServiceDir /var/lib/tor/hidden_service_btcpay/
+  HiddenServiceVersion 3
+  HiddenServicePort 80 127.0.0.1:23000
+  ```
+
+  Update Tor configuration changes and get your connection address.
+
+  ```sh
+  $ sudo systemctl reload tor
+  $ sudo cat /var/lib/tor/hidden_service_rtl/hostname
```
greenart7c3, line 320: Should be /var/lib/tor/hidden_service_btcpay/hostname

greenart7c3 (changes requested):

guide/bonus/lightning/btcpayserver.md
nicheosala: Fixed. Thank you!

Mynima:
I realise this has been open a while. Just wondering if you mind adding an uninstall section at the end?

hdlslm:
btcpay

so again i am stumped. i was able to get to login page for like a week and then i get the: " This page isn’t working
didn’t send any data. ERR_EMPTY_RESPONSE. "error i tried multiple browsers , tried another computer. everything seems to be running (see attached pic) i setup ngnix  reverse proxy the same as the other installs i.e. electrum etc. i tried removing it and still nothing. any suggestions?

hdlslm:
> btcpay
> 
> so again i am stumped. i was able to get to login page for like a week and now i get: " This page isn’t working didn’t send any data. ERR_EMPTY_RESPONSE. "error i tried multiple browsers , tried another computer. everything seems to be running (see attached pic) i setup ngnix reverse proxy the same as the other installs i.e. electrum etc. i tried removing it and still nothing. any suggestions?

i am able to get o the login page via tor not sure why i can't via https anymore.

nicheosala:
> > btcpay
> > so again i am stumped. i was able to get to login page for like a week and now i get: " This page isn’t working didn’t send any data. ERR_EMPTY_RESPONSE. "error i tried multiple browsers , tried another computer. everything seems to be running (see attached pic) i setup ngnix reverse proxy the same as the other installs i.e. electrum etc. i tried removing it and still nothing. any suggestions?
> 
> i am able to get o the login page via tor not sure why i can't via https anymore.

@hdlslm The errors you are reporting are not related to the guide. Please, post them elsewhere (Reddit), so as not to cause confusion in this thread.

guide/bonus/lightning/btcpayserver.md
```diff
@@ -0,0 +1,327 @@
...
+  ```sh
+  $ git clone https://github.com/dgarage/NBXplorer
+  $ cd NBXplorer
+  $ ./build.sh
+  ```
+
+* Modify NBXplorer run script
+
+  ```sh
+  $ nano /home/btcpay/src/NBxplorer/run.sh
```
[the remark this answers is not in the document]
nicheosala: The command I wrote specifies the absolute path to the file, so it shouldn't be wrong. Could you please check that the `run.sh` file actually exists in the directory? Did you build NBXplorer with the command just above?

guide/bonus/lightning/btcpayserver.md
```diff
@@ -0,0 +1,327 @@
...
+  User=btcpay
+
+  Type=simple
+  Restart=always
+  TimeoutSec=120
+  RestartSec=30
+  KillMode=process
+
+  [Install]
+  WantedBy=multi-user.target
```
greenart7c3, line 286: If you restart the service its not killing the proccess.
The service file should be something like (same applies to the nbxexplorer service):

[Unit]
Description=BTCPay Server
Requires=nbxplorer.service
After=nbxplorer.service

[Service]
WorkingDirectory=/home/btcpay/src/btcpayserver
ExecStart=/home/btcpay/src/btcpayserver/run.sh
User=btcpay

Type=simple
PIDFile=/run/btcpayserver/btcpayserver.pid
TimeoutSec=120
RestartSec=30

[Install]
WantedBy=multi-user.target

greenart7c3 (changes requested):

hdlslm:
ok now how do we connect our lightning node to btcpay? the use internal node is greyed out!! thanks in advance

guide/bonus/lightning/btcpayserver.md
jesusangelm: I'm currently trying this in Raspberry Pi OS and the package manager method doesn't work.


notice the important info on [Microsoft page](https://learn.microsoft.com/en-us/dotnet/core/install/linux-debian):
> Package manager installs are only supported on the x64 architecture. Other architectures, such as Arm, must install .NET by some other means such as with Snap, an installer script, or through a manual binary installation.


The reason the package manager method works for @kristapsk  is probably because he is using Ubuntu, which has the snap package manager with dotnet runtime available.

kristapsk: I don't use snap, just Ubuntu apt packages from Microsoft.

PR #1203, opened 2022-12-02
Change the default base fee and fee rate for lnd

Bitcoinite:
This PR changes the default base fee to 0 to support the #zerobasefee initiative. This will make pathfinding a linear mathematical problem that is much easier to solve. For further information see here:
https://arxiv.org/pdf/2107.05322.pdf
http://www.rene-pickhardt.de

Additionally, this PR sets the default fee rate for newly opened channels to 2500 ppm. This will prohibit channels opened to sink nodes from draining instantly after opening.

### DIFF
--- a/guide/lightning/lightning-client.md
+++ b/guide/lightning/lightning-client.md
@@ -193,8 +193,8 @@ To improve the security of your wallet, check out these more advanced methods:
   wallet-unlock-allow-create=true
 
   # Channel settings
-  bitcoin.basefee=1000
-  bitcoin.feerate=1
+  bitcoin.basefee=0
+  bitcoin.feerate=2500
   minchansize=100000
   accept-keysend=true
   accept-amp=true


guide/lightning/lightning-client.md
```diff
@@ -193,8 +193,8 @@ To improve the security of your wallet, check out these more advanced methods:
   wallet-unlock-allow-create=true
 
   # Channel settings
-  bitcoin.basefee=1000
-  bitcoin.feerate=1
+  bitcoin.basefee=0
```
kristapsk, line 196: Don't think we should enforce this. I personally don't like to route anything that doesn't pay me at least 1 sat. Better alternative would be to add commented out `bitcoin.basefee=0` line with a description in comment why users should consider enabling it.

vicariousdrama:
NACK.
While I participate in zerobasefee, I prefer to manually assign this on a per channel basis and actually default channel creations with a higher then standard fees to help avoid immediate channel drains under certain scenarios. 

Would prefer sticking to the defaults

twofaktor:
NACK

I would prefer this:
```
# Fee settings - default LND base fee = 1000 (mSat), default LND fee rate = 1 (ppm) #You can choose whatever you want i.e ZeroFeeRouting (0,0)
#bitcoin.basefee=0
#bitcoin.feerate=0
```

kristapsk:
Closing, as there are multiple NACKs, and no ACKs.

PR #1204, opened 2022-12-02
Update lightning-client.md

### LINKED ISSUE #1115: [FEATURE REQ] Set tlsautorefresh=true in default lnd.conf (pros/cons?)
### What & why

According to the [official documentation](https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf):
> ; If set, then all certs will automatically be refreshed if they're close to 
> ; expiring, or if any parameters related to extra IPs or domains in the cert 
> ; change.
> ; tlsautorefresh=true

Setting it to true would prevent potential issues and downtimes for users once their certificate expires.

Raspiblitz has this option as true in their [default config file](https://github.com/rootzoll/raspiblitz/blob/v1.8/home.admin/config.scripts/lnd.install.sh).

Any reasons why we should NOT do this?

If we decide to include this option in the default config for the guide, we probably need to also look into related TLS certif options, e.g. `tlsdisableautofill`.

### How

Edit default `lnd.conf` in [this section of the guide](https://raspibolt.org/guide/lightning/lightning-client.html#configuration).

### More context

Check the short discussion in the Telegram group with @vicariousdrama starting at [this post](https://t.me/raspibolt/13653).

vicariousdrama:
Set tlsautorefresh=true.
Fixes #1115

#### What

Add setting to auto refresh TLS cert

### Why

Certs expire (I think setting is 1 year after creation).  And when they do remote queries and RPC calls like unlock wallet can fail.

#### How

Configuration change added to lnd.conf

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

Fixes #1115

#### Test & maintenance

To test this issue requires running LND for some time and waiting for cert expiration and then check if recreated.

### DIFF
--- a/guide/lightning/lightning-client.md
+++ b/guide/lightning/lightning-client.md
@@ -193,7 +193,13 @@ To improve the security of your wallet, check out these more advanced methods:
   wallet-unlock-allow-create=true
 
   # Automatically regenerate certificate when near expiration
-  tlsautorefresh=1
+  tlsautorefresh=true
+  # Do not include the interface IPs or the system hostname in TLS certificate.
+  tlsdisableautofill=true
+  # Explicitly define any additional domain names for the certificate that will be created.
+  # tlsextradomain=raspibolt.local
+  # tlsextradomain=raspibolt.public.domainname.com
+  
 
   # Channel settings
   bitcoin.basefee=1000
--- a/guide/lightning/lightning-client.md
+++ b/guide/lightning/lightning-client.md
@@ -192,6 +192,9 @@ To improve the security of your wallet, check out these more advanced methods:
   wallet-unlock-password-file=/data/lnd/password.txt
   wallet-unlock-allow-create=true
 
+  # Automatically regenerate certificate when near expiration
+  tlsautorefresh=1
+
   # Channel settings
   bitcoin.basefee=1000
   bitcoin.feerate=1


twofaktor (approved):
tACK! TLS certificates were automatically regenerated! good addon! thanks!

Captura98723747242983
Captura98723747242

guide/lightning/lightning-client.md
```diff
@@ -192,6 +192,9 @@ To improve the security of your wallet, check out these more advanced methods:
   wallet-unlock-password-file=/data/lnd/password.txt
   wallet-unlock-allow-create=true
 
+  # Automatically regenerate certificate when near expiration
+  tlsautorefresh=1
```
twofaktor, line 196: Change to `tlsautorefresh=true` to standardize with the rest of the guide

Should we add `tlsdisableautofill=true` as well??
```sh
# Do not include the interface IPs or the system hostname in TLS certificate, use first --tlsextradomain as Common Name instead, if set.
tlsdisableautofill=true
 ```

**Review by twofaktor — CHANGES_REQUESTED** [review]

File: guide/lightning/lightning-client.md (hunk 1 above)
Comment (vicariousdrama) [reply guide/lightning/lightning-client.md:196]: I'm not opposed to that and will make the adjustment.  Here's some context for posterity for anyone curious.

Doing this will default to only using 127.0.0.1 and ::1 in the cert that is created.

This means it'll exclude external facing ipv4 and ipv6 addresses on your local network, along with the hostname (the default operation).  I make use of the auto stuff for my own needs on some, and explicitly declare on others.

Overall, setting `tlsdisableautofill=true` is a greater security posture since it avoids leaking data, and anything essential should be explicitly defined anyways.

For those that want to see what their current certificate has, you can run the following command and review the X509v3 Subject Alternative Name field
```sh
openssl x509 -in /data/lnd/tls.cert -text -noout
```

twofaktor:
in which case do we use this parameter? could you give an example?

```
# Explicitly define any additional domain names for the certificate that will be created.
# tlsextradomain=raspibolt.local
# tlsextradomain=raspibolt.public.domainname.com
```

guide/lightning/lightning-client.md
twofaktor: Good!

Capturaet54t45676u5

vicariousdrama:
> in which case do we use this parameter? could you give an example?
> 
> ```
> # Explicitly define any additional domain names for the certificate that will be created.
> # tlsextradomain=raspibolt.local
> # tlsextradomain=raspibolt.public.domainname.com
> ```

Yes. If you want to use REST access to your node remotely, then you need to have the certificate created to support the addressable point, whether that is by IP address or domain name.  One case for this could be BTCPay server on a different system linking to the Lightning service.  Another scenario is accessing the node via a wallet, like Zeus or BlueWallet.  You can create the certificate and also include the onion address as an extra domain, though specifics on how that is setup may or may not need SSL with Tor etc..

twofaktor (approved):
Tested ACK!

blckbx (approved):
tACK

PR #1223, opened 2023-01-06
Fix broken image link in updated BitBox guide

sutterseba:
#### What

Image link in updated BitBox guide is broken

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

#### Test & maintenance

Tested it offline now and it should work... 

#### Animated GIF (optional)

giphy


blckbx (approved):
Difficult to verify but LGTM

PR #1231, opened 2023-01-22
[UPDATE BONUS GUIDE] Sparrow Terminal

DanielPX1:
#### What

Changing small mistake in the path when removing Sparrow Terminal

### Why

Avoid error/confusion for beginners in linux


twofaktor:
LGTM! Good catch!

twofaktor (approved):

PR #1233, opened 2023-01-23
Change bitcoind systemd service to use exec process type

russelldavies:
#### What

With the present configuration, the bitcoind service outputs its log output to a file. This change alters the systemd service definition so that output is sent to stdout which can be managed by systemd journal.

### Why

Having systemd journal manage the log output is more consistent with how logs from other services are managed.

#### How

The `bitcoind.service` file was altered to use an exec process type configuration. Any references to the debug.log file were replaced with the appropriate journal call.

#### Scope

- [x] moderate change to core configuration

#### Test & maintenance

This change can be easily tested by editing the `bitcoind.service` file with the changes, reloading systemd (`sudo systemctl daemon-reload`), and then restarting the bitcoin service (`sudo systemctl restart bitcoind`).

There is no ongoing maintenance.

### DIFF
--- a/guide/bitcoin/bitcoin-client.md
+++ b/guide/bitcoin/bitcoin-client.md
@@ -268,12 +268,6 @@ Still logged in as user "bitcoin", let's start "bitcoind" manually.
 
 * Once everything looks ok, stop "bitcoind" with `Ctrl-C`
 
-* Grant the "bitcoin" group read-permission for the debug log file:
-
-  ```sh
-  $ chmod g+r /data/bitcoin/debug.log
-  ```
-
 * Exit the “bitcoin” user session back to user “admin”
 
   ```sh
@@ -312,7 +306,7 @@ We use “systemd“, a daemon that controls the startup process using configura
   # Service execution
   ###################
 
-  ExecStart=/usr/local/bin/bitcoind -daemon \
+  ExecStart=/usr/local/bin/bitcoind -nodebuglogfile \
                                     -pid=/run/bitcoind/bitcoind.pid \
                                     -conf=/home/bitcoin/.bitcoin/bitcoin.conf \
                                     -datadir=/home/bitcoin/.bitcoin \
@@ -320,7 +314,7 @@ We use “systemd“, a daemon that controls the startup process using configura
 
   # Process management
   ####################
-  Type=forking
+  Type=exec
   PIDFile=/run/bitcoind/bitcoind.pid
   Restart=on-failure
   TimeoutSec=300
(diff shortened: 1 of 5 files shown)


Mynima:
Does look on the surface like this makes sense. However, I'm not massively experienced so forgive me a if I ask a couple of (most likely uneducated) questions:
- Is there any benefit (other than consistency) for running with the "nodebuglogfile" option rather than running the traditional "daemon" and logging as the guide already has ?
- I couldn't find other details elsewhere about the "nodebuglogfile" option, can you maybe share more info about is (I guess the name makes it clear) but why is it used as an alternative to the "daemon" option in your code? (I cannot see anything to say they are mutually exclusive)
- Are there any dependencies on the current setup that could be affected by removing the "daemon" option?

Anyway, thanks in advance for humoring me with any responses.

russelldavies:
By default, `bitcoind` runs as a foreground process unless the `-daemon` option is passed. Also by default, output is written to a logfile called `debug.log` unless the `-nodebuglogfile` option is passed. From the man page:

```
  -debuglogfile=<file>
       Specify location of debug log file. Relative paths will be prefixed by a
       net-specific datadir location. (-nodebuglogfile to disable;
       default: debug.log)
```

From what I can see in the [logging code](https://github.com/bitcoin/bitcoin/blob/e9262ea32a6e1d364fb7974844fadc36f931f8c6/src/logging.cpp#L460) the log file doesn't continue to grow forever but its size may exceed what you'd prefer without configuration. By disabling this file and letting systemd journal store the output you can get the benefits of a single interface and system to manage your logs, including storage sizes and rotation.

I can't see how removing the `daemon` option would affect the current setup. There were a few references to looking at the `debug.log` for troubleshooting but I have replaced those references with using `journalctl` instead.

Mynima:
Excellent, I appreciate the additional information/comments. From my own research (though limited) I'd agree I couldn't see any issue with moving this logging process into line with the way it is handled for the other installation guides.

kristapsk:
Arguments seem logical, but, then again, everybody has used to have `bitcoind` log output go to `debug.log`.

Are there other Linux Bitcoin distributions doing this?

twofaktor:
> Arguments seem logical, but, then again, everybody has used to have `bitcoind` log output go to `debug.log`.
> 
> Are there other Linux Bitcoin distributions doing this?

Tested on MiniBolt and running without problems...

https://minibolt.info/guide/bitcoin/bitcoin-client.html

https://github.com/twofaktor/minibolt/pull/23

kristapsk:
> Tested on MiniBolt and running without problems...

I didn't say this will not work. My argument was that doing this completely differently than others may not be worth. But maybe it is worth and maybe we are not the first ones doing this way. This is worth small research and discussing.

russelldavies:
@blckbx:
* If you want to use syslog it's better to configure systemd to forward its journal logs to syslog rather than individually configuring a service to do this.
* `exec` is a newer [execution type](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=) than `simple` and preferred as subsequent dependent units won't start unless the process starts without issue.

PR #1271, opened 2023-04-29
Overview

nonamecrypto:
#### What

Added two changes to the overview:

1) an indicator to show if there are pending updates and if there are, how many; and
2) the  local network IP address of the Raspibolt.

### Why

The updates indicator is important, so that it is clear if there are updates pending, which should be applied asap. The second change is a convenience change.

#### How

Updated the bash file.

#### Scope

[ ] significant change to core configuration
[x] independent bonus guide
[ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

Both changes are visible in the overview. It should not require maintenance, but if so I'm willing to do it.

#### Animated GIF (optional)

change-screenshot

### DIFF
--- a/resources/20-raspibolt-welcome
+++ b/resources/20-raspibolt-welcome
@@ -72,6 +72,9 @@ ${color_yellow}━━━━━━━━━━━━━━━━━━━━━�
 printf "%0.s#" {1..40}
 echo -ne '\r### Loading System data \r'
 
+# get local network ip address
+local_ip=$(ip addr show | grep -w inet | grep -v 127.0.0.1 | awk '{print $2}' | cut -d/ -f1)
+
 # get available update
 updates=$((`sudo apt update &>/dev/null && sudo apt list --upgradable 2>/dev/null | wc -l`-1))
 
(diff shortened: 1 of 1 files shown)


zynos:
The current system overview script is now this one:
https://github.com/raspibolt/raspibolt-pulse

nonamecrypto:
@Mynima Yes, it is possible to use set it up without sudo. A restricted user van be created who can only run specific commands using sudo without needing a password for those commands. This is achieved by editing the sudoers file. This does make the setup more complicated.

BTW I have made this change and others to the new script as remarked by @zynos (https://github.com/raspibolt/raspibolt-pulse).

resources/20-raspibolt-welcome
```diff
@@ -72,6 +72,18 @@ ${color_yellow}━━━━━━━━━━━━━━━━━━━━━
 printf "%0.s#" {1..40}
 echo -ne '\r### Loading System data \r'
 
+# get local network ip address
+local_ip=$(ip addr show | grep -w inet | grep -v 127.0.0.1 | awk '{print $2}' | cut -d/ -f1)
+
+# get available update
+updates=$((`sudo apt update &>/dev/null && sudo apt list --upgradable 2>/dev/null | wc -l`-1))
```
kristapsk, line 79: 1) `apt list --upgradable` does not require `sudo`;

2) could we add `sudo apt update` to `root` `crontab` or run as a `systemd` timer instead?

Both combined will not need entering password or modifyin `sudoers`.

nonamecrypto: I wasn't sharp when I responded to the previous messages. My PI setup also requires me to enter a password when sudo is used. That said, I run the script as part of the shell startup (i.e. added to the `.bashrc`) and it does not ask for the password.

Issue #1282, opened 2023-05-17
[BUG] remove black background color in PS1 variable


aheinzer:
### Describe the bug
The PS1 variable sets the background color of the Bitcoin Symbol (and the space before) fix to a black color. This might not match the background color of the shell.

### Location in guide
https://raspibolt.org/guide/bonus/raspberry-pi/command-line.html#command-prompt

### Expected behavior
No fixed background color should be set.

### Screenshots
<img width="183" alt="PS1 colors" src="https://github.com/raspibolt/raspibolt/assets/4063700/88c1c4b5-215d-4861-bdd7-83941b6bdc3a">

### Environment
 - Hardware platform: Raspberry Pi 4
 - Operating system: Raspberry Pi OS 64bit
 - Version: Debian GNU/Linux 11 (bullseye)

PR #1283, opened 2023-05-17
Fixes raspibolt/raspibolt#1282

aheinzer:
#### What

Proposal to fix the issue raspibolt/raspibolt#1282

### Why

It's not important, but it looks a little bit nicer :-)

#### How

Removed the black background color within the PS1 variable.

#### Scope

- [ ] independent bonus guide
- [X] simple bug fix

Fixes raspibolt/raspibolt#1282

#### Test & maintenance

- As soon the updated script is installed, the bug is fixed.
- This should not need any further maintenance.


twofaktor (approved):
tACK!

kristapsk (approved):

Issue #1285, opened 2023-05-17
[BUG] shorten the menu separators on the website


aheinzer:
### Describe the bug
The menu separator is too long and is wrapped to a second line.

### Location in guide
https://raspibolt.org

### Expected behavior
The menu separator should fit into one line.

### Screenshots
![RaspiBolt-Menu-Seperator](https://github.com/raspibolt/raspibolt/assets/4063700/ca235142-204d-47ee-85ca-37aec2925c2d)

### Environment
 - Hardware platform: MacBook Pro
 - Operating system: macOS Ventura Version 13.3.1 (a)
 - Browser: Safari Version 16.4 (18615.1.26.110.1)

PR #1286, opened 2023-05-19
shorten the menu separators on the website

aheinzer:
#### What

shorten the menu separators on the website

### Why

looks better on the website

#### How

shortened the separator lines from 26 to 20 chars

#### Scope

- [ ] independent bonus guide
- [X] simple bug fix

Fixes raspibolt/raspibolt#1285

#### Test & maintenance

Did not know how to test this locally.

### DIFF
--- a/guide/bonus-section-separator.md
+++ b/guide/bonus-section-separator.md
@@ -1,6 +1,6 @@
 ---
 layout: default
-title: __________________________
+title: ____________________
 nav_order: 99
 ---
 <meta http-equiv="refresh" content="0; URL=bonus/">
--- a/guide/troubleshooting-separator.md
+++ b/guide/troubleshooting-separator.md
@@ -1,6 +1,6 @@
 ---
 layout: default
-title: __________________________
+title: ____________________
 nav_order: 199
 ---
 <meta http-equiv="refresh" content="0; URL=troubleshooting.html">


Mynima:
I was wonder why this was but I guess it is due to screen display sizes?

No problem here:
image

Resized just before the menu auto moves to the dropdown in the top right:
image

I guess I don't have an issue with this update never really spotted the issue tbh.

LGTM.

Mynima (approved):
utACK

aheinzer:
Even on a bigger screen the lines are broken because the width of the menu sees to be fixed and too small.
Here 2 screenshots, both on MacBook on a external monitor with 2560 x 1440 pixel, on Safari and Chrome.

Safari

Chrome

Mynima:
> Even on a bigger screen the lines are broken because the width of the menu sees to be fixed and too small. Here 2 screenshots, both on MacBook on a external monitor with 2560 x 1440 pixel, on Safari and Chrome.

Yeah doesn't look ideal. Thanks for catching it and making the update. I've no issues with this, @blckbx see what you think.

kristapsk:
I have this issue both with Firefox and Google Chrome under Linux, with 1920x1080 screen resolution, without this change, on current master.

image

image

Mynima:
@kristapsk is it worth pushing this one through? It doesn't seem contentious and could resolve the issue for a few folks.

kristapsk:
I'm not sure this is the right approach to fix this. Couldn't we use some CSS instead to just hide overflow, if it happens, instead of wrapping?

Mynima:
> I'm not sure this is the right approach to fix this. Couldn't we use some CSS instead to just hide overflow, if it happens, instead of wrapping?

Ah yeah, it would be better to have a more robust approach to fixing this.

PR #1318, opened 2023-07-21
RTL: Print version you are upgrading to when upgrading

kristapsk:
#### What

Print version you are upgrading to when upgrading.

### Why

It is useful feedback. Currently you don't know to which version you are actually upgrading. You are just blindly assuming that `git tag...` oneliner does the magic.

#### Scope

- [ ] independent bonus guide
- [X] simple bug fix


twofaktor (approved):
utACK!

PR #1327, opened 2023-07-26
[UPDATE GUIDE] Update Electrum server (versioning)

### LINKED ISSUE #1305: Upgrading Electrs to 0.10.0-rc.1
### Describe the issue
After command cargo clean there is error:
error: failed to parse manifest at `/home/admin/rust/electrs/Cargo.toml`
Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2021` edition, and only supports `2015` and `2018` editions.

### Location in guide
[Link to the corresponding page and section on <https://raspibolt.org>](https://raspibolt.org/guide/bitcoin/electrum-server.html#for-the-future-electrs-upgrade)

### Environment
 - Hardware platform: Raspberry Pi 4
 - Operating system: Raspberry Pi OS 64bit, Debian 11
 - Version: Description:	Debian GNU/Linux 11 (bullseye)

### Additional context
Cargo in Bullseye is 0.47.0-3+b1, in Bookworm is 0.66.0+ds1-1. Do we need wait for Raspberry Pi OS upgrade or can we do some manual intervention? For example: curl https://sh.rustup.rs -sSf | sh.

RydalWater:
#### What

Updating Electrum server  guide to v0.10.0

### Why

New version available ([release notes](https://github.com/romanz/electrs/releases))

#### How

- Review/update guide.
- Included additional section for upgrading to bookworm (so that cargo source is updated as required)

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix

Fixes #1305

#### Test & maintenance

Tested by upgrading.

Note to reviewers: Please take extra care during this review and provide any feedback necessary, particularly around the upgrade from debian 11 (bullseye) to 12 (bookworm).

Update identified via GitLurker https://www.gitlurker.info/

### DIFF
--- a/guide/bitcoin/electrum-server.md
+++ b/guide/bitcoin/electrum-server.md
@@ -98,7 +98,7 @@ We get the latest release of the Electrs source code, verify it, compile it to a
   Other releases might not have been properly tested with the rest of the RaspiBolt configuration, though.
 
   ```sh
-  $ VERSION="0.9.14"
+  $ VERSION="0.10.0"
   $ mkdir /home/admin/rust
   $ cd /home/admin/rust
   $ git clone --branch v$VERSION https://github.com/romanz/electrs.git
@@ -174,7 +174,7 @@ We get the latest release of the Electrs source code, verify it, compile it to a
   ```
 
   ```sh
-  Starting electrs 0.9.14 on aarch64 linux with Config { network: Bitcoin, db_path: "/data/electrs/db/bitcoin", daemon_dir: "/home/bitcoin/.bitcoin", daemon_auth: CookieFile("/home/bitcoin/.bitcoin/.cookie"), daemon_rpc_addr: 127.0.0.1:8332, daemon_p2p_addr: 127.0.0.1:8333, electrum_rpc_addr: 127.0.0.1:50001, monitoring_addr: 127.0.0.1:4224, wait_duration: 10s, jsonrpc_timeout: 15s, index_batch_size: 10, index_lookup_limit: Some(1000), reindex_last_blocks: 0, auto_reindex: true, ignore_mempool: false, sync_once: false, disable_electrum_rpc: false, server_banner: "Welcome to electrs 0.9.14 (Electrum Rust Server)!", args: [] }
+  Starting electrs 0.10.0 on aarch64 linux with Config { network: Bitcoin, db_path: "/data/electrs/db/bitcoin", daemon_dir: "/home/bitcoin/.bitcoin", daemon_auth: CookieFile("/home/bitcoin/.bitcoin/.cookie"), daemon_rpc_addr: 127.0.0.1:8332, daemon_p2p_addr: 127.0.0.1:8333, electrum_rpc_addr: 127.0.0.1:50001, monitoring_addr: 127.0.0.1:4224, wait_duration: 10s, jsonrpc_timeout: 15s, index_batch_size: 10, index_lookup_limit: Some(1000), reindex_last_blocks: 0, auto_reindex: true, ignore_mempool: false, sync_once: false, disable_electrum_rpc: false, server_banner: "Welcome to electrs 0.10.0 (Electrum Rust Server)!", args: [] }
   [2021-11-09T07:09:42.744Z INFO  electrs::metrics::metrics_impl] serving Prometheus metrics on 127.0.0.1:4224
   [2021-11-09T07:09:42.744Z INFO  electrs::server] serving Electrum RPC on 127.0.0.1:50001
   [2021-11-09T07:09:42.812Z INFO  electrs::db] "/data/electrs/db/bitcoin": 0 SST files, 0 GB, 0 Grows
@@ -328,14 +328,16 @@ Make sure to check the [release notes](https://github.com/romanz/electrs/blob/ma
   ```sh
   $ cd /home/admin/rust/electrs
 
-  # Clean and update the local source code and show the latest release tag (example: v0.9.14)
+  # Clean and update the local source code and show the latest release tag (example: v0.10.0).
   $ git clean -xfd
   $ git fetch
   $ git tag | sort --version-sort | tail -n 1
-  > v0.9.14
-  
+  > v0.10.0
+
+  # Note that a version appended with '-rc' (for example: v0.10.0-rc.1) refers to a Release Candidate, which is effectively a pre-release of a given version. Where possible it is recommend to choose the formal release. To see more than one recent release modify the git tag command -n argument above and replace 1 with another positive integer (example: git tag | sort --version-sort | tail -n 3). 
+
   # Set the VERSION variable as number from latest release tag and verify developer signature
-  $ VERSION="0.9.14"
+  $ VERSION="0.10.0"
   $ git verify-tag v$VERSION
   > gpg: Good signature from "Roman Zeyde <me@romanzey.de>" [unknown]
   > [...]
@@ -348,6 +350,13 @@ Make sure to check the [release notes](https://github.com/romanz/electrs/blob/ma
   $ cargo clean
   $ cargo build --locked --release
 
+  # When executing the cargo clean command you may receive the following message. If you do, please check the 'Update cargo source' section before continuing.
+  > error: failed to parse manifest at `/home/admin/rust/electrs/Cargo.toml`
+  > Caused by:
+      failed to parse the `edition` key
+  > Caused by:
+      this version of Cargo is older than the `2021` edition, and only  supports `2015` and `2018` editions.
+
   # Back up the old version and update
   $ sudo cp /usr/local/bin/electrs /usr/local/bin/electrs-old
   $ sudo install -m 0755 -o root -g root -t /usr/local/bin ./target/release/electrs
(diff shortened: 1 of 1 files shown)


Mynima:
@luksocculta and @keyser-soeze I would really appreciate your feedback/review of the proposed updates. If you have any suggestions/comments or you feel that I've missed anything critical please let me know. I based these updates in large part on your discussion for issue #1305, thanks for that.

In regards to the ots breaking I did test this on my side and found that ots worked as normal after the upgrades so I've left that out of the guide. If you feel strongly about adding this please make a proposal for code changes.

Finally, I did run into a minor issue with regards to the [system overview](https://raspibolt.org/guide/bonus/raspberry-pi/system-overview.html) breaking for both RTL and the Block Explorer. I will investigate these later this evening, it looks like RTL is working fine so for some reason it just isn't pulling through correctly to the overview. The Block Explorer is having some other issues so I'll dive into that one first.

luksocculta (approved):
During my upgrade there wasn't need to manually install 'libnginx-mod-stream'. It upgraded automatically.

And you can add info about do not overwriting existing settings. If you overwrite for example 'sshd' settings you can be cut out from ssh connection and you will be forced to connecting external monitor to your Raspberry and it could be problematic.

Mynima:
> During my upgrade there wasn't need to manually install 'libnginx-mod-stream'. It upgraded automatically.
> 
> And you can add info about do not overwriting existing settings. If you overwrite for example 'sshd' settings you can be cut out from ssh connection and you will be forced to connecting external monitor to your Raspberry and it could be problematic.

Thanks, will make sure to highlight the issues with sshd and take a look at removing the libnginx-mod-stream bit too if it is non-essential. 

Just working on getting my block explorer online, it seems to be having issues with connecting to the node. Not liking the old cookie authentication for some reason.

guide/bitcoin/electrum-server.md
```diff
@@ -324,41 +324,144 @@ Make sure to check the [release notes](https://github.com/romanz/electrs/blob/ma
...
+
+  ```sh
+  $ sudo nano /etc/apt/sources.list.d/*
+  ```
+  
+  * Update, upgrade and then do a dist-upgrade to debian 12 (bookworm)
+  ```sh
+  $ sudo apt update 
+  $ sudo apt upgrade
+  $ sudo dist-upgrade
```
Derek701, line 430:
```suggestion
  $ sudo apt dist-upgrade
```

guide/bitcoin/electrum-server.md
```diff
@@ -324,41 +324,144 @@ Make sure to check the [release notes](https://github.com/romanz/electrs/blob/ma
...
+  ```
+
+  * Install RocksDB per the [v0.10.0-rc.1 release notes](https://github.com/romanz/electrs/blob/master/RELEASE-NOTES.md#0100-rc1-jun-21-2023). You will be prompted to restart services during this process make sure to select "No" otherwise you may have issues connecting via SSH.
```
Derek701, line 433: In my testing, there was not a prompt to restart services for this particular command.
```suggestion
  * Install RocksDB per the [v0.10.0-rc.1 release notes](https://github.com/romanz/electrs/blob/master/RELEASE-NOTES.md#0100-rc1-jun-21-2023).
```

guide/bitcoin/electrum-server.md
```diff
@@ -324,41 +324,144 @@ Make sure to check the [release notes](https://github.com/romanz/electrs/blob/ma
...
+  ```sh
+  $ sudo apt install librocksdb-dev=7.8.3-2
+  ```
+
+  * Install the stream module for nginx. This will resolve any issues with restarting the nginx service after the upgrade. When prompted, select "N" to keep the current-installed version of the config files. Then restart the nginx service.
```
Derek701, line 438: Perhaps change wording to an install check or note that the stream module may already be installed.

guide/bitcoin/electrum-server.md
```diff
@@ -324,41 +324,144 @@ Make sure to check the [release notes](https://github.com/romanz/electrs/blob/ma
...
+  ```sh
+  $ sudo apt update
+  $ sudo apt install libnginx-mod-stream
+  $ sudo systemctl restart nginx
+  ```
+
+   * Update and upgrade to the latest pacakage versions. Again, if prompted to replace any config files select "N" to keep the current versions throughout.
```
Derek701, line 445: No config replacement prompt here in my testing.

guide/bitcoin/electrum-server.md
Mynima: Updated, added a check for the installed package prior to installation step

guide/bitcoin/electrum-server.md
Mynima: Removed comment

guide/bitcoin/electrum-server.md
```diff
@@ -324,41 +324,150 @@ Make sure to check the [release notes](https://github.com/romanz/electrs/blob/ma
...
+* Restart Electrs
+  ```sh
   $ sudo systemctl restart electrs
   ```
 
+---
+
+## Update cargo source (v0.10.0-rc.1 onwards)
+
+Although updating Electrs is normally straight-forward, if you encorntered an error during the upgrade from v0.9.14 to v0.10.0 then it means you will need to upgrade the version of debian from 11 (bullseye) to 12 (bookworm) as the previous version has outdated sources for cargo. If you do not wish to do this then you can always continue to run an older version of electrs.
```
blckbx, line 391:
```suggestion
Although updating Electrs is normally straight-forward, if you encounter an error during the upgrade from v0.9.14 to v0.10.0 then it means you will need to upgrade the version of debian from 11 (bullseye) to 12 (bookworm) as the previous version has outdated sources for cargo. If you do not wish to do this then you can always continue to run an older version of electrs.
```

guide/bitcoin/electrum-server.md
```diff
@@ -324,41 +324,150 @@ Make sure to check the [release notes](https://github.com/romanz/electrs/blob/ma
...
+  ```sh
+  $ sudo apt update 
+  $ sudo apt upgrade
+  $ sudo apt dist-upgrade
+  ```
+
+  * Install RocksDB per the [v0.10.0-rc.1 release notes](https://github.com/romanz/electrs/blob/master/RELEASE-NOTES.md#0100-rc1-jun-21-2023).
+  ```sh
+  $ sudo apt install librocksdb-dev=7.8.3-2
+  ```
```
blckbx, line 436: As far as I understood, we're making use of static linking by running `cargo build --locked --release` so that package `librocksdb` is not required to be installed on the system (I checked locally and it isn't installed).

blckbx (commented):
Some comments. Looks good. But tbh I'm really unsure how we should handle the Debian dist-upgrade on RaspiBolt. Leaving this open for discussion.

guide/bitcoin/electrum-server.md
Mynima: Did you skip this step on testing and you're up and running without any issues? I initially added it on recommendation from @keyser-soeze and with cross checking the release notes.

blckbx: Yes, never installed librocksdb (neither on initial installation nor on upgrade to v0.10.0) and running it productively. 
The guide always used static linking as installation method, so that on upgrading we don't have to install the rocksdb library imho.

Static Linking: https://github.com/romanz/electrs/blob/master/doc/install.md#static-linking
Dynamic Linking: https://github.com/romanz/electrs/blob/master/doc/install.md#dynamic-linking

Mynima: > Yes, never installed librocksdb (neither on initial installation nor on upgrade to v0.10.0) and running it productively. The guide always used static linking as installation method, so that on upgrading we don't have to install the rocksdb library imho.
> 
> Static Linking: https://github.com/romanz/electrs/blob/master/doc/install.md#static-linking Dynamic Linking: https://github.com/romanz/electrs/blob/master/doc/install.md#dynamic-linking

Great, thank for confirming/highlighting. I've removed this step from the guide to avoid unnecessary installations.

blckbx:
@kristapsk  @twofaktor what's your thoughts on the debian upgrade part? last time I checked, RaspiOS image is still based on Debian 11.

kristapsk:
> @kristapsk @twofaktor what's your thoughts on the debian upgrade part? last time I checked, RaspiOS image is still based on Debian 11.

I don't use Raspbian myself, my RaspiBolt is running on different hardware with Ubuntu 20.04 LTS (focal). We should avoid adding stuff that depends on a specific Debian versions if possible.

blckbx:
> I don't use Raspbian myself, my RaspiBolt is running on different hardware with Ubuntu 20.04 LTS (focal). We should avoid adding stuff that depends on a specific Debian versions if possible.

So we should postpone electrs v0.10.0 update until RaspiOS has officially upgraded the base OS?

kristapsk:
> > I don't use Raspbian myself, my RaspiBolt is running on different hardware with Ubuntu 20.04 LTS (focal). We should avoid adding stuff that depends on a specific Debian versions if possible.
> 
> So we should postpone electrs v0.10.0 update until RaspiOS has officially upgraded the base OS?

Did `grep` through while guide, there is another place that updates `/etc/apt.sources.list` (actually `/etc/apt/sources.list.d/...`, that is a better way) and are specific tor Debian version and that is [privacy guide](https://raspibolt.org/guide/raspberry-pi/privacy.html). So, if this is added here, that needs to be changed too.

twofaktor:
> @kristapsk @twofaktor what's your thoughts on the debian upgrade part? last time I checked, RaspiOS image is still based on Debian 11.

I use Dynamic Linking and Rustup instead of Static linking and cargo using apt, on MiniBolt (Ubuntu Server 22.4). Also, I install `librocksdb` from the source to get v7.8.3. I think this is the best solution also to the current RaspiBolt (Debian 11)

Ref: https://v2.minibolt.info/bonus-guides/bitcoin/electrs

zynos:
This PR could continue now, when we update: https://github.com/raspibolt/raspibolt/issues/1382

sig-noi:
I used these instructions to update to Debian 12 and they worked OK. I added the command `sudo apt autoremove` which may have contributed to the following applications breaking during the install:
-bitcoin explorer
-joinmarket
-OTS

To fix these things, I ended up reinstalling python and OTS with 
`sudo apt-get install python3 python3-dev python3-pip python3-setuptools python3-wheel`
`sudo pip3 install opentimestamps-clinet --break-system-prackages`

Then I reinstalled joinmarket per the bonus guide Installation instructions, which went fine.

Bitcoin explorer had stopped working due to an nginx issue that was fixed by running: 
`sudo apt install libnginx-mod-stream`

Not certain which of the above issues were caused by the PR instructions vs adding `sudo apt autoremove` but figured it was worth sharing this experience here.

Thank you!

soc7:
> I used these instructions to update to Debian 12 and they worked OK. I added the command `sudo apt autoremove` which may have contributed to the following applications breaking during the install: -bitcoin explorer -joinmarket -OTS
> 
> To fix these things, I ended up reinstalling python and OTS with `sudo apt-get install python3 python3-dev python3-pip python3-setuptools python3-wheel` `sudo pip3 install opentimestamps-clinet --break-system-prackages`
> 
> Then I reinstalled joinmarket per the bonus guide Installation instructions, which went fine.
> 
> Bitcoin explorer had stopped working due to an nginx issue that was fixed by running: `sudo apt install libnginx-mod-stream`
> 
> Not certain which of the above issues were caused by the PR instructions vs adding `sudo apt autoremove` but figured it was worth sharing this experience here.
> 
> Thank you!

Could you please share how did you upgrade to debian 12? simply change the repo and run apt update && apt upgrade then apt dist-upgrade?

PR #1331, opened 2023-07-27
Remove v1/v2 LND auto-unlock bonus guide

kristapsk:
#### What

For current RaspiBolt v3 auto-unlock is already handled by the main Lightning (LND) guide.

### Why

I don't see a point keeping deprecated RaspiBolt v1/v2 guides on git master branch. It will still be accessible for those users at https://v2.raspibolt.org/raspibolt_6A_auto-unlock.html.

#### Scope

- [X] independent bonus guide
- [ ] simple bug fix


twofaktor:
Concept ACK!

twofaktor (approved):

PR #1343, opened 2023-09-09
create home folders as 0755 on Debian 12

openoms:
#### What
Adding changes in Raspibolt as well because the user creation command in Raspiblitz was originating from here, but now needs to be changed.

Changing the user creation in every script from:
```
sudo adduser --disabled-password --gecos "" ${APPID}
```
to
```
sudo adduser --system  --shell /bin/bash --group --home /home/${APPID} ${APPID}
```
These changes are required because of the change in the default permissions in Debian 12. 
Tested the commands work to the same effect on Debian11 as well.

### Why

See the permission issues documented in: https://github.com/raspiblitz/raspiblitz/issues/4154

#### How
`man adduser`:
>If called with one non-option argument and the --system option, adduser will add a dynamically allocated
       system user, often abbreviated as system user in the context of the adduser package.
       By  default,  system  users are placed in the nogroup group.  To place the new system user in an already
       existing group, use the --gid or --ingroup options.  If the --group is given and the  identically  named
       group does not already exist, it is created with the same ID.
       If  no  home  directory  is specified, the default home directory for a new system user is /nonexistent.
       This directory should never exist on any Debian system, and adduser will never create it automatically.

`man adduser.conf`
>DIR_MODE
              The  permissions  mode  for  home directories of non-system users that are created by adduser(8).
              Defaults to 0700.  Note that there are potential configurations (such as /~user web services,  or
              in-home mail delivery) which will require changes to the default.  See also SYS_DIR_MODE.

>SYS_DIR_MODE
              The  permissions  mode  for home directories of system users that are created by adduser(8).  De‐
              faults to 0755.  Note that changing the default permissions for system users may cause some pack‐
              ages to behave unreliably, if the program relies on the default setting.  See also DIR_MODE.

#### Scope

#### Test & maintenance
Create users on Debian 11 and Debian 12
Compare the permissions in:
```
ls -la /home
```

### DIFF
--- a/guide/bitcoin/blockchain-explorer.md
+++ b/guide/bitcoin/blockchain-explorer.md
@@ -116,7 +116,7 @@ An attacker would not be able to do much within this user's permission settings.
 * Create a new user, assign it to the "bitcoin" group, and open a new session
 
   ```sh
-  $ sudo adduser --disabled-password --gecos "" btcrpcexplorer
+  $ sudo adduser --system --group --shell /bin/bash --home /home/btcrpcexplorer btcrpcexplorer
   $ sudo adduser btcrpcexplorer bitcoin
   $ sudo su - btcrpcexplorer
   ```
--- a/guide/bitcoin/electrum-server.md
+++ b/guide/bitcoin/electrum-server.md
@@ -129,7 +129,7 @@ We get the latest release of the Electrs source code, verify it, compile it to a
 * Create the "electrs" service user, and make it a member of the "bitcoin" group
 
   ```sh
-  $ sudo adduser --disabled-password --gecos "" electrs
+  $ sudo adduser --system --group --shell /bin/bash --home /home/electrs electrs
   $ sudo adduser electrs bitcoin
   ```
 
--- a/guide/bonus/bitcoin/Jam.md
+++ b/guide/bonus/bitcoin/Jam.md
@@ -160,7 +160,7 @@ $ sudo ufw status
 * Create a dedicated user for Jam and switch to user
 
 ```sh
-$ sudo adduser --disabled-password --gecos "" jam
+$ sudo adduser --system --group --shell /bin/bash --home /home/jam jam
 
 $ sudo su - jam
 ```
--- a/guide/bonus/bitcoin/dojo.md
+++ b/guide/bonus/bitcoin/dojo.md
@@ -235,7 +235,7 @@ If you are using Electrs instead of Fulcrum, it is necessary to make following c
 * Create the user "dojo" and add him to the group “bitcoin” as well
 
   ```sh
-  $ sudo adduser --disabled-password --gecos "" dojo
+  $ sudo adduser --system --group --shell /bin/bash --home /home/dojo dojo
   $ sudo adduser dojo bitcoin
   ```
 
--- a/guide/bonus/bitcoin/fulcrum.md
+++ b/guide/bonus/bitcoin/fulcrum.md
@@ -236,7 +236,7 @@ Now that Fulcrum is installed, we need to configure it to run automatically on s
 * Create the "fulcrum" service user, and add it to "bitcoin" group
 
   ```sh
-  $ sudo adduser --disabled-password --gecos "" fulcrum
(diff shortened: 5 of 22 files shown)


twofaktor:
Raspberry Pi OS lite doesn't have a Debian version 12 based system yet

openoms:
Yes, this problem surfaces with Debian12, but the solution is backwards compatible.

Also anyone using this guide to install on amd64 PC would have the issue.

twofaktor:
> Yes, this problem surfaces with Debian12, but the solution is backwards compatible.

Thanks for your suggestion, but I think that we should cover this when the time comes for migration

> Also anyone using this guide to install on amd64 PC would have the issue.

I don't think other platforms/architecture are covered in the main RaspiBolt guide, which should be covered in some bonus guide, or better yet, use [MiniBolt](https://minibolt.info)

openoms:
No urgency here, can keep this pending and update when needed.

openoms:
Closed by error here with the force push.

needed to add 
`--shell /bin/bash` 
to the adduser command. Will open again after further testing.

openoms:
needed to add
```
--shell /bin/bash
```
to the adduser command.

openoms:
Can't reopens so will open again after further testing.

openoms:
needed to add
```
--shell /bin/bash
```
to the adduser command.

openoms:
needed to add
```
--shell /bin/bash
```
to the adduser command.

openoms:
Note that the default  RaspberryOS is now Debian12 and the guide made for Debian11 fails: https://t.me/raspibolt/22787

Feel free to comment or change anything here as I am not running a raspibolt setup myself.

blckbx:
@kristapsk @twofaktor 
Shall we merge this PR now and try to upgrade the guide to bookworm step by step? 

I guess the major problem is none of us is running a Raspberry Pi any longer? Not sure if we can adapt the installation section for bookworm easily or how to upgrade existing systems. We could probably take parts of @Mynima 's electrs upgrade section (https://github.com/Mynima/raspibolt/blob/electRS-versioning/guide/bitcoin/electrum-server.md#update-cargo-source-v0100-rc1-onwards) and create a separate bonus guide out of it.

kristapsk:
I will look at this. Basically, if it doesn't break anything for older Debians, it's safe to merge.

twofaktor:
> @kristapsk @twofaktor Shall we merge this PR now and try to upgrade the guide to bookworm step by step?

I think this PR is not needed, I tested RaspiBolt on a fresh installation, using Bookworm 64 bits OS on a physical Raspberry Pi 4, and created users, which only output me this additional logs (e.g electrs user):

```
Adding user `electrs' ...
Adding new group `electrs' (1002) ...
Adding new user `electrs' (1002) with group `electrs (1002)' ...
Creating home directory `/home/electrs' ...
Copying files from `/etc/skel' ...
Adding new user `electrs' to supplemental / extra groups `users' ...
Adding user `electrs' to group `users' ...
```

The rest worked fine, except with this change that wasn't supposed to be done yet: https://github.com/raspibolt/raspibolt/pull/1366 and something on the [OTS client installation](https://t.me/raspibolt/22542), anyway, see my last comment ⬇️

> I guess the major problem is none of us is running a Raspberry Pi any longer? 

Yes, I still have x2 Raspberry Pi 4

> Not sure if we can adapt the installation section for bookworm easily or how to upgrade existing systems. We could probably take parts of @Mynima 's electrs upgrade section (https://github.com/Mynima/raspibolt/blob/electRS-versioning/guide/bitcoin/electrum-server.md#update-cargo-source-v0100-rc1-onwards) and create a separate bonus guide out of it.

As I have said on [occasion](https://github.com/raspibolt/raspibolt/pull/1388#issuecomment-1849661170), I think we should not update without testing first, there could be more things that don't work in the rest of the guide with Bookworm. I am not a fan of the "utACKs", and there is no rush to update if the Raspibolt environment is working well at the moment.

hellresistor:
i am using debian 12 .. and work good. this is not needed

openoms:
ok, closing as not needed

PR #1346, opened 2023-09-14
CLN REST certs upgrade note

s373nZ:
CLN REST certs upgrade note

#### What

When upgrading CLN REST plugin, user will need to copy over their prior `certs` directory in order to preserve the access credentials. Otherwise client apps like Zeus will lose connectivity.

### Why

Should be a helpful reminder to acknowledge CLN version and CLN REST versions are functionally related, probably upgraded together, and a manual port of credentials / certs is necessary because of the tarball install method in CLN REST documentation.

#### How

Upgraded and lost access, fixed and updated docs.

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

Peform an upgrade and make sure RTL and Zeus still connect.

#### Animated GIF (optional)

#nosnazz


blckbx (approved):
LGTM. Thanks!

PR #1349, opened 2023-09-25
Troubleshooting - Locked Out Guide

bliotti:
Maybe useful for troubleshooting section? 

This info saved me :)

"What to do when unable to login to admin and/or root user account."


kristapsk:
Concept ACK.

For Linux desktop users another option would be to mount SD card locally, then use `chroot` and `passwd` from that machine.

PR #1364, opened 2023-10-13
[UPDATE BONUS GUIDE] LNDg: update to v1.8.0

blckbx:
#### What

This PR updates LNDg to latest version 1.8.0. This requires substinential changes to required systemd services.
By implementing a central controller unit, there's now only one single systemd service required that collects data, runs rebalances and fetches htlc stream data (next to uwsgi).

### Why

Keeping up with latest changes. Simplifying the guide with this update.

#### How

Local test following installation insctructions: https://github.com/cryptosharks131/lndg/releases/tag/v1.8.0

### DIFF
--- a/guide/bonus/lightning/lndg.md
+++ b/guide/bonus/lightning/lndg.md
@@ -277,7 +277,7 @@ LNDg stores the LN node routing statistics and settings in a SQL database. We'll
   Description=LNDg uWSGI app
   After=lnd.service
   PartOf=lnd.service
-  Wants=controller-lndg.service
+  Wants=lndg-controller.service
   
   [Service]
   ExecStart=/home/lndg/lndg/.venv/bin/uwsgi --ini /home/lndg/lndg/lndg.ini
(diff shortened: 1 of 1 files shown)


d6n13l0l1v3r (commented):
change  99 look fine 

line  101 and 590, why you add tags/$LATEST_RELEASE it's not necessary, just doing 

`git checkout $LATEST_RELEASE ` 

after you can confirm that you are in the right place with 

`git log `

Note: the comment below  are not part for your initial  commit but will be good  if you test and take in consideration 

line 589  I think will be better
`git tag | grep -E "v[0-9]+.[0-9]+.[0-9]+$" | sort --version-sort | tail -n 1`
in order to retry the last tag and not the full list of all tags 

 line 587 I think will be better  `git fetch --all --tags`

blckbx:
> line 101 and 590, why you add tags/$LATEST_RELEASE it's not necessary

There's both a v1.8.0 branch and tag which differed at the time of writing. Author advised to use the tag explicitly. 

Looking at your other proposals in depth later.

d6n13l0l1v3r:
ack thanks for clarify

cryptotar:
Attempted the upgrade and got 502 Bad Gateway when trying to navigate in browser. Went through uninstall and followed instructions for new install and continued to get the error. 

Ended up having to change /etc/nginx/nginx.conf user from www-data to lndg. Thunderhub and RTL still work (I know, a lot of apps that do the same thing). Guessing something with www-data permissions, but haven't looked into it yet.

blckbx:
> Attempted the upgrade and got 502 Bad Gateway when trying to navigate in browser. Went through uninstall and followed instructions for new install and continued to get the error.
> 
> Ended up having to change /etc/nginx/nginx.conf user from www-data to lndg. Thunderhub and RTL still work (I know, a lot of apps that do the same thing). Guessing something with www-data permissions, but haven't looked into it yet.

Hey, thanks for reporting back. Nginx config hasn't changed from 1.7.x to 1.8, so I wonder what could have been the root cause here. Did you add user lndg to group www-data and vice-versa?
```
$ sudo adduser lndg www-data
$ sudo adduser www-data lndg
```

PR #1365, opened 2023-10-17
Update security.md

SigmaCodeLLC:
The nginx_stream module adds stream proxy support to nginx, producing a successful output to "sudo nginx -t". Otherwise, "sudo nginx -t" fails and produces: "unknown directive "stream" in /etc/nginx/nginx.conf:20 nginx: configuration file /etc/nginx/nginx.conf test failed"

#### What

Improving the base code structure.

### Why

To enable the bitcoin user to successfully employ a bitcoin node.

#### How

Importing a required package for NGINX.

#### Scope

- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

Start from a root install and replicate the bug by not installing the recommended package.

#### Animated GIF (optional)

*-|

### DIFF
--- a/guide/raspberry-pi/security.md
+++ b/guide/raspberry-pi/security.md
@@ -232,6 +232,7 @@ This setup is called a "reverse proxy": NGINX provides secure communication to t
 
   ```sh
   $ sudo apt install nginx
+  $ sudo apt install libnginx-mod-stream
   ```
 
 * Create a self-signed SSL/TLS certificate (valid for 10 years)


kristapsk:
Provided sample `nginx.conf` in https://raspibolt.org/guide/raspberry-pi/security.html#prepare-nginx-reverse-proxy does not reference `nginx_stream` module. It's not good security practice to install modules we don't need.

SigmaCodeLLC:
> Provided sample `nginx.conf` in https://raspibolt.org/guide/raspberry-pi/security.html#prepare-nginx-reverse-proxy does not reference `nginx_stream` module. It's not good security practice to install modules we don't need.

This error came about after I tried a fresh install with the Raspberry Pi OS Bookworm (on a Raspberry PI 4). That might have different package management overall that needs a closer look, namely with nginx and tor.

kristapsk:
> Provided sample `nginx.conf` in https://raspibolt.org/guide/raspberry-pi/security.html#prepare-nginx-reverse-proxy does not reference `nginx_stream` module. It's not good security practice to install modules we don't need.

My bad, it does, with `stream {`. But I think #1366 approach is better. Apparently for some Debian versions `stream` support in nginx is on by default, for some you need to install additional package.

kristapsk:
Resolved with #1366.

PR #1366, opened 2023-10-19
guide/rpi: command for installing nginx stream module

ellemouton:
#### What

Add the command to install the nginx stream module incase of ``unknown directive "stream"`. 

### Why

I was making my way step-by-step through the guide but ran into the `unknown directive "stream" in /etc/nginx/nginx.conf:10` error when trying to validate the nginx config. It's not immediately clear what the
issue is for peeps unfamiliar with nginx modules. 

#### How

#### Scope

simple helper incase of nginx config issues.

### DIFF
--- a/guide/raspberry-pi/security.md
+++ b/guide/raspberry-pi/security.md
@@ -284,6 +284,13 @@ This setup is called a "reverse proxy": NGINX provides secure communication to t
   > nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
   > nginx: configuration file /etc/nginx/nginx.conf test is successful
   ```
+  
+If you run into the `unknown directive "stream"` error, then you need to install
+the stream module first:
+
+  ```sh
+  $ sudo apt install libnginx-mod-stream
+  ```
 
 ## Disable wireless interfaces
 


kristapsk (approved):
utACK 9abd7b1f21c547a17d7e8d6c78c7b06127dc1b82

twofaktor:
This PR should be reverted, `sudo apt install libnginx-mod-stream` is redundant, `libnginx-mod-stream` package is installed by default when you install ngnix using the `sudo apt install nginx` command

Tested 2 days ago from zero using Raspberry Pi OS arm64 (bullseye)

imagen

kristapsk:
Could this (which nginx modules are installed by default) differ from Debian version to version?

In any case, this starts with:

> If you run into the `unknown directive "stream"` error

twofaktor:
Ok, I think we shouldn't add to the guide additional commands that not explicit affect to a current RaspiBolt environment

All changes should be tested in a RaspiBolt environment (preferably with a freshly installed operating system) (and currently Raspberry Pi OS 64 bits bullseye) before merge to avoid future issues and bad UX

ellemouton:
@twofaktor  - I followed the guide step by step exactly before running into this error.

twofaktor:
> @twofaktor - I followed the guide step by step exactly before running into this error.

You didn't use the current Raspberry Pi OS (bulleye) - Debian 11 based, used on RaspiBolt still (not the latest Bookworm)

Issue #1345, opened 2023-09-14
[BUG] Upgrade RTL to 0.14.0 breaks CLN 22.11.1


s373nZ:
### Describe the bug
The commit to Upgrade RTL to `0.14.0` here:

https://github.com/raspibolt/raspibolt/commit/8277ef7148695af05ac52884b0e84583f6daadab

broke CLN on `22.11.1`. I needed to upgrade CLN to the latest `v23.08.1` as well as c-lightning-rest to `0.10.5` in order to get things back up.

### Location in guide
https://raspibolt.org/guide/bonus/lightning/cln.html

### Expected behavior
Instructions for CLN bonus guide should be updated for latest versions.

### Environment
 - Hardware platform: Raspberry Pi 4
 - Operating system: Debian 11
 - Version: Debian GNU/Linux 11 (bullseye)

s373nZ:
Looks like this is addressed in PR #1340

PR #1372, opened 2023-11-05
fix c-lightning-REST plugin support for CLN>=23.08

LeoSpyke:
#### What
Add instructions to support c-lightning-REST plugin for CLN version >=23.08

### Why
The lightningd process would not start without disabling the built-in clnrest plugin.
This PR partially addresses #1345 as well.

#### How
Tested on my own lab.

#### Scope
- [ ] independent bonus guide
- [x] simple bug fix

#### Test & maintenance
The guide will eventually migrate to the new plugin when RTL will natively support it (@ShahanaFarooqui is working on it [in this branch](https://github.com/Ride-The-Lightning/RTL/tree/clnrest-migration)).

### DIFF
--- a/guide/bonus/lightning/cln.md
+++ b/guide/bonus/lightning/cln.md
@@ -495,6 +495,16 @@ c-lightning-Rest: REST APIs for c-lightning written with node.js and provided wi
   UNUSUAL plugin-plugin.js: --- cl-rest api server is ready and listening on port: 3092 ---
   UNUSUAL plugin-plugin.js: --- cl-rest doc server is ready and listening on port: 4091 ---
   ```
+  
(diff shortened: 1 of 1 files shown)


ShahanaFarooqui:
@LeoSpyke Core lightning's `clnrest` plugin requires some python libraries and it will auto disable itself if any of them are missing. 

So you might NOT have to disable it explicitly if either those libraries are not installed already or python is not installed on your system.

However, disabling the plugin (specially if you are using `c-lightning-REST`) is also fine till RTL and other apps are not migrated to `clnrest`.

kristapsk:
[RTL 0.15.0](https://github.com/Ride-The-Lightning/RTL/releases/tag/v0.15.0) has been released, which switches from c-lightning-REST to CLNRest.

ShahanaFarooqui:
This `disable-plugin` is not required from CLN v23.11 onwards. clnrest's config options have been changed to [prefix `cln`](https://docs.corelightning.org/docs/rest) like `clnrest-port`, `clnrest-host` etc.

PR #1376, opened 2023-11-14
Update fulcrum.md to include NGINX reverse proxy configuration

mzdrowski:
#### What

Need to configure NGINX reverse proxy for Fulcrum, otherwise SSL connection won't work. 

### Why

Otherwise SSL connection won't work.

#### How

Create NGINX config file for Fulcrum; method copied from the Electrs guide.

### DIFF
--- a/guide/bonus/bitcoin/fulcrum.md
+++ b/guide/bonus/bitcoin/fulcrum.md
@@ -139,7 +139,32 @@ zram-swap is a compressed swap in memory and on disk and is necessary for the pr
   ...
   ```
 
-### Configure Firewall
+### Configure firewall and reverse proxy
+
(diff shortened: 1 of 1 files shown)


twofaktor:
It works perfectly on my RaspiBolt mainnet & testnet node:

282940610-415f1831-90fe-459d-a2f8-6fd91be2cbd3
imagen


NACK

DanielPX1:
Fulcrum does not need NGINX config to work with SSL

mzdrowski:
> Fulcrum does not need NGINX config to work with SSL

Okay, thanks. I did a clean install on a PC running Debian 12 (no electrs or other electrum install) and it didn’t work for me — until I created an NGINX config file for it. I’ll go back, delete the file and see what happens.

PR #1379, opened 2023-11-17
web-app: fix enumeration and formatting

blckbx:
#### What

This small PR fixes enumeration and formatting of NodeJS installation part in RTL's installation guide. 

### Why

Styleguide

#### Scope

- [ ] independent bonus guide
- [x] simple formatting improvement


twofaktor (approved):
LGTM, but I think we should separate Node + NPM in a independent bonus guide as a common language
WDYT?

blckbx:
> LGTM, but I think we should separate Node + NPM in a independent bonus guide as a common language WDYT?

long-term yes, definitely!

PR #1388, opened 2023-12-07
[UPDATE] Bump LND to v0.17.3

twofaktor:
#### What

Updates LND guide to newly released v0.17.3

### Why

Keeping up with latest developments and improvements (see [release notes](https://github.com/lightningnetwork/lnd/blob/v0.17.3-beta/docs/release-notes/release-notes-0.17.3.md))

#### How

- Updated variable environment and outputs

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

Tested for myself. Working fine ✅

imagen

### DIFF
--- a/guide/lightning/lightning-client.md
+++ b/guide/lightning/lightning-client.md
@@ -32,7 +32,7 @@ We'll download, verify and install LND.
 
   ```sh
   $ cd /tmp
-  $ VERSION="0.17.2"
+  $ VERSION="0.17.3"
   $ wget https://github.com/lightningnetwork/lnd/releases/download/v$VERSION-beta/lnd-linux-arm64-v$VERSION-beta.tar.gz
   $ wget https://github.com/lightningnetwork/lnd/releases/download/v$VERSION-beta/manifest-v$VERSION-beta.txt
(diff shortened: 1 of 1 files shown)


blckbx:
tACK amd64

kristapsk:
Why are we hurrying so fast with merges? This doesn't look so critical. It's better to let more people review too, just in case.

kristapsk (commented):
post-merge ACK

twofaktor:
> Why are we hurrying so fast with merges? This doesn't look so critical. It's better to let more people review too, just in case.

Why is this different for you? https://github.com/raspibolt/raspibolt/pull/1366#pullrequestreview-1687850813

I'm not a fan of the utACK, btw

kristapsk:
> > Why are we hurrying so fast with merges? This doesn't look so critical. It's better to let more people review too, just in case.
> 
> Why is this different for you? [#1366 (review)](https://github.com/raspibolt/raspibolt/pull/1366#pullrequestreview-1687850813)
> 
> I'm not a fan of the utACK, btw

Configuration / documentation change is more easy to rollback than software update in case something goes wrong. It just gave hint what to do in case of specific error, didn't change core software or even core config.

PR #1396, opened 2023-12-17
Add Python 3.9 installation into Troubleshooting

hellresistor:
#### What

solve problem if fail install python 3.9

### Why

Because can happen

#### How

Do installation compiling by source code

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

#### Test & maintenance

Just follow steps to install python3.9 if fails by apt-get


kristapsk:
> Because can happen

Care to share an example? Only way that comes into my mind is if system is already messed up and there are some dependency conflicts.

PR #1402, opened 2024-01-04
Increase default mempool to 2 GB

Bitcoinite:
Recently we have seen many node operators struggling with transactions that are't broadcasted due to being below the purging limit. If a (force) close transactions fee rate is below the purging limit, the node will drop it immediately and therefore it won't be broadcasted at all. We should increase the default mempool size significantly to mitigate this issue.

I'm open to discuss the ideal default size that we should recommend during the initial setup. The 2 GB I'm proposing is just what I have set on my own node.

### DIFF
--- a/guide/bitcoin/bitcoin-client.md
+++ b/guide/bitcoin/bitcoin-client.md
@@ -257,6 +257,9 @@ We'll also set the proper access permissions.
   maxconnections=40
   maxuploadtarget=5000
 
+  # Set the size of the nodes internal mempool (default is 300)
+  maxmempool=2000
+
   # Initial block download optimizations
   dbcache=2000
   blocksonly=1


openoms:
Important to say how much RAM you have available. A 4GB computer would probably struggle with this.

Bitcoinite:
Do we have any data or estimations on what RPi model most node operations are actually using?
Maybe now is the time to discuss whether it is necessary to have at least 8GB of RAM in order to be able to operate a proper lightning node in such a mempool environment?

We might want to change the recommendation at the beginning of the guide to 8GB: https://github.com/raspibolt/raspibolt/blob/master/guide/raspberry-pi/preparations.md#hardware-requirements

Also we could add a remark (telling that you might comment out the maxmempool line if you are running a 4GB model) here: https://github.com/raspibolt/raspibolt/blob/master/guide/bitcoin/bitcoin-client.md#configuration

kristapsk:
2 GB is definitely overkill and a lot of people will struggle with this. Mempool.space currently shows 1.24 GB, but that includes even 1 sat/vB transactions.

Also, no Bitcoin software should even attempt to broadcast a transaction that's below current `mempoolminfee` of `bitcoind`.

I'm running RaspiBolt on different ARM hardware with 4 GB RAM. Have default 300 MB mempool, haven't experienced any issues because of that.

blckbx:
https://jochen-hoenicke.de/queue/#BTC,24h,weight

600MB seem to be sufficient (for now).

PR #1405, opened 2024-02-07
Added Watchdog Timer

SatsForFreedom:
#### What

The reason for enabling the watchdog timer is to automatically recover the system from unresponsive states, such as freezes or crashes, without manual intervention. This is crucial for ensuring the Raspberry Pi can operate continuously and reliably, particularly in remote, unattended, or critical applications.

### Why

This change is important because it significantly improves the system's reliability and uptime. It ensures that the Raspberry Pi can autonomously reset itself in case of software failures, minimizing downtime and maintaining service availability, which is essential for critical applications and deployments where manual resets are impractical.

#### How

This change was accomplished by installing the watchdog software package on the Raspberry Pi, configuring the watchdog timer settings in the system's configuration files to specify the conditions under which the system should be rebooted, and enabling the hardware watchdog device via the boot configuration. Finally, the watchdog service was activated to start monitoring the system, completing the setup for automatic system recovery.

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

How can the changes be tested? 
One way to test the watchdog is to force a kernel panic, which should make the system unresponsive, triggering the watchdog. This can be done by running:

`echo c | sudo tee /proc/sysrq-trigger`
**Warning**: This command will crash your system immediately, causing unsaved data to be lost.

Is there ongoing maintenance effort? 
No

If this is a bonus guide: are you willing to update it from time to time?
Yes


kristapsk:
Concept ACK. Enabled this on my Odroid HC4 device (not an RPi, but that shouldn't be a big difference in this case, it is ARM device that has `/dev/watchdog` too), will test for a few days.

blckbx:
A menu link is missing for this bonus guide in here: https://github.com/raspibolt/raspibolt/blob/master/guide/bonus/raspberry-pi/index.md

PR #1406, opened 2024-02-07
Using timer service instead of crontab to send ping to Amboss

SatsForFreedom:
What

The modification was initiated to address a specific issue where the curl command, responsible for sending the ping, could become unresponsive and consume 100% of a CPU's capacity. This problem was particularly evident during an incident where the internet was unavailable for an extended period, like a one-hour power outage. To prevent such occurrences, the service has been configured with a runtime limit to avoid the issue of the script getting stuck.

Why
If the ping script becomes stuck, it not only consumes excessive CPU resources but also potentially impacts the overall performance and reliability of the system. Moreover, in situations where each instance of the script doesn't terminate as expected, it could lead to cumulative resource drainage. This change is crucial to ensure that system resources are managed efficiently and to maintain the stability and responsiveness of the monitoring service.

How
The service has been enhanced with the RuntimeMaxSec option, which is set to 60 seconds. This means that if the script does not complete its execution within 60 seconds, Linux will automatically terminate the ping script. This safety mechanism ensures that no single instance of the script can monopolize CPU resources indefinitely, thereby protecting the system from potential overloads and maintaining the consistent performance of the monitoring tasks.

Scope
 significant change to core configuration
 independent bonus guide
 simple bug fix
Fixes # (link issue)

Test & maintenance
The script is currently operational, and the performance and status of my node, named "SatsForFreedom," can be verified in real-time on Amboss.


guide/bonus/lightning/ambossping.md
```diff
@@ -113,36 +112,83 @@ Table of contents
...
+### Creating a Timer Service
 
-  Depending on what you have selected in your [amboss monitoring dashboard](https://amboss.space/owner?page=monitoring), you would like to schedule your script to send pings on preselected intervals. The ambossping crontab should be clean at this point, so add the following at the bottom:
- 
-  ```ini
-  * * * * * /home/ambossping/ping.sh >> /home/ambossping/ping.log
-  ```
-  This will execute the bash script every minute.
+1. **Initiate a Service File:**
+   While logged in as an “admin” user, initiate a timer service by creating a new file:
```
blckbx, line 122:
```suggestion
1. **Initiate a Service File:**

   While logged in as an “admin” user, initiate a timer service by creating a new file:
```

guide/bonus/lightning/ambossping.md
```diff
@@ -113,36 +112,83 @@ Table of contents
...
+    OnBootSec=900
+    OnUnitActiveSec=900
+    AccuracySec=1
+
+    [Install]
+    WantedBy=timers.target
+    ```
+
+2. **Set Up the Ping Service:**
+   Create a corresponding service file:
```
blckbx, line 146:
```suggestion
2. **Set Up the Ping Service:**

   Create a corresponding service file:
```

guide/bonus/lightning/ambossping.md
```diff
@@ -113,36 +112,83 @@ Table of contents
...
+    ExecStart=/home/ambossping/ping.sh >> /home/ambossping/ping.log
+    User=ambossping
+    RuntimeMaxSec=60
+
+    [Install]
+    WantedBy=multi-user.target
+    ```
+
+3. **Activate the Services:**
+   Enable and start both the timer and service:
```
blckbx, line 169:
```suggestion
3. **Activate the Services:**

   Enable and start both the timer and service:
```

blckbx (commented):
ConceptACK
added some newlines after listings. preview has shown some concatenation.

PR #1408, opened 2024-02-22
Update JoinMarket to v0.9.11

kristapsk:
#### What & Why

Update to newest version. There are various performance improvements, bugfixes and adds new features to RPC API which will be needed for the next release of Jam.

Also removed `--linenumbers` from configuration section, as they change from version to version.

#### Scope

- [X] independent bonus guide
- [ ] simple bug fix


blckbx (approved):
LGTM

PR #1422, opened 2024-04-25
CLN: Remove deprecated configs

kristapsk:
#### What

CLN configs `autocleaninvoice-cycle` and `autocleaninvoice-expired-by` were deprecated with v22.11 and EOL v24.02. https://github.com/ElementsProject/lightning/pull/7094

### Why

I had random startup failures on different Ubuntu Server machine running v24.02.2 (not a RaspiBolt, it's x86_64, but that shouldn't change things much here.

#### Scope

- [X] independent bonus guide
- [ ] simple bug fix

### DIFF
--- a/guide/bonus/lightning/cln.md
+++ b/guide/bonus/lightning/cln.md
@@ -159,9 +159,6 @@ We will download, verify, install and configure CLN on your RaspiBolt setup. Thi
   large-channels
   # channel confirmations needed
   funding-confirms=2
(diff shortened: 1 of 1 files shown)


kristapsk:
Here removal of these solved problem for RaspiBlitz user - https://github.com/raspiblitz/raspiblitz/issues/4216.

blckbx (approved):
Good catch!

PR #1442, opened 2024-08-13
Removal of Tunnelsats Guide

blckbx:
#### What

Removal of Tunnelsats guide due to frequent changes of scripts which makes it difficult to keep up with.

#### Why

Using setup script is better alternative and less prone to errors. 

#### How

Removal of guide, image and index listing.

#### Scope

- [x] removal of independent bonus guide
- [ ] simple bug fix


twofaktor (approved):
LGTM

PR #1446, opened 2024-09-09
Change usage of .bitcoin, .lnd, and .loop links

### LINKED ISSUE #1443: [BUG] Commands specifying "/home/bitcoin/.bitcoin/..." when "/home/admin/.bitcoin/..." might be better
### Describe the bug
After bitcoind autostart is set up and after restarting, the instructions say to use these commands to verify that bitcoind is working:
```
ls -la /home/bitcoin/.bitcoin/.cookie
tail -f /home/bitcoin/.bitcoin/debug.log
```

I get permission denied.

### Location in guide
https://raspibolt.org/guide/bitcoin/bitcoin-client.html#verification-of-bitcoind-operations

### Expected behavior
The reason for the permission denied is that I am installing on Debian 12 instead of Raspian and it seems that the default permissions for /home directories are `drwx------` on Debian and `drwxr-xr-x` on Raspian.

Regardless of Debian vs. Raspian, I feel that it might be better to go through `/home/admin/.bitcoin` when using the admin account here because that is what the `.bitcoin` link is for. Anyway, this is very minor because those commands work on Raspian.

### Environment
 - Hardware platform: x86, 12 GB RAM, 2 TB SSD
 - Operating system: Debian
 - Version: Bookworm

jambolo:
#### What

Change usage of "/home/bitcoin/.bitcoin" to "/data/bitcoin" for users other than _bitcoin_.
Change usage of "/home/lnd/.lnd" to "/data/lnd" for users other than _lnd_.
Change usage of "/home/lit/.loop" to "/data/loop" for users other than _lit_.

### Why

Links to `/data/bitcoin` and `/data/lnd` owned by a user should only be used by that user. Also, note that while home directories are `rwxr-xr-x` by default on Raspian, they are `rwx------` on Debian, making these changes necessary for Debian.

#### How

Replaced `/home/bitcoin/.bitcoin` with `/data/bitcoin` and `/home/lnd/.lnd` with `/data/lnd` where referenced by users other than _bitcoin_ and _lnd_.

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

Fixes #1443

#### Test & maintenance

No maintenance is necessary while bitcoind and lnd remain installed in `/data`.

### DIFF
--- a/guide/bitcoin/electrum-server.md
+++ b/guide/bitcoin/electrum-server.md
@@ -153,7 +153,7 @@ We get the latest release of the Electrs source code, verify it, compile it to a
 
   # Bitcoin Core settings
   network = "bitcoin"
-  daemon_dir= "/home/bitcoin/.bitcoin"
+  daemon_dir= "/data/bitcoin"
   daemon_rpc_addr = "127.0.0.1:8332"
   daemon_p2p_addr = "127.0.0.1:8333"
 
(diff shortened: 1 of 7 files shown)


kristapsk (approved):
cr utACK, this makes sense.

PR #1447, opened 2024-09-09
Update troubleshooting.md showing Lightning now running under the user lnd

jambolo:
#### What

The troubleshooting page is out-of-date. It shows lnd under the user bitcoin, when it is run as the user lnd.

### Why

It could be confusing if the output doesn't match.

#### How

I just changed "bitcoin" to "lnd" where necessary and added commands to switch to lnd user when doing lnd stuff.

#### Scope

- [ ] independent bonus guide
- [x] simple bug fix


kristapsk (approved):
utACK 872bf26ec79beb5cb81b668efdfe9002e214fa1f

PR #1483, opened 2025-05-01
🛠️ Update Bonus Guide Bisq: prune & alt start cmds

AngelLozan:
#### What

The Bisq connection guide for connecting to your own node specifies that you must connect using certain configuration, including an option to not prune old blocks `prune=0` in `bitcoin.conf`. https://bisq.wiki/Connecting_to_your_own_Bitcoin_node#Troubleshooting

Without this setting, Bisq has difficulty maintaining connection to a hidden service (onion address) for a node using a custom connection and may fail to connect to your node. 

This also adds helpful context to the start commands for Bisq via the CLI, whereby a user who does not have `Bisq` as part of their system's `PATH` could be confused as to how to properly start the application using the given command. Adds additional commands, using the full application path in one scenario (most common) for a user who installed Bisq using the "drag and drop" method on MacOS.

### Why

Allows the connection of a self-hosted node to Bisq via the Tor network .
Allows less experienced users to find alternatives to the given commands to start Bisq via the CLI as instructed. 

#### How

Adding additional context to two sections of the md file, one being the section on the `bitcoin.conf` and one on the section relating to starting Bisq via the CLI.

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

Fixes # (link issue)

#### Test & maintenance

Testing is done via starting Bisq and connecting successfully. Unfortunately I did not write a test for this, but rather referenced the documentation and my own experience, for the respective changes described in the `What` section.

I am available to maintain this change and am seeking to make more updates based on my experience implementing this guide successfully. 

#### Animated GIF (optional)

BTC To the Moon!

### DIFF
--- a/guide/bonus/bitcoin/bisq.md
+++ b/guide/bonus/bitcoin/bisq.md
@@ -83,6 +83,7 @@ For Bisq to connect to your Bitcoin Core node, the bloom filters have to be acti
   ```sh
   # Filters
   peerbloomfilters=1   # for Bisq
+  prune=0              # for Bisq (disable pruning old blocks)
   ```
 
 * Add the following line under `"whitelist=download@127.0.0.1       # for Electrs"` line to whitelist our own P2P connection.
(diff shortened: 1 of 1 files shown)


guide/bonus/bitcoin/bisq.md
```diff
@@ -83,6 +83,7 @@ For Bisq to connect to your Bitcoin Core node, the bloom filters have to be acti
   ```sh
   # Filters
   peerbloomfilters=1   # for Bisq
+  prune=0              # for Bisq (disable pruning old blocks)
```
kristapsk, line 86: Bitcoin Core doesn't prune by default, you don't need to specify this if there isn't `prune=` setting in configuration.

AngelLozan: Agreed. More of an explicit call out if it's set as you suggest, which would effectively fail any connection to bisq.

PR #1501, opened 2025-10-12
bitcoin-client: update to 30.0

blckbx:
### What

Update Bitcoin Core client to latest version 30.0. 
Added: old default value of setting `datacarriersize` to `bitcoin.conf` to make users aware of it and adjust it to their preference.

#### Scope

- [ ] independent bonus guide
- [ ] simple bug fix

### DIFF
--- a/guide/bitcoin/bitcoin-client.md
+++ b/guide/bitcoin/bitcoin-client.md
@@ -23,7 +23,7 @@ We install [Bitcoin Core](https://bitcoin.org/en/bitcoin-core/){:target="_blank"
 ## This may take some time
 
 Bitcoin Core will download the full Bitcoin blockchain, and validate all transactions since 2009.
-We're talking more than 800'000 blocks with a size of over 500 GB, so this is not an easy task.
+We're talking more than 900'000 blocks with a size of over 700 GB, so this is not an easy task.
 It's great that the Raspberry Pi 4 can do it, even if it takes a few days, as this was simply not possible with earlier models.
 
 ---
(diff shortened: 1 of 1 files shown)


guide/bitcoin/bitcoin-client.md
```diff
@@ -261,6 +261,10 @@ We'll also set the proper access permissions.
   zmqpubrawtx=tcp://127.0.0.1:28333
   whitelist=download@127.0.0.1          # for Electrs
 
+  # Maximum size of data in OP_RETURN outputs we relay and mine. New default: 100_000 bytes
+  # old setting pre 30.0 below, adjust to your preference
+  datacarriersize=83
```
kristapsk: Note that this does not have pre-30 behaviour fully, which permitted only single OP_RETURN output, this setting now sets limit for combined size of all OP_RETURNs in a single tx.

2pac1:
What about the following:

* update the guide for v0.30 Bitcoin Core (as long as v0.30 doesn’t break existing instructions), but include concise notes that explain **OP_RETURN** in the simplest possible way
* mention that users who want to preserve legacy behavior can still use the `-datacarriersize=` option
* note that anyone who does **not** want to relay any non-Bitcoin-related data can set this to `0`, or alternatively use `-datacarrier=0`
* this is useful for users who want to run a node without relaying data unrelated to Bitcoin transactions
* add Bitcoin Knots as an optional alternative for users who prefer it

Given that Knots includes more than just the OP_RETURN change — with many additional lines of code — and considering the tyranny of the defaults in v0.30 Core, it’s especially important for anyone using RaspiBolt to read the guide carefully and configure their setup according to their preferences, needs, and wants.

At the end of the day, if someone prefers defaults and one-click setups without granular control, they’ll probably choose a different installation method/guide.

guide/bitcoin/bitcoin-client.md
okeiprivat55: I am ok

PR #1505, opened 2025-10-17
Update Bitcoin client from 29.1 to 29.2

kristapsk:
#### What

Update Bitcoin Core from 29.1 to 29.2. There is a pull request #1501 to update to 30.0, but: 1) that is currently not yet compatible with JoinMarket bonus guide (and probably others, haven't checked), 2) could take some time for consensus because of all the drama currently going on regarding changes of default mempool relay policies. This one should be no brainer, let's merge it first and then discuss path forward.

#### Scope

- [X] significant change to core configuration
- [ ] independent bonus guide
- [ ] simple bug fix


kristapsk:
> Something worth mentioning- 29.0, 29.1, 29.2 and 30.0 perform horribly on rpi’s - I have tried upgrading multiple times and have had to go back to 28.2 every time. There are open issues on the bitcoin core GitHub about it

That's different issue, here we are just changing from 29.1 to 29.2.

blckbx (approved):
tACK b551942

PR #1506, opened 2025-10-22
Update bullseye for bookworm for TOR repo and RaspiOS description

Pinipon-1:
- Changed bullseye for bookworm

Using the rpi imager Legacy image now installs the version based in Debian Bookworm insted of Bullseye, therefore it should be updated

Just to keep the guide up to date


blckbx (approved):
utACK (but verified on Debian 13 PR #1507)

PR #1508, opened 2025-10-29
Update bitcoin-client.md

Pinipon-1:
As the link was created in a previous step, the correct path would be "tail -f /data/bitcoin/debug.log" instead of "tail -f /home/bitcoin/.bitcoin/debug.log", otherwise there could be a permission error

### DIFF
--- a/guide/bitcoin/bitcoin-client.md
+++ b/guide/bitcoin/bitcoin-client.md
@@ -432,7 +432,7 @@ After rebooting, "bitcoind" should start and begin to sync and validate the Bitc
   Exit with `Ctrl-C`
 
   ```sh
-  $ tail -f /data/bitcoin/debug.log
+  $ tail -f /home/admin/.bitcoin/debug.log
   ```
 
 * Use the Bitcoin Core client `bitcoin-cli` to get information about the current blockchain


blckbx (approved):
tACK 769dad9
Thanks! Yes, I think that's one case where we hit the new restriction of users' home dir (since Debian 12, see #1343). IMHO to access the data dir instead of the bitcoin user's home dir is a good solution. Alternatively as user `admin` we can get access via `tail -f /home/admin/.bitcoin/debug.log` because we have set this link a few steps earlier.

blckbx:
@Pinipon-1 would you want to change path to `/home/admin/.bitcoin/...` or leave as is?

Pinipon-1:
It could be either `/home/admin/.bitcoin/debug.log` or `/data/bitcoin/debug.log`, I'm not sure which one would be more correct, probably /home/admin as we're supposed to be using the admin user

blckbx:
I think `/home/admin/` would be reasonable but I'll leave it up to you to decide. Gimme a shout what you'd like to to and I'll merge it.

guide/bitcoin/bitcoin-client.md
```diff
@@ -432,7 +432,7 @@ After rebooting, "bitcoind" should start and begin to sync and validate the Bitc
   Exit with `Ctrl-C`
 
   ```sh
-  $ tail -f /home/bitcoin/.bitcoin/debug.log
+  $ tail -f /data/bitcoin/debug.log
```
blckbx:
```suggestion
  $ tail -f /home/admin/.bitcoin/debug.log
```

blckbx (approved):

PR #1510, opened 2025-11-17
Update Fulcrum to v2.0

Pinipon-1:
Update Fulcrum to 2.0 and change config to delete deprecated options

#### Scope

Update Fulcrum to the latest version, which is supposed to be more stable and less prone to corrupted db issues.


blckbx (changes requested):
some nits. I don't run fulcrum, so it would be good to get a third-party confirmation here. Formally it LGTM

PR #1511, opened 2025-11-21
Update the netcat package for one used in Debian 12 - For Fulcrum install

Pinipon-1:
Following the update to Debian 12 (or 13), netcat has no installation candidate for Debian 12, either netcat-openbsd or netcat-traditional should be used instead


blckbx (approved):
I don't remember installing netcat-openbsd explicitly for Debian 13 but it could be me... so, I'd give it a LGTM after checking `apt search ^netcat` and there's indeed only the two options you mentioned.

PR #1512, opened 2025-11-23
Updated nodejs, and using the latest version of BTC RPC Explorer

Pinipon-1:
Just following the update uf the guide as the latests PRs


blckbx:
Regarding version: I'm running my node with nodejs LTS which currently is v22.15.0. Shall we point to `setup_lts.x` in general? 

Also I think we need a separate NodeJS guide because it's used in more parts of the guide. Todo for a separate PR.

PR #1555, opened 2026-08-31
content(bonus): add Alby Hub guide, ported to v4

Stadicus:
#### What

Adds `guide/bonus/lightning/alby-hub.mdx`: running [Alby Hub](https://albyhub.com) against the LND node the reader already has, so NWC apps get per-app spending budgets without ever seeing the admin macaroon.

Ported from @MoritzKa's contribution in [Stadicus/RaspiBolt#4](https://github.com/Stadicus/RaspiBolt/pull/4), which was opened against the archived mirror and has been sitting unanswered for two months. He is credited as co-author on the commit.

### Why

The mirror is not where anyone looks, so the contribution was going to rot there. The content is worth having, and the v4 port was the work it needed.

#### How

Structure, LND-backend approach, systemd unit, Tor hidden service, and uninstall flow are the contributor's. The v4 conversion is the usual: front matter, `%versions.X%` / `%files.X%` / `%urls.X%` tokens, one command per `bash` block, `<Callout>` instead of Jekyll labels, `/docs/...` links, image to `public/images/`.

The original was marked **"Status: Untested"** by its author, and that turned out to matter. Five defects found and fixed, each verified against the real v1.24.0 release:

| Defect | Evidence |
|---|---|
| Download filenames wrong (`server-linux-aarch64.tar.bz2`) | The URL returns **404**. Actual asset: `albyhub-Server-Linux-aarch64.tar.bz2` |
| No verification step at all | Release ships a signed `manifest.txt`. Added SHA256 + PGP against `github.com/rolznz.gpg`, verified: `Good signature from "Roland Bewick"`, fingerprint `5D92 1859 38E6 DBF8 93DC CC5B A5EA BD88 3509 2B08` |
| `PORT=8080` collides with LND's REST listener | Moved to 8095, with a callout explaining why |
| Upgrade section calls `./update.sh` | Not in the server archive. `tar -tjf` shows only `bin/albyhub` and `lib/libldk_node.so` |
| nginx `streams-enabled` reverse proxy | Replaced with Caddy on `:4090`, matching v4 |

Two further changes: macaroon handling now follows the RTL pattern (read-only copies owned by the service user, rather than `adduser albyhub lnd`), and the `bin/`+`lib/` layout is documented, since the binary resolves `libldk_node.so` through an embedded `$ORIGIN/../lib` RPATH and breaks if the directories are separated.

A gotcha worth flagging for review: `sha256sum --check manifest.txt` always reports `FAILED` for `manifest.txt` itself, because the manifest lists its own checksum as the hash of an empty file. The guide scopes the check with `grep` and explains why, so readers don't hit a false alarm.

#### Not verified

**Nobody has walked this on hardware.** Every artifact claim above is checked, but the running system is not: the LND gRPC handshake, the first-run web setup, and the Caddy path are reasoned from Alby Hub's docs and the v4 conventions. The page carries a callout saying so. If that is not the bar for merging a bonus page, this should sit until someone runs it on a Pi.

#### Checks

`npm run build`, `npm run types:check`, Vale, markdownlint, Prettier, gitleaks, typos: all clean locally.

#### Scope

- [x] independent bonus guide
- [ ] simple bug fix

Closes Stadicus/RaspiBolt#4

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011ykGx7WuyQ14k9FB3eJVfu


Stadicus:
### Note on the red linkcheck

The failure is not a link in this page. It is the auto-generated "edit this page" link, and **any PR that adds a new page will hit it**:

```
[404] https://github.com/Stadicus/RaspiBolt/blob/feature/v4-rewrite/guide/bonus/lightning/alby-hub.mdx
```

`src/lib/shared.ts` points non-production builds at the mirror:

```ts
export const gitConfig = isProductionSite
  ? { user: 'raspibolt', repo: 'raspibolt', branch: 'master' }
  : { user: 'Stadicus', repo: 'RaspiBolt', branch: 'feature/v4-rewrite' };
```

Existing pages resolve fine there (`.../feature/v4-rewrite/guide/bonus/lightning/cln.mdx` returns 200), so this has never fired before. A page added in a PR does not exist on that branch yet, so its edit link 404s. The content links in this PR all check out.

Two ways to fix it, both outside the scope of a content PR:

1. Add the source-link pattern to the exclude list in `lychee.toml`, mirroring what `.lychee.toml` already does for `github\.com/.*/edit/`. That file's excludes do not cover the `/blob/` form.
2. Point the non-production `gitConfig` at `raspibolt/raspibolt` and derive the branch from `GITHUB_HEAD_REF`, so PR builds link at their own branch. More correct, more moving parts.

I'd take option 1 now and option 2 if PR previews ever land. Happy to open it as its own PR, but it changes CI config, so that is your call rather than mine.

Stadicus:
### Walked it against a live LND

Since the open question was whether the `.env` in this guide actually makes Alby Hub talk to LND, I built the stack and ran it: **Bitcoin Core v31.1 + LND v0.21.2-beta on regtest + Alby Hub v1.24.0**, driven through its setup API.

| Step | Result |
|---|---|
| `grep ... \| sha256sum --check` | `./albyhub-Server-Linux-x86_64.tar.bz2: OK` |
| `gpg --verify manifest.txt.asc` | `Good signature from "Roland Bewick"` |
| Archive layout + RPATH | Binary starts with no `LD_LIBRARY_PATH`; `bin/` + `lib/` as documented |
| The `.env` from this guide | `/api/info` → `"backendType":"LND"` |
| gRPC handshake with the copied cert + macaroon | log: `{"alias":"raspibolt-test","msg":"Connected to LND"}` |
| `/api/node/status` | `isReady: true`, real pubkey `02001ac2…`, network `regtest`, block height 65 |
| `/api/apps` | Issued a working `nostr+walletconnect://…` pairing URI with a 10k sat monthly budget |

The **port collision is confirmed empirically**, not just inferred. LND's own startup log says `gRPC proxy started at 127.0.0.1:8080`, and with Alby Hub moved to 8095 both processes hold their sockets at the same time. On the original PR's `PORT=8080` one of the two would have failed to bind.

#### Still unproven

No Docker or systemd on the machine I have, so `testing/vm/` was not an option:

- **ARM64 binary.** The walk used the x86_64 build of the same release. Same code, different artifact.
- **The systemd unit.** Never executed. Reviewed by eye only.
- **Caddy and Tor sections.** Not exercised.
- **Mainnet macaroon path.** The walk used the regtest path; the guide's `/data/lnd/data/chain/bitcoin/mainnet/admin.macaroon` is unchanged from the main guide's own convention.
- **The browser wizard.** I drove the same API the UI calls, not the UI.

The callout on the page has been rewritten to say exactly this rather than "nobody has run them end to end". Pushed as 7bf0cb6.

